PR #32, opened 2014-07-27
Added ability to disable L2TP/IPsec at installation.

ddworken:
This way the most easily detectable service can be disabled at installation.

### DIFF
--- a/streisand
+++ b/streisand
@@ -10,6 +10,12 @@ then
     exit -1;
 fi
 
+read -p "Do you want to install L2TP/IPsec? Installing this is generally safe but it is easily detectable. <y/N> " prompt
+if [[ $prompt == "n" || $prompt == "N" || $prompt == "no" || $prompt == "NO" ]]
+then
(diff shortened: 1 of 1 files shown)


pjrobertson:
To disable l2tp-ipsec, do you only need to remove the line from `streisand.yml`? Installing l2tp-ipsec takes a pretty long time on small servers, and it's not very useful in a lot of countries/cases.

ddworken:
@pjrobertson It actually did work in the past but it looks like that has changed. With enough work one could probably disable it but doesn't seem to be quite as easy anymore.

pjrobertson:
I managed to get it working yesterday. I think the difference was that I wanted to deploy on my own server, not an AWS/Google/DigitalOcean server. What I did was:

* Copy the `amazon.yml` file to `my-playbook.yml`
* Remove all Amazon specific stuff (making sure to leave in `l2tp_ipsec_firewall_rules: ""` and `l2tp_ipsec_gateway_location: ""`)
* Update inventory with my server's IP address/SSH port
* run `ansible-playbook playbooks/my-playbook.yml`

I think the key was defining the two `l2tp_ipsec...` rules somewhere

PR #34, opened 2014-07-27
Server Provisioning Cleanup

ShenZhouHong:
I did some minor cleanup and standardized the prompts in the server provisioning yml files at streisand/playbooks/[host].yml. The prompts were slightly changed to inform users of the default values, and for server providers that have multiple datacenters in a geographical regions, I used tabs to make it look better.

### DIFF
--- a/playbooks/amazon.yml
+++ b/playbooks/amazon.yml
@@ -25,19 +25,20 @@
     - name: "aws_region_var"
       prompt: >
         What region should the server be located in?
-          1. Asia Pacific (Tokyo)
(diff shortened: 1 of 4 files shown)


romseguy:
Any place where I can track the progress of the local provisioning feature I'm interested in? Thank you for this awesome script.

peterpacz1:
I've became more aquainted with ansible in the meantime - and I'm sure I can work out another commit/bash script for local provisioning. It should be no more difficult than setting the hosts file as 127.0.0.1:22. I'll try to figure something out.

PR #69, opened 2014-09-07
Use current-generation AWS instances - twice the RAM for the same price

DanielHeath:


jlund:
Not even the brand-new Debian 8 AMIs support t2.micro instances. Mysterious...

df-sh:
https://aws.amazon.com/marketplace/pp/B00WUNJIEE debian builds support t2.micro instance in the free tier. 

I am trying to spin up a couple of these now

pchaganti:
Anyone able to get t2.micro to work with the Debian images?

thx!

liamuk:
It seems like openvpn isn't getting started automatically, causing dnsmasq to fail when it sees it needs to bind to an openvpn-provided ip.

I was able to get t2.micro working with Debian 8 by removing the restart dnsmasq line, and then manually starting openvpn and dnsmasq at the end (/etc/init.d/openvpn start, /etc/init.d/dnsmasq restart). It looks like everything is in working order.

jlund:
This has been fixed as part of #174. `t2.micro` instances are now the default for new AWS instances in the `ubuntu` branch that will be merged into master soon.

jlund:
Soon is now. This is finished.

Getting the other providers switched to 14.04 ended up being really easy after the massive L2TP/IPsec overhaul was done. Performance on the `t2.micro` instances is _significantly_ better.

PR #94, opened 2014-11-27
Adding force=yes Tor and Nginx

cdario:


cdario:
It works flawlessly now. Thank you for the amazing work!

PR #107, opened 2015-01-03
Bumped stunnel version number

lingz:
Current stunnel verison is no longer hosted at the given url. Causes the installer to fail. I've tested and verified that bumping the version number lets it succeed.


jlund:
I'm not sure how I missed this, but I made the same change in another commit. Thank you for your help!

PR #116, opened 2015-01-21
tor 64 bit for OSX

hogsim:
Tor browser 32 bits doesn't exit for the version 4.5a3.


jlund:
Thanks for letting me know! It looks like the format of the `RecommendedTBBVersions` file has changed somewhat. `4.0.3` is still the latest stable release (with a 32-bit binary for OS X), and this alpha version was being pulled in by mistake.

PR #134, opened 2015-03-04
Update instructions.md.j2

skeeve42:
Updated instructions to add servers using a QR code on both Windows and OS X.


skeeve42:
So I just tried to install shadowsocks on a different windows machine using the streisand instructions, and I realized the instructions and mirrors are pointed to the old shadowsocks-gui:
https://github.com/shadowsocks/shadowsocks-gui.
The project has now been split, so the links are are outdated...the mirror downloads and old version of shadowsocks that does not support QRcode scanning.

skeeve42:
https://github.com/shadowsocks/shadowsocks
This is the main page for the server which has links to all the client projects.

peterpacz1:
Looks good to me - let's hope Jlund sees this and the other requests.

jlund:
Thanks, @skeeve42! I'm updating the files that get mirrored now.

jlund:
All set!

skeeve42:
Cool!

PR #150, opened 2015-04-06
Fixed logic in Debian 7 warning

lionello:
Show show the warning when the OS is not Debian based OR not version 7.


jlund:
Thanks! Can you explain what issue this was causing? The "and" version seems a little more robust because Debain _and_ 7 is the only supported combination right now. It's entirely possible that I'm missing something though.

lionello:
If you want to ensure Debian 7, you'd want to use an 'and' for positive logic: OS=Debian and version=7. But you're testing for the negative case, so the 'and' becomes an 'or': OS!=Debian or version!=7.

This is a boolean transformation rule:
!(A & B) becomes !A | !B

jlund:
You are absolutely right. Thank you!

PR #151, opened 2015-04-07
Don't hardcode eth0 in dante configuration

PiPeep:
Ansible provides `ansible_default_ipv4.interface`, which may or may not
be eth0. For example, on OpenVZ hosts, this is usually venet0.

### DIFF
--- a/playbooks/roles/dante/tasks/main.yml
+++ b/playbooks/roles/dante/tasks/main.yml
@@ -3,9 +3,9 @@
   apt: name=dante-server
 
 - name: Copy Dante configuration file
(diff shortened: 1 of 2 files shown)


pjrobertson:
+1
I just came across this on a KiviWM host. The changes here fixed the problem.

jlund:
Good catch. I am using `ansible_default_ipv4.interface` in several places, but I missed this one.

jlund:
Also, Firefox for Android fixed the bug that necessitated the Dante workaround. I will likely be removing the Dante role entirely soon.

Thank you very much for your help!

PR #152, opened 2015-04-07
Make openvpn crypto parameters easier to modify

PiPeep:
Because of ansible's [variable precedence][], non-default variables can't be easily modified via config files. Because the OpenVPN crypto parameters might be something someone wants to modify, they should instead be set as default values.

[variable precedence]: http://docs.ansible.com/playbooks_variables.html#variable-precedence-where-should-i-put-a-variable


jlund:
Thanks!

PR #158, opened 2015-05-06
add Linode Singapore DC, replace Tokyo as default

displague:
Tokyo availability is limited, Singapore was recently added and does not have these limitations


jlund:
Thank you!

PR #159, opened 2015-05-21
Improved connection list description

peterpacz1:
Hi there. I added a bit more information on the streisand gateway page about the connection methods, and the list of mirrors. A common cause of confusion for my users is that they are not sure which method to pick, so I made that clearer.

This is a no-frills commit, so it should be easily mergeable.


jlund:
Thanks!

PR #181, opened 2015-07-25
Adding in Ping Keepalives to Servers Internal IP

### LINKED ISSUE #180: TCP Timeout on OpenVPN sslh based connections
Having issues that connection resets on say corporate networks which clean up idle TCP sessions.

Russell-IO:
Adds in ping keepalives into OpenVPN Client common configuration, this causes keep alives to be sent, resolves #180 :)


jlund:
Unfortunately, it looks like the iOS OpenVPN client does not support this option. I had to revert the change for now. Hopefully that will be updated soon.

PR #192, opened 2015-09-28
Updating gce

CorbanR:
- Adding latest gce.py
- Updating gce image with proper Ubuntu name.
- Resolving ntp conflict.


jlund:
Thanks! Google's documentation on the Ubuntu image names [appears to be incorrect.](https://cloud.google.com/compute/docs/operating-systems/#ubuntu)

PR #197, opened 2015-10-28
option to customize which DNS servers are used

aeibrahim:
PR created to allow users to set alternative dns servers other than Google's DNS and customize the TLS cipher suites used by OpenVPN. 

Created `group_vars` directory to store global variables used across all roles. Two DNS servers can be set here:

    upstream_dns_servers: 
     - 8.8.8.8
     - 8.8.4.4

`etc_dnsmasq.conf` and `options.xl2tpd` converted to templated files and are transferred to the server using the template module. 

TLS cipher suites can be set in `openvpn/defaults/main.yml` with strong defaults.


jlund:
Thanks!

Unfortunately, I had to temporarily revert the OpenVPN control channel cipher changes because they were breaking older clients (like OpenVPN on iOS). We are back to using the defaults for now until we can come up with a list that is more broadly compatible.

aeibrahim:
Thanks for reviewing it. Sorry, I didn't test the cipher options against iOS clients, but I will and suggest ones that are compatible for Windows, Linux, and Android/iOS clients

jlund:
No problem! I appreciate the help.

Issue #43, opened 2014-07-30
Update instance


ghost:
What is the recommended way to update my `streisand` instance? Just running `sudo apt-get update && sudo apt-get upgrade` after ssh into my machine or running `./streisand` once again?

pjrobertson:
Should this issue be re-opened until a section is added to the README about updating servers? I know it's been a year...!

PR #199, opened 2015-11-15
Add information on upgrading Streisand

pjrobertson:
I've added the information from #43 to the readme, so hopefully it's easier for other people to find in the future

### DIFF
--- a/README.md
+++ b/README.md
@@ -150,6 +150,15 @@ You can also run Streisand on any number of new Ubuntu 14.04 servers. Dedicated 
 
 The servers should be accessible using SSH keys, and *root* is used as the connecting user by default (though this is simple to change by updating the streisand.yml file or ansible.cfg file).
 
+### Upgrading a Streisand instance ###
+
+If all you wish to do is make sure installed packages are up to date, you can run `apt-get update` and then `apt-get upgrade` on your Streisand server.    
+Otherwise, if you wish to upgdate Streisand's web pages & documentation, or install new services provided by Streisand in a recent update, you can do so by re-running the `streisand.yml` playbook on an existing server. For example:
+
+    ansible-playbook playbooks/streisand.yml
+
+**Note:** re-running the `streisand.yml` playbook may result in duplicate firewall entries, but this should not affect security or performance.
+
 Upcoming Features
 -----------------
 * Native [Microsoft Azure](https://azure.microsoft.com/en-us/) support


jlund:
I'll have to think about this one. The move from Debian 7 to Ubuntu 14.04 was a pretty massive change, for example, and users with an old Debian 7-based Streisand server would not be able to simply re-run the playbook to get the latest updates due to the significant differences between those two distributions.

Given how quick the process is, I would almost rather encourage users to create a new server from scratch if they want the latest Streisand features. Trying to track the necessary migrations would be a huge and complicated undertaking, and wouldn't even be possible for something like a change in the base Linux distribution.

As for package updates, I want to automate that using unattended-upgrades so that no one has to run `apt-get update && apt-get upgrade` by hand. I will try to get to that soon.

pjrobertson:
> I would almost rather encourage users to create a new server from scratch
if they want the latest Streisand features

In reality that is what I do. It's a little bit of a hassle to re-deploy
all keys, so perhaps the simplest thing to do would be to add a way of
"exporting" an existing Streisand's set of keys/configs, then on a new
server uploading them somewhere to make streisand use them.

Maybe the playbook, once complete, should copy all user-generated stuff to
~/streisand/backup or something, then the first thing a playbook should do
is check to see if ~/streisand/backup exists. If it does, then it copies
the keys/configs from there as opposed to creating them from scratch.

If, for example, some config or key *has* to change (and be created from
scratch) then there could be some kind of versioning. E.g. you could place
all the 'backup' file in a commit-SHA named folder (e.g.
~/streisand/backup/ef284ab928c) then each playbook could run something at
the start like "if commit is older than X, don't use the backup copy"

You may have a better idea since you know Ansible/Streisand well, but
that's the only gripe I have with deploying a new server.

kydan:
While I would agree that spinning a new instance instead of upgrading an
existing server would be preferred I think Patrick's configuration export
option is warranted.

I currently have to spin a new instance every few days or so due to Chinese
ISPs identifying and subsequently actively throttling connections to that
IP. I achieve this by using a snapshot of an original Streisand instance,
but any time I want to upgrade, I have to deploy new keys across my many
devices as well as  inform other users of the change.

I'm afraid I probably don't have the expertise to implement this in a
timely fashion though.

sent from my mobile

CorbanR:
I happened to run `apt-get update` and then `apt-get upgrade` on an old Streisand install (testing something unrelated). Doing this caused TOR to fail. Something to think about when suggesting a full update.

genewitch:
re: https://github.com/jlund/streisand/pull/199#issuecomment-160012698

Depending on the provider you can bounce the IP without spinning a new instance. I am not sure how much of the software requires a static IP (if any). On AWS a terminate + start will issue a new IP, for instance.

strixaluco:
https://github.com/jlund/streisand/pull/199#issuecomment-166738992
I don't know which kind of setup @kydan has, but I have Streisand running on my self-hosted router with constant IP and yes, eventually Chinese ISPs identify and substantially throttle connections to that IP address and this is frustrating.

df-sh:
@kydan not sure which ISP is actively throttling your streisand IP connection, but if you are using AWS, you can change the IP of the server in about 30 seconds without having to spin up a new server...

I am based in Shanghai and use rackspace and amazon hosts for streisand and apart from spinning up new servers for testing, I dont experience any throttling or server disruption and this with more than 1 TB of DAILY traffic to each of these servers.

jlund:
Looks like I forgot to close this PR. We now have `unattended-upgrades` support built-in thanks to @CorbanR! The README has been updated to reflect this.

@df-sh Those traffic numbers make me very, very happy :)

kydan:
@df-sh I am in Yunnan with China Telecom for home service. Who is your ISP
and which method do you most commonly use to connect? My service is also
spotty as it is, (I think it may be the shitty fiber modem they give you),
as things seem to usually operate better over mobile data, but I see
similar issues, especially DNS poisoning on nearly any connection.

I am on DO based out of Singapore, and usually speeds of less than 1 mbps.



sent from my mobile
2016年2月19日 下午7:36，"df-sh" <notifications@github.com>写道：

kydan:
@genewitch sorry just noticed your message here. Shadowsocks, for one, does require the IP. it may be possible to get around this by providing it with a hostname instead of an IP (I've currently have an extremely short TTL on a subdomain that is updated by the script I wrote to tear down and pop up a new instance on DigitalOcean). I believe other services require this as well.

kydan:
@df-sh I would also be interested in which datacenters you are running these hosts on, any optimizations you've done etc. The extremely sorry state of internet connectivity drives me insane. Would you be comfortable taking this convo off of a issue thread? Perhaps you can reach out to me at @kydan2002 on twitter as a first step?

PR #201, opened 2015-11-15
Switch to using shadowsocks-libev

### LINKED ISSUE #200: Switch to shadowsocks-libev
Since the python version of Shadowsocks has been [taken down](https://github.com/shadowsocks/shadowsocks), Streisand should switch to use the -libev version of the library. I've also found this to be a little bit faster than the python version, and it's just as easy to install.

pjrobertson:
This fixes #200 

I'm new to ansible, so there are probably a few things that could do with changing.

Some things to note:

* shadowsocks-libev provides an init script, which should make maintenance of this playbook simpler
* in using the provided init script, the shadowsocks-libev service is not run as root (what they default to). Not sure of the consequences of this
* The config file is set in the init script to `/etc/shadowsocks-libev/config.js` so I have just symlinked the file that ansible creates to this location (again, not sure if this is good/bad)


playbooks/roles/shadowsocks/tasks/main.yml
```diff
@@ -17,27 +11,23 @@
   command: cat {{ shadowsocks_password_file }}
   register: shadowsocks_password
 
-- name: Install Shadowsocks
-  pip: name=shadowsocks
+- name: Add the Shadowsocks-libev APT key
+  apt_key: url=http://shadowsocks.org/debian/1D27208A.gpg state=present
```
jlund: The convention in Streisand is to include a local copy of the key in the repository and to add it using the `data` argument of the `apt_key` module. [Here](https://github.com/jlund/streisand/blob/master/playbooks/roles/nginx/tasks/main.yml#L12) is an example of how that is being done in the Nginx role.

This is particularly important for security in a situation like this where the APT key is being served over plaintext HTTP and is therefore highly susceptible to a MITM attack. There are speed and reliability advantages as well.

playbooks/roles/shadowsocks/tasks/main.yml
```diff
@@ -17,27 +11,23 @@
...
+  
+- name: Add the Shadowsocks-libev repository
+  apt_repository: repo='deb http://shadowsocks.org/ubuntu trusty main' state=present
```
jlund: We would want to use the gathered Ansible facts in place of `ubuntu` and `trusty` in this line. [Here](https://github.com/jlund/streisand/blob/master/playbooks/roles/nginx/tasks/main.yml#L18) is an example of how that is being done in the Nginx role.

For the sake of future upgrades (e.g. to Ubuntu 16.04) we want to avoid having as many static references to the current LTS Trusty codename as we possibly can.

playbooks/roles/shadowsocks/tasks/main.yml
```diff
@@ -17,27 +11,23 @@
...
-            dest=/etc/init.d/shadowsocks
-            owner=root
-            group=root
-            mode=755
 
-- name: Enable the Shadowsocks init script so the service will start at boot
-  command: update-rc.d shadowsocks defaults
-           creates=/etc/rc0.d/K01shadowsocks
+- name: Symlink config to /etc/ directory
+  file: src={{ shadowsocks_location }}/config.json dest=/etc/shadowsocks-libev/config.json state=link force=yes
```
jlund: We should just generate the config file in the correct location and skip the symlink step.

jlund: Another convention within the Streisand repository is to place all module arguments on separate lines with proper indentation for maximum readability.

For example, instead of this:
```
file: src={{ shadowsocks_location }}/config.json dest=/etc/shadowsocks-libev/config.json state=link force=yes
```

We do this:
```
file: src={{ shadowsocks_location }}/config.json
      dest=/etc/shadowsocks-libev/config.json
      state=link
      force=yes
```

jlund:
This looks really great. Thank you!

I'm excited to see that the libev port is being actively maintained. It looks like there are a lot of cool new features in addition to better performance. I'm completely on board with the switch.

I added a few minor comments. Let me know if you have any questions. I can also handle making these changes if you are short on time.

pjrobertson:
Great! I will work on this in the next few hours and update the pull. As I
said I'm new to Ansible so thanks for your comments!

pjrobertson:
@jlund I've made the updates based on your comments, and rebased against master. Let me know if there's anything else you want doing :-)

jlund:
I really appreciate the help! Thanks again. You definitely seem to be learning Ansible quickly.

jlund:
I really appreciate the help! Thanks again. You definitely seem to be learning Ansible quickly.

pjrobertson:
> I really appreciate the help!

Thanks for the merge! I'm glad I could help

> Thanks again. You definitely seem to be learning Ansible quickly.

Learning by example is the best way to learn :-)

PR #207, opened 2015-11-24
made some tasks asynchronous to speed up deployment

aeibrahim:
Noticed two tasks that take a long time to complete and hold up the rest of the playbook. Made these two tasks run asynchronously, so they can run in the background while the rest of the playbook continues. 
- The DH parameters generation for OpenVPN needed to be complete before the service starts, so I placed a task to verify that it completed before the openvpn restart task. 
- As for compiling Libreswan, I placed the verify task at the end of the play right before the handlers run.

### DIFF
--- a/playbooks/roles/streisand-gateway/tasks/main.yml
+++ b/playbooks/roles/streisand-gateway/tasks/main.yml
@@ -128,6 +128,12 @@
         state=directory
         recurse=yes
 
+- name: Verify libreswan compilation has completed
+  async_status: jid={{ compile_libreswan.ansible_job_id }}
+  register: async_task_result
+  until: async_task_result.finished
+  retries: 100
+
 - meta: flush_handlers
 
 - name: Success!
--- a/playbooks/roles/l2tp-ipsec/tasks/main.yml
+++ b/playbooks/roles/l2tp-ipsec/tasks/main.yml
@@ -25,6 +25,9 @@
 - name: Compile and install Libreswan (this will take a while)
   shell: make programs && make install
          chdir={{ libreswan_compilation_directory }}
+  async: 300
+  poll: 0
+  register: compile_libreswan
 
 - name: Generate IPsec configuration file
   template: src=ipsec.conf.j2
--- a/playbooks/roles/openvpn/tasks/main.yml
+++ b/playbooks/roles/openvpn/tasks/main.yml
@@ -14,6 +14,13 @@
     - ca
     - server
 
+- name: Generate Diffie-Hellman parameters asynchronously
+  command: openssl dhparam -out {{ openvpn_dhparam }} {{ openvpn_key_size }}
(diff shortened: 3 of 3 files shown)


playbooks/roles/streisand-gateway/tasks/main.yml
```diff
@@ -128,6 +128,12 @@
         state=directory
         recurse=yes
 
+- name: Verify libreswan compilation has completed
```
jlund: Libreswan does not get compiled and installed on AWS and GCE instances, so this task would fail on those providers. I'm also not sure how I feel about a Libreswan-related task being placed outside of the l2tp-ipsec role.

jlund:
This is a great idea. I will run some benchmarks later, but it looks like this could provide a nice speed boost.

playbooks/roles/streisand-gateway/tasks/main.yml
aeibrahim: In that case, place this task at the bottom of the ` streisand/playbooks/roles/l2tp-ipsec/tasks/main.yml`

pjrobertson:
any chances of getting this merged? Probably streisand's biggest hinderance now is the speed of setup.
(I'd also like to see an option to choose which features to install - e.g. if I only want shadowsocks it'd be cool if I could use some config file to tell ansible to only install that)

jlund:
Yes, I just haven't had time to refactor it yet, and I wanted to focus on some higher-priority changes like OpenConnect/AnyConnect. In its current state, this PR is broken on Amazon Web Services and Google Compute Engine.

df-sh:
@pjrobertson if you only want shadowsocks, you can comment out the other roles you dont want

pjrobertson:
I've tried that (although not just for shadowsocks) and it doesn't work fully as it seems there are inter-dependent variables across roles. Maybe that should be another ticket that can be opened

aeibrahim:
I agree with @pjrobertson. I'd too like to be able to run parts of the playbook. I believe we can achieve this using tags. The tags can be tied to the name of the roles, so you can specify which roles you wish to run. For example: 

    ansible-playbook streisand.yml -i inventory --tags=common,openvpn,sslh,monit,ufw,streisand-mirror,streisand-gateway

However, this will require ensuring limited interdependencies between roles so that you can run one or more without having the playbook fail.

jlund:
Yeah, that's also on the list. Being able to run pieces is a prerequisite for the upcoming ability to run Streisand directly on a host itself (i.e. `ansible_connection=local`).

I will be revisiting this pull request and likely making additional tasks asynchronous after that is done.

cpu:
It looks like this PR has grown some conflicts over time. @ghost are you willing to rebase against master or should this PR be closed? Thanks!

cpu:
I'm going to close this PR since there hasn't been any activity. Thanks!

PR #209, opened 2015-11-25
Set firewall rules using ufw.

L-P:
Hi,

I needed this for my server and wondered why it was not done already.  
Does using one of the supported cloud providers forbid configuring the firewall
automatically?

### DIFF
--- a/playbooks/roles/ufw/tasks/main.yml
+++ b/playbooks/roles/ufw/tasks/main.yml
@@ -1,0 +1,56 @@
+---
+- name: Ensure UFW allows SSH
+  ufw: to_port="{{ ssh_port }}"
+       proto="tcp"
+       rule="allow"
+
+- name: Ensure UFW is enabled and denies by default
+  ufw: state="enabled"
+       policy="deny"
+       direction="incoming"
+
+- name: Ensure UFW allows L2TP/IPSec
+  ufw: to_port="{{ item }}"
+       proto="udp"
+       rule="allow"
+  with_items:
+    - "500"
+    - "1701"
+    - "4500"
+  when: l2tp_ipsec_gateway_location
+
+- name: Ensure UFW allows nginx
+  ufw: to_port="{{ nginx_port }}"
+       proto="tcp"
+       rule="allow"
+
+- name: Ensure UFW allows OpenVPN DNS from 10.8.0.1
+  ufw: to_port="53"
+       proto="udp"
+       rule="allow"
+       from_ip="10.8.0.1"
+
+- name: Ensure UFW allows OpenVPN
+  ufw: to_port="{{ openvpn_port }}"
+       proto="{{ openvpn_protocol | lower }}"
+       rule="allow"
+
+- name: Ensure UFW allows Shadowsocks
+  ufw: to_port="{{ shadowsocks_server_port }}"
+       proto="tcp"
+       rule="allow"
+
+- name: Ensure UFW allows stunnel
+  ufw: to_port="{{ stunnel_remote_port }}"
+       proto="tcp"
+       rule="allow"
+
+- name: Ensure UFW allows TOR bridge
+  ufw: to_port="{{ tor_orport }}"
+       proto="tcp"
+       rule="allow"
+
+- name: Ensure UFW allows TOR bfs3 pluggable transport
+  ufw: to_port="{{ tor_obfs3_port }}"
+       proto="tcp"
+       rule="allow"
--- a/playbooks/streisand.yml
+++ b/playbooks/streisand.yml
@@ -15,5 +15,6 @@
     - tor-bridge
     - sslh
     - monit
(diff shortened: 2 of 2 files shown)


playbooks/roles/ufw/tasks/main.yml
```diff
@@ -0,0 +1,72 @@
+---
+- name: Ensure UFW allows SSH (22/TCP)
+  ufw:
```
jlund: Minor quibble, but the convention in the repo is that the module arguments start on the same line:

```
- name: Ensure UFW allows SSH (22/TCP)
  ufw: rule="allow"
       name="OpenSSH"
```

playbooks/roles/ufw/tasks/main.yml
```diff
@@ -0,0 +1,72 @@
...
+    rule="allow"
+    name="OpenSSH"
+
+- name: Ensure UFW is enabled and denies by default
+  ufw:
+    state="enabled"
+    policy="deny"
+    direction="incoming"
+
+- name: Ensure UFW allows L2TP / IPSec (500/UDP)
```
jlund: What do you think about combining the L2TP/IPsec firewall rules into a single task and looping over the ports using `with_items`?

playbooks/roles/ufw/tasks/main.yml
```diff
@@ -0,0 +1,72 @@
...
+    proto="tcp"
+    rule="allow"
+
+- name: Ensure UFW allows TOR bridge (8443/TCP)
+  ufw:
+    to_port="8443"
+    proto="tcp"
+    rule="allow"
+
+- name: Ensure UFW allows TOR bfs3 pluggable transport (57366/TCP)
```
jlund: The port numbers for the obfsproxy pluggable transports are randomized and will be different on every server. The tor-bridge role sets a fact called `{{ tor_obfs3_port }}` that can be used here.

jlund:
No, there are no restrictions, and I think this is a nice addition. AWS and GCE users get network-level port restrictions (e.g. through the EC2 security group that Streisand configures) but I think adding firewall configuration to the server itself is a good idea.

We'll also have to make sure that this does not interfere with the firewall rules that the OpenVPN and L2TP/IPsec roles configure. Those rules have less to do with port restrictions and more to do with allowing masquerading, etc. Have you tried those services on your server where UFW is configured?

L-P:
Done, done and done.

I'm using both OpenVPN and L2TP/IPsec, they work fine.

playbooks/roles/ufw/tasks/main.yml
```diff
@@ -0,0 +1,54 @@
+---
+- name: Ensure UFW allows SSH
+  ufw: rule="allow"
+    name="OpenSSH"
```
jlund: The indentation of the arguments should match.

playbooks/roles/ufw/tasks/main.yml
```diff
@@ -0,0 +1,54 @@
...
+    proto="tcp"
+    rule="allow"
+
+- name: Ensure UFW allows TOR bridge
+  ufw: to_port="{{ tor_obfs3_port }}"
+    proto="tcp"
+    rule="allow"
+
+- name: Ensure UFW allows TOR bfs3 pluggable transport
+  ufw: to_port="57366"
```
jlund: The port numbers for the obfsproxy pluggable transports are randomized and will be different on every server. The tor-bridge role sets a fact called {{ tor_obfs3_port }} that can be used here.

This variable must be used, otherwise this firewall rule will be broken and the obfsproxy pluggable transport will not work.

jlund:
There's a tiny bit of cleanup required with the indentation, and it looks like the obfsproxy prot comment from the last diff got missed (or possibly the relevant change was not saved). Thank you for being willing to help. I'm glad to hear that these new rules will work well the existing iptables configuration.

L-P:
I cleaned up the indentation and used every var I could find.  
That's what you get when you rush things: you do them twice.

It became difficult to test since vars are defined inside roles and they are
not idempotent. Running the playbook twice does not work.

```
TASK: [tor-bridge | Restart Tor so the obfsproxy ports and server fingerprint will be available in the state file, and the hidden service for the Gateway will start running] *** 
failed: [example.com] => {"failed": true}
msg: aa-exec: ERROR: profile 'system_tor' does not exist
```

That's just FYI. I may look into it at some point. I'll just spin up a Vagrant for this.

jlund:
All tasks should be idempotent, and I routinely run the playbook several times against a server while testing. I'm not sure what could be causing that error. It's like the Tor AppArmor profile got removed somehow.

I'll test your fork with these modifications.

L-P:
Ran the thing against a Vagrant. Everything seems okay.

jlund:
Looks good to me as well. Thanks again! This is a really nice feature.

PR #224, opened 2016-01-08
Add new Seoul region for AWS playbook

kiddkai:
AWS just add a new region `ap-northeast-2 (Seoul)`. Add it to the regions list and keep the default region
to Singapore.

### DIFF
--- a/playbooks/amazon.yml
+++ b/playbooks/amazon.yml
@@ -8,14 +8,15 @@
   vars:
     regions:
       "1": "ap-northeast-1"
-      "2": "ap-southeast-1"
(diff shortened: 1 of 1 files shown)


jlund:
Thank you!

jlund:
Unfortunately, it looks like the latest stable release of the boto library doesn't support this new region quite yet. Once it does, we can apply this commit again. I added the new AMI information and its related task [here](https://github.com/jlund/streisand/commit/9d0ba6ed1ff8b9714ff560d55daaa359ea383666) so we should be ready once a new boto version is released.

PR #225, opened 2016-01-08
Simplified instructions for configuring Firefox to use the SOCKS prox…

apollux:
…y on Windows.


jlund:
Thank you! This is much simpler. I'm going to move these instructions into other related sections as well.

PR #251, opened 2016-02-16
(Fix) Add openconnect default vars file

CorbanR:


jlund:
Good catch! Thank you.

PR #261, opened 2016-02-22
Fix Shadowsocks OS X instructions to use correct UI terminology

MinimaxHeadroom:
In the OS X portion of the Shadowsocks instructions, two references to 'taskbar’ are made, however, 'menu bar' is the correct usage. Both instances of the former term were replaced with the latter in order to reduce any ambiguity.


jlund:
Thanks!

Issue #262, opened 2016-02-24
Installing on ubuntu minimal


pjrobertson:
I've just tried to install this on a minimal install of Ubuntu 14.04, which turns out not to have Python installed on it.

Anyway, the very first command that seems to be run over SSH is:

```
ssh -C -vvv -o ControlMaster=auto -o ControlPersist=60s -o StrictHostKeyChecking=no -o Port=XXXX -o KbdInteractiveAuthentication=no -o PreferredAuthentications=gssapi-with-mic,gssapi-keyex,hostbased,publickey -o PasswordAuthentication=no -o User=root -o ConnectTimeout=100 -o ControlPath=/somewhere/.ansible/cp/ansible-ssh-%h-%p-%r 1.2.3.4 'LANG=en_US.UTF-8 LC_ALL=en_US.UTF-8 LC_MESSAGES=en_US.UTF-8 /usr/bin/python'
```

This was failing with the following error:

```
fatal: [1.2.3.4]: UNREACHABLE! => {"changed": false, "msg": "ERROR! SSH Error: data could not be sent to the remote host. Make sure this host can be reached over ssh", "unreachable": true}
```

which was a bit misleading. When I ran the SSH command on its own it was spitting out:

```
bash: /usr/bin/python: No such file or directory
```

which pointed me in the right direction. It should probably first check if the SSH connection works (no command) then if python is installed.

pjrobertson:
I think [this](http://stackoverflow.com/a/22763407/305324) answer gives a method for installing Python using ansible. I'm not sure if @jlund would want to go down the route of shipping a separate script file with the repo for the few cases where ansible is being installed on ubuntu-minimal?

pjrobertson:
OK so to reiterate, for ubuntu-minimal 14.04 you first need to:

* install python with `apt-get install python`

After which, the playbook will run until the task `common : Install Streisand common packages` when it bails with this error:

```
install 'build-essential' 'curl' 'markdown' 'openntpd' 'python-software-properties' 'qrencode' 'unattended-upgrades' 'wamerican-huge'' failed: debconf: delaying package configuration, since apt-utils is not installed
update-default-wordlist: Question empty but elements installed for class \"wordlist\"
  dictionaries-common/default-wordlist: return code: \"0\", value: \"\"
  Choices: , Manual symlink setting
  shared/packages-wordlist: return code: \"10\" owners/error: \"shared/packages-wordlist doesn't exist\"
[...]

PR #265, opened 2016-02-27
Add task to install dictionary packages

pjrobertson:
These aren't installed by default on ubuntu-minimal and as such cause problems when setting up streisand

This is semi-related to #262. There just needs to be a way now to check if Python is installed and if not install it.

### DIFF
--- a/playbooks/roles/common/vars/main.yml
+++ b/playbooks/roles/common/vars/main.yml
@@ -1,5 +1,8 @@
 ---
 streisand_common_packages:
+  # Required for generating passwords (not pre-installed on Ubuntu minimal)
+  - dictionaries-common
+  - wamerican-huge
   # Used to compile Libreswan and OpenConnect Server (ocserv)
   - build-essential
   # Used to perform API requests, including the version check for


playbooks/roles/common/vars/main.yml
```diff
@@ -1,5 +1,8 @@
 ---
 streisand_common_packages:
+  # Required for generating passwords (not pre-installed on Ubuntu minimal)
+  - dictionaries-common
+  - wamerican-huge
```
jlund: The `wamerican-huge` package is already installed (it's the last entry in this alphabetical list). The `dictionaries-common` package is a dependency for `wamerican-huge`. I'm fine with adding `dictionaries-common` to this list, but APT should already be taking care of this. I'm not sure why that isn't the case?

pjrobertson: Interesting. When I was running streisand on an ubuntu-minimal install, apt was bailing on this. I had to uninstall the dictionaries-common package then reinstall it, after which it all worked. It could be that user input was required. I'll try it again and post more info

pjrobertson:
@jlund - I've posted the error in #262

pjrobertson:
...and this PR doesn't actually fix the problem!

PR #287, opened 2016-04-21
Prevent hanging on re-run

gjedeer:
Running this script on certain servers fails a lot of times and requires 4-5 re-executions of ansible-playbook. I'm not trying to fix that.

After one of the re-runs Nginx was already running but client-side docs weren't generated yet, and Ansible hanged on "Make sure that nothing is listening on port 80". Stopping nginx manually allowed Ansible to proceed, so my commit is making this automated.

### DIFF
--- a/playbooks/roles/nginx/tasks/main.yml
+++ b/playbooks/roles/nginx/tasks/main.yml
@@ -3,6 +3,9 @@
   apt: name={{ item }}
        state=absent
   with_items: "{{ apache_packages_to_remove }}"
+  
+- name: Stop nginx for configuration
+  service: name=nginx state=stopped
 
 - name: Make sure that nothing is listening on port 80 on the public interface
   wait_for: host={{ streisand_ipv4_address }}


jlund:
I ended up just removing that check entirely. It was failing in my test case because UFW drops packets to port 80, so the `wait_for` module never actually got the response it was looking for indicating that the port was closed.

On what servers/providers are you seeing that many failures, and what types of errors are occurring? That definitely shouldn't be happening, and I'll do my best to fix it. It works the first time on all of the natively supported providers, and lately I've been doing a lot of experiments on Vultr and on a local VPS provider in my hometown. It works the first time on those as well.

gjedeer:
The issues I had were with Aruba Cloud and errors were mostly with failed file downloads. I tried with two instances, on one it took about five runs and on another it was more.

It's a $1/mo vps, though, so I had no expectations.

On April 24, 2016 3:39:27 AM CEST, Joshua Lund <notifications@github.com> wrote:
>I ended up just removing that check entirely. It was failing in my test
>case because UFW drops packets to port 80, so the `wait_for` module
>never actually got the response it was looking for indicating that the
>port was closed.
>
>On what servers/providers are you seeing that many failures, and what
>types of errors are occurring? That definitely shouldn't be happening,
>and I'll do my best to fix it. It works the first time on all of the
>natively supported providers, and lately I've been doing a lot of
>experiments on Vultr and on a local VPS provider in my hometown. It
>works the first time on those as well.
>
>---
>You are receiving this because you authored the thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/jlund/streisand/pull/287#issuecomment-213864021

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

Issue #292, opened 2016-04-29
SSH config broken by commit 0daa52c


SummerStorm:
I checked out the latest version, ran it, and successfully setup the streisand server with no error messages and all the non-SSH services seems to work. Then I tried to SSH in but couldn't (full SSH log at the bottom).

Thinking maybe I did something wrong, I tried again and setup a second server using the identical steps, and got the identical error.

Since I successfully setup a streisand server without any SSH problems about a month ago, I suspected that maybe some recent changes, i.e. commit 0daa52c, have made the SSH configs overly strict. So I checked out e0b8564 and repeated the setup a third time and third time was the charm.

Here's the the client side SSH log with a single v (let me know if you need `-vvv`):

```
XXX@XXX:~/.ssh$ ssh -v SERVER_NAME
OpenSSH_7.2p2 Ubuntu-1, OpenSSL 1.0.2g-fips  1 Mar 2016
debug1: Reading configuration data /home/XXX/.ssh/config
debug1: /home/XXX/.ssh/config line 13: Applying options for vpn-sing
debug1: Reading configuration data /etc/ssh/ssh_config
debug1: /etc/ssh/ssh_config line 19: Applying options for *
debug1: Connecting to XXX.XXX.XXX.XXX [XXX.XXX.XXX.XXX] port 22.
debug1: Connection established.
debug1: key_load_public: No such file or directory
debug1: identity file /home/XXX/.ssh/streisand_rsa type -1
debug1: key_load_public: No such file or directory
debug1: identity file /home/XXX/.ssh/streisand_rsa-cert type -1
debug1: Enabling compatibility mode for protocol 2.0
debug1: Local version string SSH-2.0-OpenSSH_7.2p2 Ubuntu-1
debug1: Remote protocol version 2.0, remote software version OpenSSH_6.6.1p1 Ubuntu-2ubuntu2.4
debug1: match: OpenSSH_6.6.1p1 Ubuntu-2ubuntu2.4 pat OpenSSH_6.6.1* compat 0x04000000
debug1: Authenticating to XXX.XXX.XXX.XXX:22 as 'forward'
debug1: SSH2_MSG_KEXINIT sent
debug1: SSH2_MSG_KEXINIT received
debug1: kex: algorithm: curve25519-sha256@libssh.org
debug1: kex: host key algorithm: ecdsa-sha2-nistp256
debug1: kex: server->client cipher: chacha20-poly1305@openssh.com MAC: <implicit> compression: none
[...]

PR #293, opened 2016-05-03
- Fixing mode permissions (when using new format, you need to add the 0)

CorbanR:
- Adding back in PAM.
- Ensure all host keys are generated
- Fixing sftp path.
- Fixes https://github.com/jlund/streisand/issues/292


CorbanR:
fixes https://github.com/jlund/streisand/issues/292

PR #297, opened 2016-05-15
Openvpn UDP configurations

alimakki:
I thought it would be nice to have OpenVPN client configs that talked over UDP since they are faster than their TCP equivalents, so I implemented the necessary changes and configured it to run over a non-standard port (UDP port 8757). 

Additionally, OpenVPN binaries are being pulled from the OpenVPN community deb repository rather than Ubuntu's, so the binaries are at the latest version.

An extra directive was added to the client and server templates to set the minimum tls-version to 1.2 for additional hardening of the control channel.

All firewall settings, NATing and monitrc configurations have also been applied where needed.

Cheers,

Ali


jlund:
Hello! Sorry that it has taken me some time to look at this. I appreciate your help, and I think this will make a nice addition to the project! I'll go through and add some in-line comments tonight and tomorrow.

I'm in the middle of some larger changes, and it's likely that some of them will conflict, so we'll probably have one more round of rebasing, addressing the comments, and squashing before we can merge.

playbooks/roles/openvpn/tasks/main.yml
```diff
@@ -1,4 +1,15 @@
 ---
+- name: Import the public GPG key from OpenVPN community
+  apt_key: url=https://swupdate.openvpn.net/repos/repo-public.gpg state=present
```
jlund: Instead of doing it like this and pulling the key from a remote location, it's a little more secure to pin the GPG key in the repo. You can see an example [here](https://github.com/jlund/streisand/blob/master/playbooks/roles/nginx/tasks/main.yml#L7).

playbooks/roles/openvpn/tasks/main.yml
```diff
@@ -1,4 +1,15 @@
...
+
+- name: Adds the OpenVPN community repositroy to sources.list
+  apt_repository: repo='deb http://swupdate.openvpn.net/apt trusty main' state=present
```
jlund: The convention is for each argument to go on a separate line, so `state=present` would be on its own line and indented to match `repo=...`. The other tasks in this file follow the same format.

This file hasn't been converted to multi-line YAML syntax yet (which makes this even more explicit) but it will be soon.

playbooks/roles/openvpn/tasks/main.yml
```diff
@@ -1,4 +1,15 @@
 ---
+- name: Import the public GPG key from OpenVPN community
+  apt_key: url=https://swupdate.openvpn.net/repos/repo-public.gpg state=present
+
+- name: Adds the OpenVPN community repositroy to sources.list
```
jlund: There are a few small typos here. The name description should be something like "Add the official OpenVPN Community repository" in order to match the convention from other roles.

playbooks/roles/openvpn/tasks/main.yml
```diff
@@ -1,4 +1,15 @@
...
+  apt_repository: repo='deb http://swupdate.openvpn.net/apt trusty main' state=present
+
+- name: Ensure the APT cache is up to date
```
jlund: This task isn't necessary. The `apt_repository` module automatically refreshes the cache when a new repository is added.

playbooks/roles/openvpn/tasks/main.yml
```diff
@@ -139,7 +158,11 @@
   template: src=etc_openvpn_server.conf.j2
             dest=/etc/openvpn/server.conf
 
-- name: Restart OpenVPN so the 10.8.0.0 interface is available to dnsmasq
+- name: Copy OpenVPN UDP configuration file into place
+  template: src=etc_openvpn_server_udp.conf.j2
+            dest=/etc/openvpn/server-udp.conf
+
+- name: Restart OpenVPN so the 10.8.0.0/10.9.0.0 interface is available to dnsmasq
```
jlund: Let's rephrase this as "Restart OpenVPN so the TUN interfaces are available to dnsmasq".

playbooks/roles/openvpn/templates/client-common-udp.j2
```diff
@@ -0,0 +1,30 @@
+dev tun
+proto {{ openvpn_protocol_udp }}
```
jlund: This file is almost identical to the `client-common.j2` template. If we pull the `proto` line into the OVPN templates, then we can avoid a whole lot of duplication.

playbooks/roles/openvpn/templates/client-common.j2
```diff
@@ -7,6 +7,7 @@ nobind
 persist-key
 persist-tun
 ns-cert-type server
+tls-version-min 1.2
```
jlund: I like the idea of this, but we'll have to make sure that it won't break existing clients. The official iOS OpenVPN client, for example, didn't work well when we tried to [update to a fancier set of cipher suites](https://github.com/jlund/streisand/commit/cb65bfc676e88b5ca9843c9dcfb8aa3aa3cb3359#diff-d667e36408c9400858277f3ff051f16b) last year.

Which clients did you test this with?

playbooks/roles/openvpn/templates/client-direct-udp.ovpn.j2
```diff
@@ -0,0 +1,3 @@
+client
+remote {{ openvpn_server }} {{ openvpn_port_udp }}
+{% include "client-common-udp.j2" %}
```
jlund: Just in case the other comment wasn't clear, I'm suggesting that if we had the `proto` line in this file (and the other OVPN templates) then we could include `client-common.j2` here and wouldn't have to duplicate a bunch of the same lines in a separate UDP-only file.

playbooks/roles/openvpn/templates/etc_openvpn_server_udp.conf.j2
```diff
@@ -0,0 +1,301 @@
+#################################################
```
jlund: This is also a lot of duplication relative to `etc_openvpn_server.conf.j2`. It would be cool to pull all of the common lines into a shared template that the TCP and UDP server configuration templates could both include. That way they would only need to contain the lines that are different.

playbooks/roles/openvpn/templates/instructions.md.j2
```diff
@@ -151,6 +151,12 @@ It's preferable to configure Ubuntu using the OpenVPN plugin for NetworkManager.
 1. Tap the slider to connect to the OpenVPN server.
 1. Success! You can verify that your traffic is being routed properly by [looking up your IP address on Google](https://encrypted.google.com/search?hl=en&q=ip%20address). It should say *Your public IP address is {{ streisand_ipv4_address }}*.
 
+<a name="UDP-profiles"></a>
+### Alternate unified profiles for access via port UDP port 8757 ###
```
jlund: This should use the `{{ openvpn_port_udp }}` variable so that the documentation will be correct when the default is changed.

"via port UDP port" -> "via UDP port"

jlund:
That's all that I can see for now. Nice work on this so far! I think that people who are able to connect directly to OpenVPN will appreciate having a faster option available, and I'm excited to get it merged.

playbooks/roles/openvpn/templates/client-common.j2
alimakki: Tested with OpenVPN connect on iOS, as well as viscosity on windows and mac os x.

Viscosity was compiled with OpenVPN 2.3.10 at the time, no issues whatsover. iOS worked just fine.

alimakki:
I'll get to work on the changes you suggested, expect a new PR soon.
Thanks for taking the time to review my changes.

Cheers!

alimakki:
At this point all the issues you have brought up have been addressed. 
The playbook was tested on a DO droplet and all roles executed successfully.

Cheers!

playbooks/roles/openvpn/templates/client-common.j2
jlund: Excellent! 1.0.7 just came out today and it supports AES-GCM now too.

PR #303, opened 2016-06-01
added alternative configuration for OpenVPN over UDP

alimakki:
I thought it would be nice to have OpenVPN client configs that talked over UDP since they are faster than their TCP equivalents, so I implemented the necessary changes and configured it to run over a non-standard port (UDP port 8757).

Additionally, OpenVPN binaries are being pulled from the OpenVPN community deb repository rather than Ubuntu's, so the binaries are at the latest version.

An extra directive was added to the client and server templates to set the minimum tls-version to 1.2 for additional hardening of the control channel.

All firewall settings, NATing and monitrc configurations have also been applied where needed.

This PR also addresses the issues raised from my previous PR #297

Cheers,

Ali


jlund:
I just tested this out, and it all works perfectly.

It's really clear from your commits in #297 how much time you spent working on this, understanding the code, and thoroughly testing everything. That level of effort is greatly appreciated. This is honestly one of my favorite pull requests that has ever happened in this project. Thank you!

alimakki:
Thank you Josh, I appreciate your feedback!

I'll look into other ways to contribute, I was thinking of maybe localizing the setup instructions in French and Arabic.

Does markdown have a good way to handle i18n?

jlund:
That would be great. I think that we can just have multiple instruction templates and loop through them. A different Markdown parser may be required, but that would be easy to do.

PR #321, opened 2016-06-04
Amend instructions for the Shadowsocks client on Android

zhaofengli:
Recent versions of the client has adopted Material Design, and this commit
updates the instructions to reflect the changes.

I've removed the part about the CHN routes, as it's no longer on by default.

Here is a screenshot of the new version:
Screenshot


jlund:
Thanks!

PR #322, opened 2016-06-04
further hardening for OpenVPN

alimakki:
Removed nsCertType directive since it is now considered depracated by openvpn, replaced with "remote-cert-tls server" on the client side and "remote-cert-tls client" on the server side

openssl-server-certificate renamed to openssl-certificate to be more generic, added client section for client cert configs to support EKU

Server common name is now randomly generated, validated by the client with verify-x509-name directive

OpenVPN mirror now mirrors latest version (2.3.11), was previously 2.3.10

Changed debian repo for OpenVPN to use https instead of http

Removed comments from etc_openvpn_server_common, was kind of noisy

Tested on OpenVPN connect 1.0.7 on iOS and Viscosity (windows) v1.6.4

(edit: clarifying debian repo change)


playbooks/roles/openvpn/tasks/main.yml
```diff
@@ -54,6 +54,15 @@
   args:
     creates: "{{ openvpn_ca }}.crt"
 
+- name: Generate a random server common name
+  shell: grep -v -P "[\x80-\xFF]" /usr/share/dict/american-english-huge | sed -e "s/'//" | shuf -n 2 | xargs | sed -e 's/ /./g'
```
jlund: Although most people only run the playbook once, it's best if tasks are idempotent and can safely run multiple times. Currently, this task will generate a different random common name on every single run and change the server configuration to match, which could cause problems for clients that have already been configured to expect a certain name.

One pattern that happens in several roles in Streisand is to place random output in a file, and then skip the random generation task if the file already exists. Then a second task registers the output from the first. You can see an example of how this pattern works in the OpenConnect role where a random password is created [here](https://github.com/jlund/streisand/blob/master/playbooks/roles/openconnect/tasks/main.yml#L109) and registered [here](https://github.com/jlund/streisand/blob/master/playbooks/roles/openconnect/tasks/main.yml#L121).

If we followed the same pattern for this, we can ensure that the server name remains constant no matter how many times the Streisand playbook is run against a server.

playbooks/roles/openvpn/tasks/main.yml
```diff
@@ -54,6 +54,15 @@
...
+  register: random_common_name
+
+- debug: msg="OpenVPN randomized common name is {{ random_common_name.stdout }}"
```
jlund: This can be removed now that you've finished testing the changes.

jlund:
Really solid work on this! Everything looks great. I only had the two minor comments above, and then this will be ready to merge.

alimakki:
All done, will squash

PR #323, opened 2016-06-05
Update genesis-amazon to use Elastic IP

johnjeffers:
For EC2 instances with public IP addresses, it’s usually desirable to assign an Elastic IP to the instance to ensure that the public IP does not change.

Without an Elastic IP, the public IP address will change any time the instance is stopped or restarted. If the public IP of the instance changes, it will break all sorts of things.

This update will associate an Elastic IP address to the EC2 instance immediately after instance creation, and assigns the Elastic IP to the streisand_ipv4_address variable for use in the rest of the playbook.


jlund:
Awesome! Thanks for putting this together. This will solve a lot of problems that users were running into when their instance came back up with a different IP.

PR #324, opened 2016-06-06
Unattended upgrades

alimakki:
Unattended upgrades will auto-update: Tor, OpenVPN, shadowsocks-libev and nginx
unattended upgrades will explicitly auto-remove unused dependencies
Unattended upgrades will reboot the server at midnight (depending on where the server location timezone is)
Tor and Nginx deb repos re-configured to use https instead of http

rc.local mode using octet notation (0755 instead of 755)

unattended upgrades can be tested by executing the following command: unattended-upgrade --debug --dry-run

note: tried switching shadowsocks deb repo to https but got errors :(

Edit: OpenVPN repo also uses https, but was configured to do so in another PR I have made: PR  #322).
shadowsocks-libev using https in the deb repo resulted in the following error:

`
W: Failed to fetch https://shadowsocks.org/ubuntu/dists/trusty/main/binary-amd64/Packages  gnutls_handshake() failed: Handshake failed

W: Failed to fetch https://shadowsocks.org/ubuntu/dists/trusty/main/binary-i386/Packages  gnutls_handshake() failed: Handshake failed

`

Invoking the same URL with curl works fine, which leads me to believe that the gnutls library has some sort of bug.

### DIFF
--- a/playbooks/roles/common/files/50unattended-upgrades
+++ b/playbooks/roles/common/files/50unattended-upgrades
@@ -3,6 +3,19 @@ Unattended-Upgrade::Allowed-Origins {
     // ${distro_id} and ${distro_codename} will be automatically expanded
     "${distro_id} stable";
     "${distro_id} ${distro_codename}-security";
+
+    //Autoupdate OpenVPN
+    "Freight:${distro_codename}";
+
+    //Autoupdate Tor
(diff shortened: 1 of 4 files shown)


jlund:
Thanks!

Issue #320, opened 2016-06-04
Assign Elastic IP to AWS EC2 instances


johnjeffers:
For EC2 instances with public IP addresses, it’s usually desirable to assign an Elastic IP to the instance to ensure that the public IP does not change.

Without an Elastic IP, the public IP address will change any time the instance is stopped or restarted. If the public IP of the instance changes, it will break all sorts of things.

This update will associate an Elastic IP address to the EC2 instance immediately after instance creation, and assigns the Elastic IP to the streisand_ipv4_address variable for use in the rest of the playbook.

EDIT: This is now a pull request.

https://github.com/jlund/streisand/pull/323

cottmann:
This addresses the problem of certificates not being valid after the IP changed (ie, instance stopped and started again). It does so with additional cost (for keeping the Elastic IP around when the server is down), and does not help for multi-region deploys (where you might want to use the same certificate). 

My preference is to generate the server certificates with CN=vpn.example.com instead of an IP  and then let the instance use Route 53 to set a CNAME to itself (hope to get around to implement that in the next days). 

So why not make it a choice? 
1. Default as is -- use the IP set by AWS to generate the certificate
2. Use elastic IP (as above)
3. Use DNS name (vpn.example.com)

kydan:
The option of a DNS name for the sll cert would be excellent for my
situation, as I switch ip's frequently.
On Jun 5, 2016 12:40 AM, "Tobias Cottmann" <notifications@github.com> wrote:

This addresses the problem of certificates not being valid after the IP
changed (ie, instance stopped and started again). It does so with
additional cost (for keeping the Elastic IP around when the server is
down), and does not help for multi-region deploys (where you might want to
use the same certificate).

My preference is to generate the server certificates with CN=vpn.example.com
instead of an IP and then let the instance use Route 53 to set a CNAME to
itself (hope to get around to implement that in the next days).

So why not make it a choice?
1. Default as is -- use the IP set by AWS to generate the certificate
2. Use elastic IP (as above)
3. Use DNS name (vpn.example.com)

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/jlund/streisand/issues/320#issuecomment-223765382>, or mute
the thread
<https://github.com/notifications/unsubscribe/ACE5KFCsLI1grWGfIVavc_WFmHji-DLWks5qIaqRgaJpZM4IuDTe>
.

johnjeffers:
[...]

Issue #319, opened 2016-06-04
Easy drop-in replacement config files


pjrobertson:
If deploying multiple servers or re-deploying an existing server, it is sometimes useful to keep the same config files (to avoid users having to change their connection details). I am going through this today as I want to update one server that is running on Debian (what Streisand previously ran on) to Ubuntu, and on another server where I want to get openconnect availability.

It would be great if it was made easier to replace the config files & certs for each connection method in one easy go. My idea is that all config files would be stored in the home folder on the server, and symlinked from the respective places.
e.g.:

```
/etc/openvpn -> /home/streisand/openvpn
/etc/shadowsocks-libev -> /home/streisand/shadowsocks-libev
/etc/nginx -> /home/streisand/nginx
...etc.
```

That way, if I wanted to deploy my settings to a brand new server, or to backup all the existing settings of a current server I would just need to copy over that one `/home/streisand` folder, and everything else would be taken care of.

kydan:
There are a few settings that need to be manually updated to reflect the IP
of the instance. In those cases, some more thought will be required. I
already do some of this programmatically for a few configurations in a node
script I wrote to redeploy on DO, probably not appropriate to add to this
project though.

All that said, I would love to see this happen! I redeploy frequently here
in China and would love to be able to spin new instances with the latest
commits without getting all new creds for everything

jlund:
Are there certain services where this type of thing would be more helpful than others?

If I have to do a search-and-replace on an .ovpn file, for instance, and then reinstall it on my phone, what's the advantage of that versus just installing a new .ovpn file? I'm sure that's a bad example though.

I understand the need to rotate IP addresses on a regular basis while in China. I'll be thinking about some ways to help make that easier.

pjrobertson:
I think you’re referring more to rotating IP addresses?
My thoughts are for when you want to upgrade a server/use the same credentials across servers. So for example if I re-installed Streisand on a server and could keep the OpenVPN/Shadowsocks config files/certs all the same, my clients would not need to re-download/install the profiles.

Think for example where you have a server with 20 clients using Shadowsocks. My server is currently running an older version of Streisand and so doesn’t have Openconnect on it. I want to install Openconnect (and all the HTML documentation files that go with it from Stresiand) by upgrading Streisand, but I don’t want my existing 20 clients using Shadowsocks to be affected. It would be great if I could just drop the old config files/Shadowsocks HTML files back in place (in a specific folder)

kydan:
[...]

PR #325, opened 2016-06-06
Create option to use FQDN for streisand server certificates

cottmann:
Hi @jlund ,

I created an option to use the domain name as CN in the certificates (see also #320 for discussion).

**Problem:** 
The default setting in Streisand is to create server certficates
based on the IP address. This creates error messages when the IP
changes.

The IP might change for example on an AWS EC2 instance
that is stopped and restarted later, or moved to another region.
To avoid the hassle of re-creating and distributing certificates,
a server should be able to have a DNS entry (ie, vpn.example.com).

**Proposed solution:** 
This edit creates this option. A new variable
   `streisand_cn_name`
can be defined in
   `playbooks/roles/common/vars/main.yml`

If it is defined, the CN name of the certificates will change to
whatever domain name streisand_cn_name is set to.

If it is not defined, there is no change in behaviour.

**CAUTION:**
- Only tested on openconnect and AWS.
- The documentation might also need adaptation (currently a mix of IP and hostname.
- Did not touch TOR.

### DIFF
--- a/playbooks/roles/common/tasks/set-default-variables.yml
+++ b/playbooks/roles/common/tasks/set-default-variables.yml
@@ -36,3 +36,18 @@
   set_fact:
     streisand_server_name: "{{ ansible_hostname }}"
   when: streisand_server_name is not defined
+
+- name: Set the streisand_cn_name variable to streisand_ipv4_address if it is not defined or empty
+  set_fact:
+    streisand_cn_name: "{{ streisand_ipv4_address }}"
+  when: (streisand_cn_name is undefined) or (streisand_cn_name == "")
+
+- name: Set ocserv_is_dyndns to true 
+  set_fact:
+    ocserv_is_dyndns: true
+  when: streisand_cn_name != streisand_ipv4_address
+  
+- name: Set ocserv_is_dyndns to true 
+  set_fact:
+    ocserv_is_dyndns: false
+  when: streisand_cn_name == streisand_ipv4_address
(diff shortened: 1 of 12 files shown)


jlund:
I think this is a good start as a possible approach to solving #319, but we'll want more people to test it out first with the full suite of services. I'll try to find some time to do that this week.

kydan:
I think this would also help solve issues that I run into when switching IP's (as noted in #319 ) such as no longer being able to connect to the nginx site, since the config is gimped expecting the old IP. This actually would help Streisand to even more ephemeral, though somewhat at the cost of privacy.

 However, personally I dont see a direct IP being all that more "private" when it comes to the authorities, given that if only one to a handful of people are connecting to a particular IP that has been identified as a VPN/censorship-runaround endpoint, vs a large VPN service makes for a much smaller pool of IP's to trace. fairly anonomous domains can be created, and having this as an option to speed up and reduce friction for getting new endpoints up as quickly as possible would be nice.

jlund:
Now that the 16.04 upgrade is done, I'm going to revisit this and see if I can come up with some ways to make it easy to swap amongst instances and IPs.

cpu:
This PR has grown some significant conflicts over time. @cottmann are you interested in rebasing this PR against master and resubmitting?

cpu:
I'm going to close this PR since there hasn't been any reply so far. Thanks!

I'm planning to start on a rework of the Streisand TLS roles that will eventually accommodate this usage case as well.

PR #326, opened 2016-06-07
Additional OpenVPN hardening

alimakki:
Removed nsCertType directive since it is now considered depracated by openvpn, replaced with "remote-cert-tls server" on the client side and "remote-cert-tls client" on the server side

openssl-server-certificate renamed to openssl-certificate to be more generic, added client section for client cert configs to support EKU

Server common name is now randomly generated, validated by the client with verify-x509-name directive

OpenVPN mirror now mirrors latest version (2.3.11), was previously 2.3.10

Changed debian repo for OpenVPN to use https instead of http

Removed comments from etc_openvpn_server_common, was kind of noisy

Tested on OpenVPN connect 1.0.7 on iOS and Viscosity (windows) v1.6.4

(edit: clarifying debian repo change)

Addresses comments raised in PR #322.

openvpn server common name is now limited to 64 characters in length.


jlund:
Thank you very much!

Issue #176, opened 2015-07-06
GCE on OSX problems


df-sh:
Trying to use the GCE option, adding the required libcloud, following the ansible instructions on making a client ID and generating a pk12 file. ansible problem 1:(the link gives a malformed command as the generated pk12 file is in a different format than the command input suggests..)

The command expects a .pkcs12 but the google key download is a .p12 file first and foremost.

After figuring out how to form the command correctly...

Streisand asks for the location of the pem file... I suggest having the default location of the PEM file be the streisand home directory...  I got around my lack of filesystem naming by dragging the PEM file into terminal directly ...

After this fun step... the service email and project id go smoothly...

and then I get a CA Cert doesnt exist error...  I tried using brew to install as the error link suggests... but doesnt work on my version of osx?   Too tired to think of a way for this to work at the moment..

CorbanR:
Did you ever figure this out? Or are you still having issues? I apologize for the late reply.

Brendan-Kai:
Hi CorbanR

I'm having the same issue, could you help.

CorbanR:
@Brendan-Kai @df-sh  I have submitted a pull request that should fix / improve the issues you are having with GCE.

alimakki:
I found that the .json authentication file works with GCE

jlund:
This should be fixed now. Thanks, @CorbanR!

PR #336, opened 2016-06-12
GCE improvements/updates

CorbanR:
- Updating GCE Playbook (Use credentials file)
- Fixes https://github.com/jlund/streisand/issues/176
- Adding us-east zones
- Removing custom gce.py (no longer requried)
- Bumping up instance size to g1-small


jlund:
Thanks!

PR #350, opened 2016-06-21
Add img.shields.io badges

ktaranov:


jlund:
These badges are cool but I am wary of adding elements to the README that could allow a third-party to discern who is looking at the project. Sorry that I didn't see this until now, and thank you for the submission!

Issue #355, opened 2016-06-29
SSL issue when downloading libreswan


mertdumenci:
Hey,

I'm trying to set up streisand on Digital Ocean, and it's giving me an SSL validation error when downloading libreswan. I tried 3 different data centers for the Droplet, and libreswan's downloads seem to be fine (I can fetch them locally just fine.)

Here's the exact error:

```
TASK [l2tp-ipsec : Retrieve the Libreswan source code] *************************
fatal: [{omitted}]: FAILED! => {"changed": false, "failed": true, "msg": "Failed to validate the SSL certificate for download.libreswan.org:443. Make sure your managed systems have a valid CA certificate installed. If the website serving the url uses SNI you need python >= 2.7.9 on your managed machine or you can install the `urllib3`, `pyopenssl`, `ndg-httpsclient`, and `pyasn1` python modules to perform SNI verification in python >= 2.6. You can use validate_certs=False if you do not need to confirm the servers identity but this is unsafe and not recommended. Paths checked for this platform: /etc/ssl/certs, /etc/pki/ca-trust/extracted/pem, /etc/pki/tls/certs, /usr/share/ca-certificates/cacert.org, /etc/ansible"}
```

Thank you!

TonyZTan:
I am experiencing the same issue here. According to [SSL Labs](https://www.ssllabs.com/ssltest/analyze.html?d=download.libreswan.org&s=193.110.157.101), the server certificates are fine, but the server does require SNI support.

TonyZTan:
I submitted a patch (#356) that should fix this issue.

denoza:
Thanks for the fix @TonyZTan. It worked for me!

mertdumenci:
This fixed my issue as well. Thank you, @TonyZTan.

jlund:
And now it's merged! I'm kind of surprised this hasn't come up before. SNI is quite common. Thanks again, @tonyztan!

nabilfreeman:
Hi guys, I'm unfortunately still receiving this exact error on the up-to-date version of `master` as of 2016-09-27.

I looked through the commit history since PR #356 and noticed that the `pip` fix has been moved around the repo a few times and that the download location of Libreswan has changed from a Cloudfront server to the official website. Is anybody else experiencing this error, and could anybody please advise a fix?

In the meantime, I rolled back to @tonyztan's commit and got a server successfully set up, but I'm 70 commits behind master & worried I'm missing important security fixes.

jlund:
[...]

PR #356, opened 2016-06-30
Fixes issue #355

tonyztan:
Fixes issue #355. Installs python-pip and the necessary python modules to enable SNI support and access the server hosting Libreswan source code.

### DIFF
--- a/playbooks/roles/l2tp-ipsec/tasks/main.yml
+++ b/playbooks/roles/l2tp-ipsec/tasks/main.yml
@@ -5,6 +5,12 @@
 
 - name: Install xl2tpd
   apt: name=xl2tpd
+  
+- name: Install python-pip
+  apt: name=python-pip
+  
+- name: Install the SNI dependencies that are required to access Libreswan source
+  shell: pip install urllib3 pyopenssl ndg-httpsclient pyasn1
 
 - name: Retrieve the Libreswan source code
   get_url: url=https://download.libreswan.org/{{ libreswan_source_filename }}
--- a/playbooks/roles/l2tp-ipsec/tasks/main.yml
+++ b/playbooks/roles/l2tp-ipsec/tasks/main.yml
@@ -5,12 +5,13 @@
 
 - name: Install xl2tpd
   apt: name=xl2tpd
-  
+
 - name: Install python-pip
   apt: name=python-pip
-  
+
 - name: Install the SNI dependencies that are required to access Libreswan source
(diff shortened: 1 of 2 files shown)


playbooks/roles/l2tp-ipsec/tasks/main.yml
```diff
@@ -5,6 +5,12 @@
 
 - name: Install xl2tpd
   apt: name=xl2tpd
+  
+- name: Install python-pip
+  apt: name=python-pip
+  
+- name: Install the SNI dependencies that are required to access Libreswan source
+  shell: pip install urllib3 pyopenssl ndg-httpsclient pyasn1
```
jlund: We should use the `pip` Ansible module here instead.

tonyztan:
@jlund I have modified the task to use the `pip` Ansible module. Thank you for this great project, by the way!

jlund:
You're welcome! Thank you for your help!

tonyztan:
@jlund I noticed that the SNI support packages seem to be causing problems for you [here](https://github.com/jlund/streisand/commit/43696dea52bb1a4aaad6cd53dd8dd0ceabf34403). Did you run into the issue while deploying streisand on a new server? That's very strange, because when I tested the code with DigitalOcean & Amazon EC2 everything worked just fine, including the `streisand-mirror` role.

jlund:
@tonyztan Yeah, when I changed the order to install the pip packages in the `common` role (thinking that more things are going to be using SNI, and it should therefore be more generic) it broke the `openconnect` role. Specifically, Ansible's `get_url` module couldn't retrieve the OpenConnect source code from the CloudFront CDN anymore and was failing with a Python "connection reset by peer" error message.

I then moved the SNI-related tasks back to the L2TP/IPsec role, but soon remembered that CloudFront is also a mirror location for some of the client downloads. A few of the other mirrors were also failing too.

I'm realizing now that one of the nitpicky changes that I made was to alphabetize the `pip` package list. I wonder if the order of installation actually matters for the SNI-related Python packages? That would be kind of silly, but everything about SNI on Trusty's version of Python seems pretty silly.

I eventually just decided to punt and temporarily mirror the Libreswan source code (which is what needed the SNI support in the first place) on the CloudFront CDN, thereby bypassing the problem for now.

The version of Python that ships with Ubuntu 16.04 (Xenial) natively supports SNI, and I'm hoping to get that upgrade mostly finished up this weekend.

PR #359, opened 2016-07-10
Update Android and Windows Shadowsocks mirrors

### LINKED ISSUE #349: Barcode Scanner (apk) should be mirrored for Android Shadowsocks
Shadowsocks for Android requires the 'Barcode Scanner' app to be installed in order to scan the QR Codes generated and displayed in the Streisand HTML docs.

I assume the Shadowsocks Android APK is mirrored by the Streisand server since it is sometimes not possible for people without a VPN to download from the Google Play Store (e.g. in China).
Still, since Bacrode Scanner is also required for the Shadowsocks APK to work fully, a connection to the Play Store is still required at some point, defeating the purpose of the mirrored file.

It would be great if the apk could also be mirrored and instructions added along the lines of "If you want to scan the QR Code, then also install Barcode Scanner alongside Shadowsocks"

A mirror to the Barcode Scanner APK: https://github.com/zxing/zxing/releases

pjrobertson:
Latest versions as obtained from GitHub

Fixes #349  because it turns out v2.10.5 of the Android app has an integrated barcode scanner

### DIFF
--- a/playbooks/roles/streisand-mirror/vars/shadowsocks.yml
+++ b/playbooks/roles/streisand-mirror/vars/shadowsocks.yml
@@ -5,18 +5,18 @@ shadowsocks_mirror_location: "{{ streisand_mirror_location }}/shadowsocks"
 shadowsocks_mirror_href_base: "/mirror/shadowsocks"
 
 # Android
-shadowsocks_android_version: "2.9.5"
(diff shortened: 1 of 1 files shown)


jlund:
Thanks!

PR #361, opened 2016-07-20
Use environment variable as default for DO token

DavidWittman:
Default the `do_access_token` to the environment variable
`DIGITALOCEAN_API_KEY`. This variable is used by doctl and other
tools and is often exported in the environment of a user which
leverages these tools.

The result is a little bit ugly in the prompt format, but it does
make provisioning easier from a user perspective:

```
If you enter nothing here, the environment variable DIGITALOCEAN_API_KEY will be used.

What is your DigitalOcean Access Token?
[{{ lookup('env', 'DIGITALOCEAN_API_KEY') }}]:
```

### DIFF
--- a/playbooks/digitalocean.yml
+++ b/playbooks/digitalocean.yml
@@ -44,7 +44,8 @@
       private: no
 
     - name: "do_access_token"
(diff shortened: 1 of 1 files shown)


DavidWittman:
Alternatively, if you'd rather not have to see the templated garbage there in the prompt, we can default this to the environment variable later on in the `genesis` role.

jlund:
This is a great idea that is already saving me time as I iterate through changes. I ended up implementing this in the alternative way that you suggested. Thanks!

PR #362, opened 2016-07-20
Use wait_for instead of pause when building DO

DavidWittman:
Rather than twiddling our thumbs for 90 seconds, the `wait_for`
module will poll port 22 until it is ready for connections. This
is faster and more resilient than using the `pause` module here.

Also add *.retry files to the .gitignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 generated-docs
+*.retry
--- a/playbooks/roles/genesis-digitalocean/tasks/main.yml
+++ b/playbooks/roles/genesis-digitalocean/tasks/main.yml
@@ -52,6 +52,10 @@
   set_fact:
     streisand_server_name: "{{ do_server_name }}"
 
-- name: New DigitalOcean servers are occasionally slow to process incoming SSH connections even after the OpenSSH daemon has started up. Pause for 90 seconds.
-  pause:
-    seconds: 90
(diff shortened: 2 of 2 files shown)


jlund:
The `wait_for` module is (and was) [already being used](https://github.com/jlund/streisand/blob/master/playbooks/roles/genesis-digitalocean/tasks/main.yml#L35). Your proposed commit adds it for a second time, twenty lines below the original instance :)

Earlier this year, DigitalOcean was failing to accept incoming connections on brand-new droplets in a timely fashion. Despite the OpenSSH daemon reporting that it was ready, connection attempts would just sit there and hang. Notably, this occurred with or without Ansible, and it also happened across multiple regions. The issue (whatever it was) would clear up around a minute later, but the resulting timeouts were completely breaking the playbook.

In other words, the janky 90-second `pause` was added to deal with this DigitalOcean jankiness. Maybe it isn't janky anymore? Let me check.

jlund:
It looks like they fixed it, and I removed the `pause` task. Thanks for bringing it up and reminding me to revisit this step! It's much, much nicer without the artificial delay, and its presence was definitely confusing. I would have made exactly the same assumption that you did.

DavidWittman:
Whoops, apologies for missing that one, Josh. Thanks!

Issue #311, opened 2016-06-02
IPsec/L2TP on AWS


trexx:
"Streisand does not install L2TP/IPsec on Amazon EC2 or Google GCE servers by default because the instances cannot bind directly to their public IP addresses which makes IPsec routing nearly impossible."

IPsec does not need to bind directly to the public IP.
For Amazon anyway, left should be the private IP of the instance, leftid should be the Elastic IP associated with said instance. Use the Elastic IP in secrets for matching PSK and tunnels will establish. L2TP should also naturally work once the tunnel is established. Obviously ensure ports 500, 4500 and IP Protocol 50 (ESP) are open in security groups.
I managed a similar setup and it worked (until we replaced it with OpenVPN)

left={{ ansible_default_ipv4.address }}
leftid={{ ansible_default_ipv4.address }}

DavidWittman:
:+1: I can confirm this configuration works just fine on AWS.

jlund:
Merged! Thanks, @DavidWittman.

joopdorresteijn:
Not sure if solved.

```
Jan 15 01:10:57 almond pluto[14007]: added connection description "nm-ipsec-l2tp-13411"
Jan 15 01:10:57 almond pluto[14007]: "nm-ipsec-l2tp-13411" #1: initiating Main Mode
Jan 15 01:10:57 almond NetworkManager[410]: 002 "nm-ipsec-l2tp-13411" #1: initiating Main Mode
Jan 15 01:10:57 almond NetworkManager[410]: 104 "nm-ipsec-l2tp-13411" #1: STATE_MAIN_I1: initiate
Jan 15 01:10:57 almond pluto[14007]: "nm-ipsec-l2tp-13411" #1: transition from state STATE_MAIN_I1 to state STATE_MAIN_I2
Jan 15 01:10:57 almond NetworkManager[410]: 002 "nm-ipsec-l2tp-13411" #1: transition from state STATE_MAIN_I1 to state STATE_MAIN_I2
Jan 15 01:10:57 almond NetworkManager[410]: 106 "nm-ipsec-l2tp-13411" #1: STATE_MAIN_I2: sent MI2, expecting MR2
Jan 15 01:10:57 almond pluto[14007]: "nm-ipsec-l2tp-13411" #1: STATE_MAIN_I2: sent MI2, expecting MR2
[...]

PR #370, opened 2016-08-03
Add l2tp/ipsec to AWS builds

DavidWittman:
As mentioned in #311, L2TP/IPSec works without issue on AWS. This commit adds the `l2tp-ipsec` (and the associated security group roles) back into the Amazon playbook.

Edit: I'd imagine this can also be applied to GCE, I just haven't got around to it.


jlund:
OMG you have no idea how many hours I spent two summers ago trying to figure out how to make L2TP/IPsec happen with Strongswan on AWS. I don't know if I have ever been more frustrated in my entire life. My hands were bruised from all of the facepalms.

Thinking about the fact that this has just been sitting there ready to work on AWS since the switch to Libreswan is making me laugh quite a bit.

I think you're right that this means we're also good to go on GCE. I'll give that a shot now. Thanks!

PR #371, opened 2016-08-03
[AWS] Add VPC support

DavidWittman:
Allows the user to optionally provide VPC and Subnet IDs when creating
a Streisand server. These values are then passed when creating security
group rules and the EC2 instances later on in the build process.

### DIFF
--- a/playbooks/amazon.yml
+++ b/playbooks/amazon.yml
@@ -39,6 +39,20 @@
       default: "2"
       private: no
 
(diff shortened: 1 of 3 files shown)


jlund:
Maybe your best one yet!

My account didn't have EC2 Classic, so I was never able to properly troubleshoot this and wasn't sure how to best fix it either. This is a really big deal. Thank you!

PR #373, opened 2016-08-03
[AWS] Use ec2_ami_find for determining AMI ID

DavidWittman:
Rather than enumerating the available AMIs for each region, we can
leverage the `ec2_ami_find` module (introduced in Ansible 2.0) to find
the appropriate image based on the selected region.

This simplifies the code a bit, ensures that we're always using the
latest Ubuntu 14.04 image provided by Amazon, and automatically accounts
for new regions.

Possible cons are the added time it takes for the lookup (though
relatively neglibile) and the lack of determinism in the AMI which
is ultimately selected.


jlund:
This is a very nice module and much better than the old approach. Thanks!

PR #374, opened 2016-08-03
[AWS] Add autorecover alarm

DavidWittman:
This task configures a CloudWatch metric alarm to trigger auto-recovery
if the host system is down for 2 minutes.

AWS (via CloudWatch) supports a feature known as auto-recovery, which
will respawn an instance on host system failure. The recovered instance
is identical to the original instance, including instance ID,
private/elastic IP addresses, and metadata.

Currently this feature is only available in the `us-east-1` region for
the C3, C4, M3, R3, and T2 instance types.

@jlund Sorry for all the PRs. I know it's a lot of work maintaining these things... I'd be happy to help in any way that I can.

### DIFF
--- a/playbooks/roles/genesis-amazon/tasks/main.yml
+++ b/playbooks/roles/genesis-amazon/tasks/main.yml
@@ -39,6 +39,28 @@
     wait: yes
   register: streisand_server
 
+# The auto-recover action is currently only available in us-east-1
+- name: Create CloudWatch alarm to auto-recover instance (us-east-1 only)
+  ec2_metric_alarm:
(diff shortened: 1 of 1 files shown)


jlund:
Ha, no need to apologize. This is great stuff, and I really appreciate your contributions!

I will find time to review these and get them merged as soon as I can. Had I noticed them earlier, I would have done it tonight instead of working on the obfs4 and Xenial upgrade :)

jlund:
This is a really cool feature. Hopefully they will roll it out to other regions soon.

PR #378, opened 2016-08-11
Include streisand.yml in amazon.yml

DavidWittman:
Adding `ansible_user` and `ansible_become` as host variables with the
`add_host` module satisfy the login requirements which are needed for
connecting to EC2 instances, and allows us to include the main play
from streisand.yml.


jlund:
This cleans things up very nicely!

PR #379, opened 2016-08-11
Add l2tp/ipsec support to GCE builds

### LINKED ISSUE #311: IPsec/L2TP on AWS

DavidWittman:
Just as in #370, adding support for L2TP/IPsec, but for Google Compute Engine this time around.

Closes #311

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,6 @@ Services Provided
 * L2TP/IPsec using [Libreswan](https://libreswan.org/) and [xl2tpd](https://www.xelerance.com/software/xl2tpd/)
   * A randomly chosen pre-shared key and password are generated.
   * Windows, OS X, Android, and iOS users can all connect using the native VPN support that is built into each operating system without installing any additional software.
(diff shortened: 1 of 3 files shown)


jlund:
@DavidWittman Sorry about the delay in getting all of these merged. I have been off the grid for the past two weeks.

I really can't thank you enough for all of the major improvements that you have contributed recently. I hope that you continue to participate in the project!

Issue #386, opened 2016-08-23
Certificate details do not match with Manual SSL verification section


suvil:
Client OS - OS X 10.11.5
Followed all instructions in the readme.
Tried with two providers - Twice with AWS and once with Digital Ocean.

Every single time there's difference in the certificate details shows by keychain access and the information in the manual ssl verification system in the generated docs. 

One example is here - 

![screenshot](https://cloud.githubusercontent.com/assets/1581200/17886561/02d14604-6940-11e6-80b6-df0feb589687.png)

I am still understanding ansible but this disparity is perplexing.

jlund:
**TL;DR** You pick one or the other (install the CA certificate, or manually verify fingerprints)--not both.

## Nerdy (and likely boring) Longer Explanation

There are two different (but complementary) certificates in your screenshot, and they therefore have different fingerprints:

1. The **Server** certificate.
2. The **CA** certificate.

The fingerprints for number 1 are in the connection instructions on the left side of your screenshot. The fingerprints for number 2 are in the OS X Keychain details on the right side of your screenshot.

If you are _not_ installing the **CA** certificate, then you can verify the **Server** certificate manually. The fingerprints in the instructions file are there for that purpose.

However, if you _are_ installing the **CA** certificate, then manual fingerprint verification isn't necessary. The **Server** certificate is signed by the **CA** certificate, and when you install the **CA** certificate the **Server** certificate passes verification.

If you want to see the **Server** certificate for manual verification to compare fingerprints with the instructions, you can either do that via your browser's interface (e.g. clicking on the lock icon and viewing the SSL connection details) or you can also use OpenSSL on the command line:

`openssl s_client -connect <ip-address-goes-here>:443 | openssl x509 -noout -fingerprint -sha256`

Both your browser and the OpenSSL command above will show you **Server** fingerprint details that should match what you see in the HTML connection instructions. The **CA** and **Server** fingerprints are expected to be different, because they are different. The fingerprint verification is only necessary if you're skipping the **CA** certificate installation steps, and comparing fingerprints manually is pretty tedious which is why the CA installation is recommended.

SSL is super confusing. I hope this explanation helps!

PR #395, opened 2016-09-05
Add documentation on verifying certificate authority

CorbanR:
@jlund 

So.. there seems to be some confusion when people install the certificate authority vs verifying just the server certificate. I'm not sure if adding this documentation will be helpful, or more confusing.. 

My thought is if anyone runs into this again https://github.com/jlund/streisand/issues/386, the added documentation would be useful. Having said all that.. please reject if you think it would be more confusing.


CorbanR:
I don't think this is accomplishing what I wanted, closing.

PR #400, opened 2016-09-13
add shadowsock OTA, enable fast_open

yuanl:
Add [One Time Auth](https://shadowsocks.org/en/spec/one-time-auth.html) support. disable by default for backward compatible.
Enable TCP fast open.


jlund:
I think the QR code needs to change as well if this is enabled. OTA seems to be well-supported across most Shadowsocks clients too, so it might make sense to enable it by default. I'll do some testing on this tomorrow.

yuanl:
OTA do help improve to stablebility of shadowsocks by hiding it more deeply. VPS IP will get blocked by Chinese gov very quickly without it.

yuanl:
If we decide to enable it by default. The QR code could be change as to https://shadowsocks.org/en/config/quick-guide.html with ease.

jlund:
Yeah, it looks pretty straightforward and it's also more secure. At this point I'm leaning heavily towards making this the default, especially if it will help with things like issue #413.

yuanl:
To my knowledge, OTA was design specify for solving that GFW machine learning thing.

denoza:
Hi, can this be used by cloning the main branch? I would like to redeploy my server using this script, in the case if it is more robust against the GFW.

yuanl:
Not yet, this PR do not enable OTA by default.



On Tue, Oct 4, 2016 at 9:30 PM +0800, "denoza" <notifications@github.com> wrote:












Hi, can this be used by cloning the main branch? I would like to redeploy my server using this script, in the case if it is more robust against the GFW.

jlund:
OTA is now enabled by default. Thank you for all of your help with this, @yuanl !

yuanl:
COOL!

geligelu:
Just read this! Made a few comments over on Issue #413 - rolling a new server now to see how quickly I get blocked!

hydrandt:
@yuanl, could you please paste here a link to an article/discussion (in Chinese or English) which explains how does OTA help against GFW? Reading the specification, I could not understand how it helps.

yuanl:
There was a long discuss about the weakness of shadowsocks protocol https://github.com/breakwa11/shadowsocks-rss/issues/38 it may have same discus of OTA somewhere in the thread. 

Also check the link in reply of @madeye https://github.com/shadowsocks/shadowsocks-libev/issues/428

PR #401, opened 2016-09-13
Add ShadowsocksX-NG to streisand-mirror

yuanl:
OSX client ShadowsocksX-NG is now recommended on [shadowsocks.org](https://shadowsocks.org/en/download/clients.html).
However, [it require OSX 10.10+](https://github.com/shadowsocks/ShadowsocksX-NG/releases)

### DIFF
--- a/playbooks/roles/shadowsocks/templates/instructions.md.j2
+++ b/playbooks/roles/shadowsocks/templates/instructions.md.j2
@@ -27,9 +27,9 @@ Shadowsocks
 
 <a name="osx"></a>
 ### OS X ###
-1. Download [ShadowsocksX](/mirror/#shadowsocks).
+1. Download [ShadowsocksX-NG](/mirror/#shadowsocks).
 1. Double-click the DMG, and drag the icon into your Applications folder.
-1. Launch ShadowsocksX. You will be prompted to enter your password so that your system proxy settings can be modified.
+1. Launch ShadowsocksX-NG. You will be prompted to enter your password so that your system proxy settings can be modified.
 1. Look for the Shadowsocks paper plane icon in your menu bar and click on it.
 1. Make sure the QR code below is centered and completely visible, and choose *Scan QR Code from Screen...*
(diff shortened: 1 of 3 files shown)


pjrobertson:
Awesome! I didn't realise the OS X client was finally being re-developed :)

I would hold off merging this until:

-[] the new .app is signed by a developer certificate
-[]  security issues are addressed (e.g. should `proxy_conf_helper` be installed in `~/Library` as opposed to `/Library`
-[] support for the new app is proven (there was a flurry of work for 3 months until Aug, but for the last month nothing has been done)

jlund:
This supports the Shadowsocks OTA functionality, which is probably the single biggest missing feature in the legacy OS X application. This might be a big enough reason alone to make the switch given the many benefits of enabling OTA support by default.

pjrobertson:
Cool, that's a good feature I didn't know about. I've just checked and there has been work over the past few weeks so I've marked my 3rd points as 'complete' since it does seem it's still being actively developed.

jlund:
Would either of you be willing to help update the OS X connection instructions for ShadowsocksX-NG and submit a PR? I don't currently have a Mac, though I may pick one up tomorrow to get this done and enable OTA by default.

yuanl:
I can't access my Mac right now. But, let me try this on my newly build Hackintosh.

yuanl:
Done, pretty much the same except the name and OTA.

jlund:
I just merged this in as the default client to mirror. I think the points that you made are valid, @pjrobertson, and hopefully they will be addressed soon. However, the fact that this new client supports One-time Authentication and is being actively maintained (unlike the legacy client it replaces which hasn't been updated in a _very_ long time) makes this the right move overall IMO.

OS X users can still skip this wrapper and just use the libev client from Homebrew directly, but most people are going to have a better experience using the NG client.

jlund:
@yuanl Thanks for this one too!

PR #410, opened 2016-09-25
Don't install xmlto recommends. Fixes #404

### LINKED ISSUE #404: `xmlto` dependency for libreswan installs texlive
It seems that (at least in my case, Ubuntu 16.04.1) installing the libreswan dependencies takes a very long time.

I've narrowed this down to the fact that `xmlto` 'suggests' `dblatex` to be installed, which then leads to the whole `latex` suite being installed. 

A quick `apt-cache depends xmlto` shows this, and after installed if I run `apt-cache --instaled depends xmlto` I can confirm that `dblatex` is in there.

Something should be done to make sure `xmlto` is installed without `dblatex` being installed.

This is untested, but I think it'd work:

``` patch
diff --git a/playbooks/roles/l2tp-ipsec/tasks/main.yml b/playbooks/roles/l2tp-ipsec/tasks/main.yml
index 761d61f..324e8b8 100644
--- a/playbooks/roles/l2tp-ipsec/tasks/main.yml
+++ b/playbooks/roles/l2tp-ipsec/tasks/main.yml
@@ -2,6 +2,7 @@
 - name: Install the Libreswan dependencies that are required for compilation
   apt:
     name: "{{ item }}"
+    install_recommends: "{{ yes if item = 'xmlto' else no }}"
   with_items: "{{ libreswan_compilation_dependencies }}"

 - name: Install xl2tpd
```

pjrobertson:
xmlto recommends includes `dblatex` which in turn installs `latex` - a huge dependency.

### DIFF
--- a/playbooks/roles/l2tp-ipsec/tasks/main.yml
+++ b/playbooks/roles/l2tp-ipsec/tasks/main.yml
@@ -2,6 +2,7 @@
 - name: Install the Libreswan dependencies that are required for compilation
   apt:
     name: "{{ item }}"
+    install_recommends: "{{ 'yes' if item == 'xmlto' else 'no' }}"
   with_items: "{{ libreswan_compilation_dependencies }}"
 
 - name: Install xl2tpd


pjrobertson:
Forget this for the time being...

jlund:
This sounds promising. Those dependencies do take a long time to install, relatively speaking.

Issue #407, opened 2016-09-20
add apt-transport-https to prereqs


saveyour:
the package apt-transport-https is needed on ubuntu for:

TASK [openvpn : Add the official OpenVPN repository] ***************************

also something people building this in a lxc container might run into is missing a tun/tap device. that can be solved by following these steps:  http://heider.io/blog/2013/10/26/openvpn-in-a-lxc-container/

thanks, keep up the good work

antagonismorg:
Should this be closed as the PR has been merged?

DavidWittman:
Closed by #412

PR #412, opened 2016-09-25
apt-transport-https

alimakki:
Ensures that apt-transport-https is installed as some cloud providers don't image their distros with it installed by default. Addresses #407


jlund:
Thanks!

PR #419, opened 2016-10-05
Ensure AppArmor dependency

### LINKED ISSUE #418: apparmor is never installed
```
TASK [tor-bridge : Restart AppArmor so the abstraction takes effect] ***********
fatal: [163.172.146.75]: FAILED! => {"changed": false, "failed": true, "msg": "Could not find the requested service \"'apparmor'\": "}
```

It seems that apparmor is never installed. Solved by ssh'ing into the server and doing `apt install apparmor`. This is on scaleway "Starter VC1S" tier running ubuntu 16.04.

alimakki:
Resolves  #418, tested on Scaleway VC1S instance

### DIFF
--- a/playbooks/roles/common/vars/main.yml
+++ b/playbooks/roles/common/vars/main.yml
@@ -1,5 +1,7 @@
 ---
 streisand_common_packages:
+  # Ensure that Apparmor is installed
+  - apparmor
   # Enables support for the HTTPS protocol in APT sources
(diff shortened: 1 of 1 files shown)


bpinto:
Awesome job you are doing @alimakki !

jlund:
Thanks again, @alimakki. This project would not be the same without your participation.

PR #420, opened 2016-10-09
Document using ansible_user in inventory

FiloSottile:
Also, moved the become directive to the playbook. While it should work
correctly from ansible.cfg, it didn't in my tests.


jlund:
This is definitely more intuitive. Thank you!

PR #426, opened 2016-10-17
[aws] Add us-east-2 region

DavidWittman:
AWS just added this region today. See https://aws.amazon.com/blogs/aws/now-open-aws-us-east-ohio-region/.


jlund:
Thanks!

PR #427, opened 2016-10-18
Create and reference streisand-users mailing list

DavidWittman:
There is a lot of chatter within the Github issues for things like
troubleshooting connectivity and basic installation questions which
would be better suited for a mailing list.

I took the liberty of creating a list on Google Groups and adding it
to the documentation here. The idea is that this will reduce clutter
in the Github issues (which should arguably be only code-related issues)
and provide users with a way to discuss non-code issues.

@jlund If you accept this, shoot me an email and I'll get you set up as an admin on the mailing list.

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,7 +75,7 @@ Services Provided
 
 Installation
 ------------
-Please read all installation instructions **carefully** before proceeding.
+Please read all installation instructions **carefully** before proceeding. If you have questions or issues regarding the installation process, please create a topic on the [streisand-users](https://groups.google.com/forum/#!forum/streisand-users/) mailing list.
 
 ### Important Clarification ###
 Streisand is based on [Ansible](http://www.ansible.com/home), an automation tool that is typically used to provision and configure files and packages on remote servers. That means when you run Streisand, it automatically sets up **another remote server** with the VPN packages and configuration.


tedd4u:
+1 - I was just about to suggest the same (a mailing list for admins to learn about important updates).

nickolasclarke:
A potential issue I see with this is that google services are blocked in China, as I am sure you are all well aware, and this mailing list would be unreachable by those who are submitting non-code related issues, at least for that significant block of users.

DavidWittman:
@nickolasclarke Good point. Can you recommend any hosted mailing lists which are accessible in China? FWIW there will always be some amount of "support" type requests showing up in Github, so I don't think it would necessarily exclude Chinese users from getting help, but they just won't have access to the mailing list.

nickolasclarke:
@DavidWittman nothing I use at the moment, but I'll ask around. An Alternative of sorts I proposed before was having a gitter or IRC channel. Not async communication though. As you noted though, China support could always come through github. Not sure what the best answer is here. I'd defer to @jlund on this.

cpu:
I'm in favour of creating a mailing list & I believe @jlund is on-board as well having chatted recently.

I would also like to avoid Google Groups. The accessibility issues from China are unfortunate and I don't think it gives a very good user experience for mailing lists. 

Can anyone speak to the reachability of https://riseup.net from China? That would be my first choice for where to situate a mailing list.

nickolasclarke:
@cpu my tests here show that it is blocked, [great fire](https://en.greatfire.org/https/riseup.net) also suggests that it is.

cpu:
@nickolasclarke Thanks for testing. That's unfortunate but I suppose not especially surprising. 

With the two immediate choices (Google Groups, Riseup) being blocked in China my own thought is we should use one anyway & acknowledge that some users will not be able to reach it.

Longer term it could be worth looking at running a Streisand mailman instance (this is what the Tor project does for instance) but that's a large resource/time commitment and could just as easily end up being blocked in China down the road.

Thoughts? I don't want to let perfect stand in the way of the good and would love to have a place to discuss Streisand outside of Github issues.

nickolasclarke:
@cpu I would agree that its probably not worth trying to go down the road of rolling our own, and users will likely find a way to reach a mailing list if its blocked. Good Enough > Perfect indeed. Just choosing whichever mailing list service works best for you is probably the way to go. 

The only other suggestion that immediately comes to mind is creating a second, "issue only" repo that could be used to track issues etc. However, to do this properly, this would mean moving Streisand into an org, and then creating another repo under the org, etc. A bit messy. 

However, I really do like issue tracking through GH rather than a mailing list. Its easier to keep track of comments and resolution of issues, and has greater discoverability, as mailing lists like google groups are not well indexed in search engines, if at all, and the added advantage that GH is not blocked here in china. 

That or a gitter/slack/mattermost/irc as I mentioned before. 

😩

cpu:
> Just choosing whichever mailing list service works best for you is probably the way to go.
:+1: 

> The only other suggestion that immediately comes to mind is creating a second, "issue only" repo that could be used to track issues etc. However, to do this properly, this would mean moving Streisand into an org, and then creating another repo under the org, etc. A bit messy.

To be clear, when you say "issue only" you mean setup/troubleshooting type issues? I agree that moving to an org is probably sufficiently messy to be a longer term goal.

I'm also OK with folks still asking troubleshooting/setup questions in GH issues, I think a mailing list will be valuable for that and other purposes but shouldn't be a total replacement for GH.

> That or a gitter/slack/mattermost/irc as I mentioned before.
Speaking personally I'm not super interested in a more real-time/chat based forum. IRC in particular is difficult to moderate and not a particularly pleasant medium in my experience.

nickolasclarke:
I mean a repo only used for github issues. There would be no code
committed, and the repos issues section would be dedicated only to non-code
discussions. An example is [code school's](
https://github.com/nodeschool/discussions) issue only repo.

We do the same at work for issue tracking non code related projects.

cpu:
> I mean a repo only used for github issues. There would be no code committed, and the repos issues section would be dedicated only to non-code discussions.

Ahh! Thanks for clarifying. That's an idea I hadn't considered.

nickolasclarke:
like I said, to manage it best, it'd probably need to be part of a Streisand organization, though @jlund could always just make another repo for it. 

Anyhow, works well for us at my org. Keeps things centralized around GH, keeps things issue based, that can be easily classified, closed, etc.  You can cross-reference issues across repos still, and even close tickets in other repos with PRs etc.  

Additional benefits for streisand users is that things are indexed by the search engines for discoverability, and privacy is arguably better maintained as users continue to only need to expose the GH user they've decided to use for interacting with the community.

jlund:
I came here ready to say "Yeah, Riseup!" or "Yeah, Google Groups!" but now I'm intrigued by the idea of an issue-only repo. I've seen a few closed-source apps use that approach and it always felt a little weird in that context--like they were taking advantage of GitHub somehow. I don't think that criticism would apply here.

I can see pros and cons in both directions.

With Riseup, people could either send an email to join the mailing list or do so via a simple form on the Riseup site. Individuals who don't necessarily have (or want) a GitHub account would still be able to participate. The email approach would be relatively censorship-resistant once someone was signed up, although it might be difficult for some users to confirm their subscription if the site is blocked. It would also be more difficult to read the archives on an ongoing basis for these users. There are the potential privacy issues that @nickolasclarke mentioned as well. Most of the same would apply to Google Groups too, albeit with the additional requirement of having a Google account (but those are far more ubiquitous than GitHub accounts).

With a GitHub issue-only repo, people will need to create a GitHub account in order to participate. That's arguably a very big hurdle. There will also likely be more confusion about where an issue should be filed. But in exchange for all of that we would get built-in Markdown formatting, labels, and other little things that aren't super essential but that are still kind of nice (like emoji :tada:). The ability to quickly and easily cross-reference issues sounds pretty great too. I can easily make a `streisand-discussions` repo...

I think this comment I'm writing would be tagged as `bikeshedding` in the NodeSchool discussions repo (especially after all of this time has passed), but I also want to get this right. I'm leaning more towards the issue-only repo right now, but the opinions of the community and of the project's collaborators are really important to me. What do you think?

CorbanR:
I have never used Riseup but looks interesting. I can say that I'm NOT a huge fan of  Google Groups. That is not so say i'm opposed to a mailing list, I think we should have something. I'm also leaning toward something like a `streisand-discussions` repo. We would just need to be diligent moving/closing issues if they were opened in the wrong repo(Example being an Issue opened in streisand when it really should be in `streisand-discussions`).  I think something like a striesand discourse would be cool, but would similarly to running mailman would take time/resources.

nickolasclarke:
- Re Mailing lists: 
  Essentially only accessible to those who are in the mailing list. I don't think I've ever found the answer to a technical question I've had on a Google Group. I've gotten help from one exactly once. I've never heard of Riseup until this discussion, which, at least for me, indicates its not a place answers can be found easily to non-subscribers. 

This may or may not be something we care about, but it seems that non-code related issues seem to break down mostly along two lines: 
1. A misunderstanding of how to deploy/manage streisand or individual components of it.
2. broader discussions about new features, architecture, etc.

The things caught in the first should probably be addressable by FAQ's/instructions and new features that address some of the pain points (such as redeploying a new node with existing creds).  These types of questions will never end, and should be as discoverable as possible. This suggests that a mailing list is probably not the right answer for addressing these.

--- 
- Re GH issues-only repo:
   - I think with some well placed text on the `streisand` README.md and a issue template for both repos, we can hopefully offload a lot of the work of ensuring that issues are put in the right repo. 
    - People are coming to GH to get streisand in the first place, and unless they are downloading the repo as a tarball and deploying that way, they ~~almost certainly have a GH account already.~~ *apparently you dont have to have a GH account to clone the repo with https*. 
    - GH is not blocked (in china ) unlike both the other suggestions, and thus already more accessible.

All this to say, I'm not highly invested in any one solution, and completely flexible to a final solution. It should be noted that we are not locked in, and if our choice does become a major point of friction, we can move.

nickolasclarke:
oh, another thought I had was a stackexchange group as another alternative...but thats also not a great solution.

cpu:
@CorbanR @nickolasclarke @jlund I think I see enough consensus here for the "use a separate repo for discussions/support" idea that I'm happy to put in my own +1 for that approach too. We can give it a shot and reevaluate based on how it goes :-)

@jlund Will you create the repo(s)? I can take a crack at issue/pr templates if nobody beats me to it. It seems like short-term creating them under your own `jlund` namespace like the Streisand project is easiest, but having a separate parent org is probably the most future-proof option if you have the cycles.

cpu:
Update: There is a `streisand-discussions` repository based on the discussion in this thread. It's private currently pending the writing of the issue template and updating the docs/readme here to point to the new forum.

I'm going to close this PR and track the work for the discussions repo here: https://github.com/jlund/streisand/issues/656

Thanks everyone!

PR #435, opened 2016-11-03
Update AWS auto-recovery conditions

DavidWittman:
EC2 auto-recovery is live in all AWS Regions, and now supports more instance types.

See https://aws.amazon.com/about-aws/whats-new/2016/05/amazon-ec2-auto-recovery-now-available-in-the-beijing-china-region/ for more info. It looks like this was available when I first introduced this PR in #374, but I just now noticed it's available in all regions.


jlund:
This is great. Thanks!

PR #447, opened 2016-12-02
More error checking in provisioning script

sparkcodeuk:
Some minor edits to the main provisioning script, checking whether ansible itself is installed before checking its version, feedback if the user get's the options wrong when selecting which provider, minor cleanup.

### DIFF
--- a/streisand
+++ b/streisand
@@ -7,7 +7,7 @@ echo
 echo -e "\033[38;5;255m\033[48;5;234m\033[1m  S T R E I S A N D  \033[0m"
 echo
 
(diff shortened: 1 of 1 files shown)


streisand
```diff
@@ -1,15 +1,23 @@
 #!/usr/bin/env bash
+# Streisand provisioning script
+
+set -e
 
 echo
 echo -e "\033[38;5;255m\033[48;5;234m\033[1m  S T R E I S A N D  \033[0m"
 echo
 
+command -v ansible 2>&1 > /dev/null || {
```
sparkcodeuk: Fixed, thanks.

PR #451, opened 2016-12-10
Added tokyo2 region for linode

zjx20:
According to this doc: https://www.linode.com/api/utility/avail.datacenters


DavidWittman:
:+1:

PR #452, opened 2016-12-10
Update monitrc.j2: don't monitor for non-installed softwares

zjx20:
The final goal is that we are able to choose what services will be installed by streisand, with arbitrary  combination.


jlund:
Sorry about the hilariously large delay in posting feedback on this. I think I'm going to go in a different direction and completely customize the monitrc depending on whether or not a given role is enabled. This PR will be helpful as a reference when I get to that point though, and I sincerely thank you for your help!

PR #453, opened 2016-12-10
Uses default values in template to avoid "undefined variable" errors

zjx20:
The goal is same to #452. If we commented out openvpn in streisand.yml for example, variables defined in `playbooks/roles/openvpn/defaults/main.yml` are unable to be accessed by other roles. Adding the `default` filter can avoid runtime errors, though it is quite ugly. @jlund if you don't like this solution, could you give me some suggestions?

### DIFF
--- a/playbooks/roles/streisand-gateway/templates/firewall.md.j2
+++ b/playbooks/roles/streisand-gateway/templates/firewall.md.j2
@@ -12,19 +12,19 @@ If you are installing Streisand on an existing server or alternate provider then
 * Nginx (Streisand Gateway)
   * TCP - {{ nginx_port }}
 * OpenConnect / Cisco AnyConnect
-  * TCP - {{ ocserv_port }}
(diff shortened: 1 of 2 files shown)


DavidWittman:
Thanks @zjx20. I don't think this is terribly ugly... In fact it ends up being a pretty common pattern in many cases.

That said, I think this makes a case for defining many of these variables which are used across roles in a central place. It could be `group_vars/all` or something like `default.config.yml`, which would then be included in each play. This would allow the user more control over configuring services without having to alter the role and could prevent the types of issues we're addressing here.

jlund:
I can definitely see the benefits of this approach. I'm considering a different direction as well. Ensuring that roles are completely isolated and that users can choose the set of services to install is my very top priority after I finish support for Wireguard.

zjx20:
Closing this PR as it's corrupt. It seems that @jlund is going to work in this part, I'm really looking forward to it :)

PR #456, opened 2016-12-12
Removed ike_port, ipsec_port and l2tp_port variables

zjx20:
Hi @DavidWittman, I'm trying to implement what you proposed in #453, to collect variables into a centralized place. However, I found that variables defined in `roles/l2tp-ipsec/defaults/main.yml` are totally useless. They are not actually "variable" since we can't change them to other values, and they had already been hardcoded in many place.

Removing them can eliminate fews "undefined variable" errors. Otherwise, we have to move them to the centralized place, but I think it will mislead users that those ports are customizable.


jlund:
Looks reasonable to me. Thanks!

PR #467, opened 2017-01-03
UFW: open udp port for Shadowsocks.

yuanl:
As a result of GFW DNS spoofing. Most Shadowsocks users in China need to relay their DNS request via ss-tunnel.
Shadowsocks-libev official repository [enable this feature by default](https://github.com/shadowsocks/shadowsocks-libev/blob/master/debian/shadowsocks-libev.default#L18) on server-side. I thinks it's better to let UFW open the udp port for Shadowsocks.


jlund:
This looks good, but we'll need to do the same for EC2 and GCE as well.

PR #492, opened 2017-01-27
Added support for f1-micro instance type in google cloud

blotzu:
Added support for the cheapest google cloud instance type.
It's 10$ cheaper than g1-small.
I've noticed no performance issues with one user.

### DIFF
--- a/playbooks/google.yml
+++ b/playbooks/google.yml
@@ -66,6 +66,14 @@
       default: "3"
       private: no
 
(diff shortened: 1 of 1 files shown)


jlund:
Thanks!

jlund:
This instance type works well enough that I don't even think we need the prompt :) I'm just going to set it as the true default for GCE instances.

blotzu:
Thanks :)

PR #504, opened 2017-02-06
 Add support for tlscrypt

alimakki:
- updated server and client to use tlscrypt directive, removed key direction as it is no longer needed
 - changed OpenVPN cipher to AES-256-GCM
 - created a new 'combined' client profile that cycles through UDP port 8757, TCP port 636, and TCP port 443
    - see '<connection>' section on https://community.openvpn.net/openvpn/wiki/Openvpn24ManPage
    - added a new variable for openvpn_port_sslh to be more consistent with the rest of the task
    - updated OpenVPN instructions to reflect combined profiles
 - updated OpenVPN mirror for windows to use v2.4.0
    - there doesn't seem to be a separate 32bit and 64 bit builds anymore, so the task to mirror was updated
      to reflect that change
    - mirroring is done through https now
 - temporarily hardcoded shadowsocks for trusty as xenial returned 404s (if that's no longer the case I'll change it)

Tested on MacOS El-Capitan with OpenVPN 2.4.0 (cli version from homebrew [works]), OpenVPN connect iOS v1.1.1 (does not work) on a DO droplet.

As of this writing, this PR should **not** be merged since OpenVPN connect on iOS doesn't seem to play nice with tlscrypt just yet.

In any case, there's no harm in making it ready. 

Feedback is always appreciated.


atulsian89:
@alimakki Thanks for adding the commits for 2.4. I was wondering if we can also change the LZO compression to LZ4?

alimakki:
@atulsian89 not a problem, once I get back home from work.

alimakki:
@atulsian89 lz4 has been added if you want it.

atulsian89:
@alimakki, Thanks for adding it. I don't know how to add it to the current setup. I'll wait until it gets merged and also are you planning to support ECDH and ECDSA?

alimakki:
@atulsian89 ECDH should already happen if client and servers are already up to date; generating dhparams the classic way is still there for backwards compatibility with older clients, although it can be explicitly disabled.

ECDSA is also possible, but a bit more work as we'd have to run slightly different commands when generating the OpenVPN PKI and client/server certificates.

jlund:
This is awesome work. I can't wait until we can turn all of this on.

I think that there's a case to be made about doing a more selective roll-out of these features. We're realistically probably a fair bit of time away before native 2.4 clients are widely deployed. In the meantime though, a lot of this stuff will work in version 2.3 and OpenVPN 2.4 introduced some nice transition features like [cipher negotiation](https://community.openvpn.net/openvpn/wiki/Openvpn24ManPage):

> Additionally, to allow for more smooth transition, if NCP is enabled, OpenVPN will inherit the cipher of the peer if that cipher is different from the local --cipher setting, but the peer cipher is one of the ciphers specified in --ncp-ciphers. E.g. a non-NCP client (<=2.3, or with --ncp-disabled set) connecting to a NCP server (2.4+) with "--cipher BF-CBC" and "--ncp-ciphers AES-256-GCM:AES-256-CBC" set can either specify "--cipher BF-CBC" or "--cipher AES-256-CBC" and both will work.

OpenVPN 2.4 has this enabled by default, and this means that many clients connecting to Streisand servers since the 2.4 server upgrade a few weeks ago are already using AES-256-GCM. The iOS client, for example, successfully connects with AES-256-GCM right now and the performance is _way_ better.

I think outside of `--tls-crypt` and lz4 compression we should be able to deploy a lot of this PR today while still supporting 2.3 users for a while longer. OpenVPN 2.3 has support for `<connection>` blocks too, for example.

These are some really nice enhancements, and it would be awesome to quickly get as many of them out there as we can--particularly updating the mirrors to get more 2.4 users into the world :)

alimakki:
Hi @jlund, thanks for the feedback!

I considered the suggestions that you have made and reverted back to `tls-auth`, `comp-lzo` for the time being, as well as adding `ncp-ciphers`, which I currently set to `AES-256-GCM:AES-256-CBC`.

Additionally, for the combined profiles I added a `timeout` command set to 15 seconds (a completely arbitrary value); my line of thought being should a profile like this be given out to someone who is not a power user, having the VPN client cycle through connections a bit quicker in a more restrictive network (let's say some airport WiFi where UDP traffic is blocked and only TCP 80 and 443 is allowed) to establish a tunnel would be a less frustrating experience than waiting the current default's 120 seconds.

If this commit as is looks good to you, let me know and I'll squash.

Thanks!

playbooks/roles/openvpn/templates/instructions.md.j2
```diff
@@ -168,3 +174,9 @@ It's preferable to configure Ubuntu using the OpenVPN plugin for NetworkManager.
 {% for client in openvpn_clients %}
 * [{{ client }}](/openvpn/{{ client }}/{{ openvpn_direct_udp_profile_filename }})
 {% endfor %}
+
+<a name="combined-profiles"></a>
+### Alternate profile that that will cycle through UDP port {{ openvpn_port_udp }}, TCP port {{ openvpn_port }}, and TCP port {{ openvpn_port_sslh }} ###
```
jlund: "That" is repeated twice.

playbooks/roles/shadowsocks/tasks/main.yml
```diff
@@ -91,4 +91,4 @@
     path: "{{ shadowsocks_password_file }}"
     owner: root
     group: root
-    mode: 0600
+mode: 0600
```
jlund: The indentation is off here. I don't think any changes to this file should be necessary.

playbooks/roles/streisand-mirror/vars/openvpn.yml
```diff
@@ -17,27 +17,19 @@ openvpn_source_sig_href: "{{ openvpn_mirror_href_base }}/{{ openvpn_source_sig_f
...
-openvpn_windows64_installer_href: "{{ openvpn_mirror_href_base }}/{{ openvpn_windows64_installer_filename }}"
-openvpn_windows64_installer_sig_href: "{{ openvpn_mirror_href_base }}/{{ openvpn_windows64_installer_sig_filename }}"
-openvpn_windows64_installer_url: "{{ openvpn_base_download_url }}/{{ openvpn_windows64_installer_filename }}"
-openvpn_windows64_installer_sig_url: "{{ openvpn_base_download_url }}/{{ openvpn_windows64_installer_sig_filename }}"
+openvpn_windows_installer_filename: "openvpn-install-{{ openvpn_version }}-I602.exe"
+openvpn_windows_installer_sig_filename: "{{ openvpn_windows_installer_filename }}.asc"
+openvpn_windows_installer_href: "{{ openvpn_mirror_href_base }}/{{ openvpn_windows_installer_filename }}"
+openvpn_windows_installer_sig_href: "{{ openvpn_mirror_href_base }}/{{ openvpn_windows_installer_sig_filename }}"
+openvpn_windows_installer_url: "{{ openvpn_base_download_url }}/{{ openvpn_windows_installer_filename }}"
+openvpn_windows_installer_sig_url: "{{ openvpn_base_download_url }}/{{ openvpn_windows_installer_sig_filename }}"
```
jlund: So much cleaner! :)

playbooks/roles/openvpn/defaults/main.yml
```diff
@@ -13,6 +13,9 @@ openvpn_clients:
 
 openvpn_port: "636"
 openvpn_port_udp: "8757"
+openvpn_port_sslh: "443"
 openvpn_key_size: "2048"
-openvpn_cipher: "AES-256-CBC"
+openvpn_cipher: "AES-256-GCM"
+openvpn_ncp_cipher: "AES-256-GCM:AES-256-CBC"
 openvpn_auth_digest: "SHA256"
+openvpn_connect_timeout: 15 #seconds
```
jlund: I think we could safely drop this down to 5 seconds. What do you think? Fifteen seconds is way better than the default, but it could potentially mean waiting 30 seconds (instead of 10) before the sslh connection option is attempted.

playbooks/roles/openvpn/defaults/main.yml
```diff
@@ -13,6 +13,9 @@ openvpn_clients:
 
 openvpn_port: "636"
 openvpn_port_udp: "8757"
+openvpn_port_sslh: "443"
 openvpn_key_size: "2048"
-openvpn_cipher: "AES-256-CBC"
+openvpn_cipher: "AES-256-GCM"
```
jlund: We need to keep this set to something 2.3-compatible like `AES-256-CBC` in order to prevent breaking older clients. The `ncp-ciphers` option will allow 2.4 clients to opportunistically upgrade their connection to something GCM-based.

playbooks/roles/openvpn/defaults/main.yml
```diff
@@ -13,6 +13,9 @@ openvpn_clients:
...
+openvpn_ncp_cipher: "AES-256-GCM:AES-256-CBC"
```
jlund: Ah, the way you've done this would work for 2.3 clients as well. I didn't notice the `AES-256-CBC` here at first.

jlund: However, this would only work if the client profiles for 2.3 clients specified a 2.3-compatible cipher. Right now they would have AES-256-GCM as their cipher, and 2.3 clients wouldn't know what to do with that.

jlund:
Looks great! I just added a few minor comments.

The biggest outstanding issue is that a 2.3 client wouldn't be able to connect to a server with the modifications in this PR using any of the generated profiles unless the profile was manually modified to specify a `cipher` setting of `AES-256-CBC`.

We need to choose one of these two options:

1. Specify a 2.3-compatible `cipher` setting for the server and let 2.4 clients opportunistically upgrade to a GCM cipher from the `ncp-ciphers` list (which is what the iOS client does, for example).
2. Generate 2.3-specific or 2.4-specific connection profiles.

I think the first approach is a little bit cleaner, but maybe the iOS client is an outlier in terms of how well it handles this. I am waiting for a new Android phone to arrive so that I can start doing extensive testing on that platform again. I'm not opposed to version-specific profiles, but that is a lot of duplication. What do you think?

If we decide to go with the first approach, the `cipher` would become `AES-256-CBC` and `ncp-ciphers` would become `AES-256-GCM:AES-128-GCM` (to match the OpenVPN default).

Unfortunately, it doesn't look like `cipher` is one of the accepted options that can be used inside of a connection block. That could have been another interesting approach too :)

playbooks/roles/openvpn/templates/instructions.md.j2
alimakki: fixed!

playbooks/roles/streisand-mirror/vars/openvpn.yml
alimakki: :)

playbooks/roles/openvpn/defaults/main.yml
alimakki: and done!

playbooks/roles/openvpn/defaults/main.yml
alimakki: I feel that 5 seconds might be a bit too low; based on experience when I used to live in Oman connectivity can be shady at times, and tunnels would take a bit longer than anticipated to establish.

15 seconds might also be a bit much in retrospect, so I dialed it down to 10. I feel that it's a reasonable compromise.

alimakki:
Gonna try and spawn an instance on DO, if it all goes well, I'll close this PR and re-open another one with with the commits squashed.

alimakki:
All tested, server is provisioned correctly and my iOS client happily connects with the combined profile (w/AES-256-GCM).

For what it's worth, MacOS users using Viscosity won't be able to use combined profile until viscosity upgrades their client.

Closing this PR and re-opening with squashed history.

PR #524, opened 2017-02-21
OpenVPN upgrades

alimakki:
OpenVPN Upgrades 
- updated server and client to use ncp_ciphers
 - created a new 'combined' client profile that cycles through UDP port 8757, TCP port 636, and TCP port 443
    - see '<connection>' section on https://community.openvpn.net/openvpn/wiki/Openvpn24ManPage
    - added a new variable for openvpn_port_sslh to be more consistent with the rest of the task
    - updated OpenVPN instructions to reflect combined profiles
 - updated OpenVPN mirror for windows to use v2.4.0
    - there doesn't seem to be a separate 32bit and 64 bit builds anymore, so the task to mirror was updated
      to reflect that change
    - mirroring is done through https now


jlund:
Thank you! These are some very nice improvements.

Issue #525, opened 2017-02-24
PuTTY could not be mirrored.


mystfox:
Relevant part of the output:
```
TASK [streisand-mirror : Mirror the PuTTY files and signatures] ****************
failed: [188.166.84.21] (item=http://the.earth.li/~sgtatham/putty/latest/x86/putty.exe) => {"dest": "/var/www/streisand/mirror/ssh", "failed": true, "gid": 33, "group": "www-data", "item": "http://the.earth.li/~sgtatham/putty/latest/x86/putty.exe", "mode": "0755", "msg": "Request failed", "owner": "www-data", "response": "HTTP Error 404: Not Found", "size": 4096, "state": "directory", "status_code": 404, "uid": 33, "url": "http://the.earth.li/~sgtatham/putty/latest/x86/putty.exe"}
failed: [188.166.84.21] (item=http://the.earth.li/~sgtatham/putty/latest/x86/putty.exe.gpg) => {"dest": "/var/www/streisand/mirror/ssh", "failed": true, "gid": 33, "group": "www-data", "item": "http://the.earth.li/~sgtatham/putty/latest/x86/putty.exe.gpg", "mode": "0755", "msg": "Request failed", "owner": "www-data", "response": "HTTP Error 404: Not Found", "size": 4096, "state": "directory", "status_code": 404, "uid": 33, "url": "http://the.earth.li/~sgtatham/putty/latest/x86/putty.exe.gpg"}

TASK [streisand-mirror : One or more of the VPN clients could not be mirrored. Please file a bug report on GitHub so that the version number, checksum, or download location can be updated. Setup will now continue.] ***
Pausing for 20 seconds
(ctrl+C then 'C' = continue early, ctrl+C then 'A' = abort)
ok: [188.166.84.21]
```

And checking in browser, said link does indeed 404.

kehao95:
Same problem here.
`
TASK [streisand-mirror : Mirror the PuTTY files and signatures] ****************
[...]

PR #526, opened 2017-02-24
Update ssh.yml

ocrozo:
New url for downloading putty.exe.
Fix issue #525


jlund:
Thanks! It makes me happy to see TLS on PuTTY's download location.

Issue #527, opened 2017-02-27
OpenVPN combined profile "--connect-timeout" option invalid


kehao95:
Got an error while trying to use the OpenVPN combined config.

> Options error: --connect-timeout doesn't make sense unless also used with --proto tcp-client or tcp6-client

Full log here:

```
*Tunnelblick: OS X 10.12.4; Tunnelblick 3.7.0 (build 4790)
2017-02-27 14:26:43 *Tunnelblick: Attempting connection with 138.197.202.176-combined-5; Set nameserver = 769; monitoring connection
2017-02-27 14:26:43 *Tunnelblick: openvpnstart start 138.197.202.176-combined-5.tblk 1337 769 0 3 0 1065264 -ptADGNWradsgnw 2.3.14-openssl-1.0.2k
2017-02-27 14:26:43 *Tunnelblick: 

Could not start OpenVPN (openvpnstart returned with status #251)

Contents of the openvpnstart log:
*Tunnelblick: openvpnstart log:
     OpenVPN returned with status 1, errno = 0:
          Undefined error: 0
     
     Command used to start OpenVPN (one argument per displayed line):
     
          /Applications/Tunnelblick.app/Contents/Resources/openvpn/openvpn-2.3.14-openssl-1.0.2k/openvpn
          --daemon
          --log
          /Library/Application Support/Tunnelblick/Logs/-SLibrary-SApplication Support-STunnelblick-SShared-Sxx.xx.xx.xx--combined--5.tblk-SContents-SResources-Sconfig.ovpn.769_0_3_0_1065264.1337.openvpn.log
          --cd
          /Library/Application Support/Tunnelblick/Shared/xx.xx.xx.xx-combined-5.tblk/Contents/Resources
          --verb
          3
          --config
          /Library/Application Support/Tunnelblick/Shared/xx.xx.xx.xx-combined-5.tblk/Contents/Resources/config.ovpn
          --verb
          3
          --cd
          /Library/Application Support/Tunnelblick/Shared/xx.xx.xx.xx-combined-5.tblk/Contents/Resources
          --management
          127.0.0.1
          1337
          --management-query-passwords
          --management-hold
          --script-security
          2
          --up
          /Applications/Tunnelblick.app/Contents/Resources/client.up.tunnelblick.sh -9 -d -f -m -w -ptADGNWradsgnw
          --down
[...]

PR #528, opened 2017-02-27
issue #527 remove "--connect-timeout" for udp protocol in combined profile

kehao95:
remove the option and it works fine


jlund:
Thank you!

PR #535, opened 2017-03-06
Add pycrypto lib as requirement for Google provider.

bryan4887:
This was required for me. Not sure if it's something unique to my environment...

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,6 +133,7 @@ Complete all of these tasks on your local home machine.
   * Google
 
             sudo pip install "apache-libcloud>=0.17.0"
+            sudo pip install pycrypto          
 
   * Linode
 


jlund:
The `pycrypto` library is required by the Ansible pip package and its [installation is also handled by Homebrew](https://github.com/Homebrew/homebrew-core/blob/master/Formula/ansible.rb#L374). Can you tell me a little bit more about your environment?

cpu:
A missing `pycrypto` package also caused a GCE provisioning failure in https://github.com/jlund/streisand/issues/584. In that case their local pip appears unusable.

Since as @jlund points out, PyCrypto is already a second level dependency that should be installed under normal circumstances, I don't think there is much value in explicitly listing it in the Streisand setup instructions, doubly so if it only applies for GCE provisioning since not all users will do that.

If the issue is reproducible for OSX users it seems better to open a new issue to track down why the Ansible formula isn't working correctly. Thoughts?

cpu:
I'm going to close this PR - there hasn't been any response to the questions posed so far and I suspect it was a fix for a special case environment since it hasn't been raised by any other GCE users. Thanks!

Issue #531, opened 2017-03-02
OpenVPN Ubuntu instructions should mention Gnome


mangyfox:
Whilst following the setup instructions for OpenVPN on Linux (Ubuntu) I hit a snag. After installing network-manager-openvpn as instructed, I was unable to select OpenVPN in the network settings.

After some searching, I discovered that I also had to install network-manager-openvpn-gnome. Once this package was installed I could select OpenVPN, follow the rest of the instructions, and everything worked perfectly.

I'm running Ubuntu 16.04, with no relevant changes from the default setup, so I imagine other users will also be running into this problem. For this reason, I think it would be useful to mention network-manager-openvpn-gnome in the Ubuntu instructions.

mlcdf:
I had the same issue. I just opened a PR in order to add a little notice about this.

jlund:
Thanks for the bug report!

PR #547, opened 2017-03-24
Update OpenVPN instructions on Ubuntu

mlcdf:
Add a notice that users might need to install`network-manager-openvpn-gnome` in order to select *OpenVPN* as a VPN Connection Type.

(closes to #531)


jlund:
I pushed a commit for this in 6354773 that takes a slightly different approach. The `network-manager-openvpn-gnome` package depends on `network-manager-openvpn` which allows us to just reference that single package. The documentation was indeed insufficient prior to this. Thank you for your help!

Issue #548, opened 2017-03-26
Streisand-Gateway Certification Request Fail


WanderingViking:
Been attempting to setup Streisand on an AWS instance but when it gets to the TASK [streisand-gateway : Create the certificate request] it fails. Anyone else having this issue or know of a work around. Once it fails no docs are generated. 

OS: Ubuntu 16.04 LTS (Running on VMWare Workstation 12 Pro)
Ansible 2.2.1.0
Python:  2.7.12
Using an AWS IAM with full permissions as the AWS Access Key ID and Secret Access Key

Error as follows:

TASK [streisand-gateway : Create a certificate authority key and certificate] **
changed: [52.52.166.247]

TASK [streisand-gateway : Create the certificate request] **********************
[...]

PR #551, opened 2017-03-26
Ensure the `streisand_server_name` variable does not get set to a value with whitespace characters

kernelpanek:
This is a proposed accommodation for issue #548. The fix takes place at the end of each `gensis-*` role where the `streisand_server_name` fact is being set for the remainder of the playbook's execution.


jlund:
This looks like a great solution to me, and kudos on so thoroughly debugging the original issue. Thanks for your help!

PR #554, opened 2017-03-29
Update README.md

matthewtrask:
Add a bit of emphasis to people using homebrew and running into errors and havent seen this note about how to fix them.

I spent a few hours running into issues with setting up an AWS server for Streisand when I finally stumbled on the info about homebrew. I think that it should be made to attract more attention from users that may be new to this kind of set up, or stuck and cant get around the error. 

Thanks for this awesome project!


jlund:
Thanks!

PR #562, opened 2017-03-31
Update README.md

march93:
Reformatted some sentences for simplification.


jlund:
Thanks! You and @hodog in #567 had very similar ideas, and that must mean they are good :)

PR #567, opened 2017-04-02
Remove redundancy in README

hodog:
Changed two sentences around so it didn't sound as redundant.


jlund:
This sounds much better. Thanks!

PR #588, opened 2017-04-09
fix: Documentation mispelling

ryanvanoss:
Happened to catch this as I was reading through the documentation :)


alimakki:
Thanks you!

Issue #568, opened 2017-04-02
[GCE] Unable to restart service xl2tpd


pchaganti:
Running the latest version on GCE. Getting this error:

```
RUNNING HANDLER [l2tp-ipsec : Restart xl2tpd] **********************************
fatal: [35.185.52.148]: FAILED! => {"changed": false, "failed": true, "msg": "Unable to restart service xl2tpd: Job for xl2tpd.service failed because the control process exited with error code. See \"systemctl status xl2tpd.service\" and \"journalctl -xe\" for details.\n"}
```

Anyone have any pointers?

thanks

cpu:
> See \"systemctl status xl2tpd.service\" and \"journalctl -xe\" for details.\n"}

Can you provide the output from both of these? It will likely help others troubleshoot the problem!

pchaganti:
Could not SSH into the launched server.  Unable to run these commands. 

Switched to using AWS. Worked fine for me.

cpu:
> Could not SSH into the launched server. Unable to run these commands.

Ok, thanks anyway! Glad to hear you were able to find a work-around.

@CorbanR - I see you assigned this to yourself (Thanks!!). Do you think you can reproduce with the information available?

CorbanR:
@cpu Yeah, ill test it out this evening! Also thank for doing all this ticket triage dude!

tedd4u:
I saw the same error today with latest code on GCP. Also couldn't ssh in. And https port wasn't responding. I already nuked the server but I'm happy to assist in any way. Have set up Streisand 3-4 times before on GCP with no problems. Last time was about March 24th.

CorbanR:
I didn't run into the same problem when testing. I assume it was just a fluke.. Having said, that if two people are having the same issue, it makes me think it's more than just a fluke
I tried with the most recent version of Ansible `2.2.2.0`, and previous `2.2.1.0`, I am also using a fairly recent version of `apache-libcloud`

@tedd4u @pchaganti can you provide me with the following information? 

OS: OSX, Ubuntu, etc.
Ansible version: `ansible --version`
apache-libcloud ` pip list | grep -i apache (assuming you used pip)`


As far as having problems when trying to ssh on the instance, ensure you are using the correct ssh key, if trying to login as root, make sure you're specifying the right user. `ssh ubuntu@<server_ip>`

tedd4u:
Thank you @CorbanR 

[...]

PR #593, opened 2017-04-10
Misc updates/fixes

CorbanR:
When creating GCE network use auto instead of legacy mode

Update GCE doc url to use https
Bump minimum apache-libcloud version
Add retry logic to l2tp-ipsec handler

Update ansible.cfg with temporary workaround related to ansible/ansible#18894

Add system update task

### DIFF
--- a/playbooks/roles/l2tp-ipsec/handlers/main.yml
+++ b/playbooks/roles/l2tp-ipsec/handlers/main.yml
@@ -8,3 +8,7 @@
   service:
     name: xl2tpd
     state: restarted
+  register: xl2tpd_restart_result
+  until: xl2tpd_restart_result|success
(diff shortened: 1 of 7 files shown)


playbooks/roles/l2tp-ipsec/handlers/main.yml
```diff
@@ -8,3 +8,7 @@
   service:
     name: xl2tpd
     state: restarted
+  register: xl2tpd_restart_result
```
CorbanR: I'm not a huge fan of this particular work around. When using the f1-micro type on GCE this appears to work around https://github.com/jlund/streisand/issues/568.

jlund:
Thanks!

Issue #585, opened 2017-04-08
Unable to reach monit web interface without additional configuration


rmcentyre:
After setting up my streisand server on Digital Ocean (say on IP 123.45.67.89), I discovered I couldn't access monit using `123.45.67.89/monit`. Using an SSH LocalForward config like `LocalForward 2812 localhost:2812` and trying `localhost:2812` was also unsuccessful. Both of these attempts resulted in the following error in `/var/log/monit.log`.

    [UTC Apr  8 20:25:31] error    : Denied connection from non-authorized client [123.45.67.89]

SSHing directly into the box with no additional LocalForward configs and doing `monit status` was similarly unsuccessful. I was able to resolve all of this by adding an `allow` item to `/etc/monit/monitrc`.

    set httpd port 2812 and
        use address localhost  # only accept connection from localhost
        allow localhost        # allow localhost to connect to the server
        allow 123.45.67.89  # add this streisand server's public IP here

Perhaps this is unnecessary and someone (maybe @Russell-IO ?) can tell me what I'm doing wrong.

PR #602, opened 2017-04-14
Add Streisand IP to Monit config

CorbanR:
fixes https://github.com/jlund/streisand/issues/585

### DIFF
--- a/playbooks/roles/monit/templates/monitrc.j2
+++ b/playbooks/roles/monit/templates/monitrc.j2
@@ -5,9 +5,11 @@ set idfile /var/lib/monit/id
 set statefile /var/lib/monit/state
 
 set httpd port 2812 and
-    use address localhost  # only accept connection from localhost
-    allow localhost        # allow localhost to connect to the server
+    use address localhost  # Bind monit to localhost
 
+    # Allow any ip to connect to monit. Binding monit to localhost
+    # and requiring basic auth (through nginx) should be sufficient to protect this interface.
+    allow 0.0.0.0/0.0.0.0 # allow any ip to connect to monit.
 
 # Nginx
 check process nginx with pidfile /var/run/nginx.pid
--- a/playbooks/roles/streisand-gateway/templates/vhost.j2
+++ b/playbooks/roles/streisand-gateway/templates/vhost.j2
@@ -25,7 +25,7 @@ server {
   location /monit/ {
     rewrite ^/monit/(.*) /$1 break;
     proxy_ignore_client_abort on;
-    proxy_pass   http://localhost:2812; 
+    proxy_pass   http://localhost:2812;
   }
 }
 


playbooks/roles/monit/templates/monitrc.j2
```diff
@@ -5,9 +5,11 @@ set idfile /var/lib/monit/id
 set statefile /var/lib/monit/state
 
 set httpd port 2812 and
-    use address localhost  # only accept connection from localhost
-    allow localhost        # allow localhost to connect to the server
+    use address localhost  # Bind monit to localhost
 
+    # Allow any ip to connect to monit. Binding monit to localhost
+    # and requiring basic auth (through nginx) should be sufficient to protect this interface.
+    allow 0.0.0.0/0.0.0.0 # allow any ip to connect to monit.
```
CorbanR: After explicitly adding `{{ streisand_ipv4_address }}` to the allowed list. I realized that this may not always work. It then occurred to me that if we consider nginx + basic auth good enough to protect streisant vpn connection info; It's good enough protect the monit url path(especially since monit is being bound to localhost).  I don't see much of a win to explicitly whitelisting client Ip's, especially since that seems to actually cause problems.

cpu: > I realized that this may not always work.

Interesting. Can you elaborate? 

`allow 0.0.0.0/0.0.0.0` with the rationale you provide might be the best path but I would want to dig a little bit deeper into why the localhost allowance isn't working, or why the client IP space can't be used here.

CorbanR: The whitelist in this context is referring to the "Client IP" that you're accessing monit with. 

So if you're striesand IP was 25.10.136.211, with the current rules it would be blocked. If we added `{{ streisand_ipv4_address }}` you should be able to access monit if `eth0`on the server and `{{ streisand_ipv4_address }}` are the same 

Regarding the ` I realized that this may not always work`  I tested adding `{{ streisand_ipv4_address }}` with digitial ocean and it worked fine. When I tested with Azure it didn't work. When I did more digging, this was because on digital ocean `eth0 == {{ streisand_ipv4_address }}`. On Azure (probably GCE, and AWS) `eth0 != {{ streisand_ipv4_address }}`. `eth0` is a private IP. 

On Azure the "Client IP" that was trying to access monit, was equal the private IP. I immediately thought we could try set_real_ip_from, and or X-Forwared-For.. but decided maybe i'm making this more difficult than it needed to be. Hence the 

>It then occurred to me that if we consider nginx + basic auth good enough to protect streisant vpn connection info; It's good enough protect the monit url path

CorbanR:
@cpu @jlund @alimakki  any strong feelings for or against this?

playbooks/roles/monit/templates/monitrc.j2
cpu: Thanks! That makes sense. I agree with your rationale here. I also tested that the monit bind is sufficiently scoped that if ufw/iptables weren't there you still can't connect from an external host.

PR #608, opened 2017-04-15
OpenVPN mirror to use latest build

alimakki:
OpenVPN provides a [static url](https://build.openvpn.net/downloads/releases/latest/) to ease automation.

This PR updates Streisand's OpenVPN mirror task to use said static url and always have the latest Windows installer and source with the added benefit of one less thing to maintain, assuming no changes in GPG keys.

Tested on DO droplet with no issues.


jlund:
Thanks! This will be very nice.

PR #610, opened 2017-04-16
Modular

blotzu:
Added support for selectable services.

### DIFF
--- a/README.md
+++ b/README.md
@@ -198,8 +198,6 @@ You can also run Streisand on any number of new Ubuntu 16.04 servers. Dedicated 
 
 The servers should be accessible using SSH keys, and *root* is used as the connecting user by default (though this is simple to change while editing the inventory file).
 
-By default all providers are enabled
-
 Upcoming Features
 -----------------
 * Easier setup.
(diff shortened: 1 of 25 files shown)


cpu:
Hi @blotzu,

This is a really large PR touching a lot of the project. Since there are no tests right now it will be difficult to merge this with certainty that nothing was missed and no clients will break. My own preference is to improve the state of automated testing & CI before trying to do any larger project-wide changes like introducing modularization. I'm 110% on board with getting there as a short term goal. 

Second, I think it is better to send individual PRs adjusting each role to resolve the cross-dependencies that you have fixed in this one PR. E.g. each role should be able to be enabled/disabled without issue before landing the PR to add the modularity toggling. That will help space out the review & make the process smoother.

@jlund @CorbanR @alimakki What are your thoughts?

jlund:
I agree with the feedback so far, but I also want to say that I think this is a solid start and I'm impressed by the scope and implementation. There are a few things that we might need to change, but this is actually very close conceptually to the experiments I have been doing in this area (with the added bonus of the command-line flags). I think once we have some testing in place this will be a nice foundation to build on. I'm also committed to modularity as a short-term goal. It has been nagging at me for a while :)

Thanks for taking this on, @blotzu. We should be able to move forward on this pretty soon.

cpu:
Given there are conflicts to be resolved, and there seems to be consensus between @jlund and I that this should be done in smaller incremental PR's I'm going to close this large PR since it won't be mergeable as-is.

@blotzu If you're still interested in taking this on I recommend that we create a separate Issue on the repo for this work where we can all collaborate on an approach before you submit any more PRs. Several of the maintainers (myself included) have been scheming about this for a little while and have a particular approach in mind. It would save review time if we could get everyone on the same page before doing any more code changes.

Thanks!

PR #611, opened 2017-04-16
Modern Shadowsocks server, config, and client options

### LINKED ISSUE #429: Please upgrade shadowsocks-ng to 1.3.1
This version fixed the connectivity under macos.

### LINKED ISSUE #586: Update shadowsocks-libev to 3.0.5, use AEAD cipher
Currently the shadowsocks server role installs [shadowsocks-libev version 2.5.6](https://github.com/jlund/streisand/blob/7f1cc442ef99ec3f5cb19e169b809541b0f3446b/playbooks/roles/shadowsocks/vars/main.yml#L21) and an [ota mode](https://github.com/jlund/streisand/blob/7f1cc442ef99ec3f5cb19e169b809541b0f3446b/playbooks/roles/shadowsocks/defaults/main.yml#L7) that was [deprecated in version 3.0.0](https://github.com/shadowsocks/shadowsocks-libev/releases/tag/v3.0.0).

[Version 3.0.5](https://github.com/shadowsocks/shadowsocks-libev/releases/tag/v3.0.5) is the latest release as of the time of this issue and Streisand should build & install a deb for it instead of 2.5.6. At the same time the ciphersuite should be changed from [`aes-256-cfb`](https://github.com/jlund/streisand/blob/7f1cc442ef99ec3f5cb19e169b809541b0f3446b/playbooks/roles/shadowsocks/defaults/main.yml#L5) to `chacha20-ietf-poly1305`, a strong elliptic curve based AEAD ciphersuite.

cpu:
**Updates shadowsocks-libev to 3.0.5, modern config**

4d00ea4 updates the shadowsocks role to install & configure
shadowsocks-libev 3.0.5 (the current release at time of writing) instead
of 2.5.6. In the process of doing so the deprecated OTA config option is
removed and the configured ciphersuite is switched from aes-256-cfb to
chacha20-ietf-poly1305.

Prior to this commit the build process involved installing a number of
dependencies that were only required for documentation or for building
a .deb to install. In the interest of slimming down the role this commit
opts to install the minimum required dependencies and installs directly
from source. The role templates its own systemd defaults and copies the
systemd service definition that the deb _would_ have installed directly
onto the system. This is a net savings of around ~800mb of dependencies
but means that there is a) no man file b) uninstallation must be done
with the Makefile instead of apt-get/dpkg.

Resolves #586

**Updates shadowsocks client mirror & docs.**

d257a1f updates the shadowsocks client mirror to match the
shadowsocks-libev update.

* The Android client is updated from 2.10.7 to **4.1.4**
* The Windows client is updated from 3.0 **4.0.1**
* The OSX client is updated from 1.2 to **1.5-beta.2**
* The Linux python client is replaced with **go-shadowsocks2 0.0.8**

Using a beta release of the OSX client is the only way to get AEAD
support without replacing the client with an alternative choice.

For Linux clients we previously recommended the Python pip `shadowsocks`
project, which remains at 2.8.2 (the version we are were mirroring
previous to this commit) and appears unmaintained. Since there's
a nicely maintained Go client with a responsive author this commit
replaces the Linux client mirror to use the x64 binary release of the
`go-shadowsocks2` project. One day I would like to see the
`shadowsocks-libev` server side role updated to use `go-shadowsocks2` as
well and so this is also a step towards that direction. The shadowsocks
Linux instructions are updated to use the new client.

The iOS shadowsocks instructions are updated to explicitly mention that
Potatso **does not support AEAD** and can't be used with the current
configuration. After some testing it appears only Cross on iOS supports
AEAD and so we recommend it. Shadowrocket is another option but is
currently only supporting AEAD in a limited test release.

Resolves #429

**Testing**

I have tested provisioning on DigitalOcean, and with an Android client as well 
as a Linux client using go-shadowsocks2.

Others in #586 have tested iOS (@jlund reports Cross works, @hybtoy reports
Shadowrocket beta works) and Windows (@hybtoy reports it works). The instructions
may need updating to match the UI/UX of Cross since I left the Potatso based
instructions as-is.

I would appreciate if someone could test with the mirrored OSX Client!


cpu:
Note: I left [a TODO in a vars file](https://github.com/cpu/streisand/blob/d257a1f05eb04528c9de8b34db9162c2e328dfe4/playbooks/roles/shadowsocks/vars/main.yml#L35) highlighting that `-v` is present in the default args. I'd prefer to remove that right before merging since it will be helpful for debugging as folks review.

alimakki:
I'll give this a test shortly!

alimakki:
@cpu Seeing an error during Streisand setup:

```TASK [shadowsocks : Ensure the Shadowsocks service is started] *****************
fatal: [159.203.28.62]: FAILED! => {"changed": false, "elapsed": 60, "failed": true, "msg": "Timeout when waiting for 159.203.28.62:8530"}
```

cpu:
@alimakki Is that on a fresh server? What provider? Is there an `ss-server` process running? It looks like its failing to check the port is open.

alimakki:
@cpu Fresh server on DO.

Doesn't appear to be an `ss-server` process running, console output:

```
root@shadowtest123:~# ps aux | grep ss-server
root     22023  0.0  0.2  12944  1084 pts/0    S+   16:57   0:00 grep --color=auto ss-server
root@shadowtest123:~#
```

If you want to poke around yourself I can add your public SSH key to it.

cpu:
> If you want to poke around yourself I can add your public SSH key to it.

@alimakki That would be great:
`ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIH+cuzpk3jabUFgLpW9Rf+3kOxKBq6iZS1iCmZR8ObYw cpu@github`

Here's my [GPG key](https://binaryparadox.net/pubkey.2016.txt) if you want to send me the details that way.

alimakki:
@cpu and done!

you should be able to login with `root@159.203.28.62`

ECDSA key fingerprint is `SHA256:wJN2/qq5QXLa1kyhYT4Zv7kS+KzYEUDdegKOLJVc9Vk.`

cpu:
@alimakki Thanks! I was able to confirm access. I'll try to do some debugging today and get back to you when its safe to destroy the instance. Appreciate the help!

alimakki:
@cpu my pleasure.

alimakki:
Oddly enough with your latest commit I'm no longer able to generate a DO droplet:

```
TASK [genesis-digitalocean : Add the SSH key to DigitalOcean if it doesn't already exist] ***
fatal: [localhost]: FAILED! => {"changed": false, "failed": true, "msg": "dopy >= 0.3.2 required for this module"}

TASK [genesis-digitalocean : fail] *********************************************
fatal: [localhost]: FAILED! => {"changed": false, "failed": true, "msg": "* The API Access Token might be incorrect or missing the Write Scope. OR * The SSH key may already exist in the DigitalOcean Control Panel under a different name. OR * The dopy Python module might not be installed. Use `pip install dopy==0.3.5` to install this module. On the latest version of OS X, dopy > 0.3.5 is currently broken."}
	to retry, use: --limit @/Users/.../projects/streisand/playbooks/digitalocean.retry
```

Going back to my master branch which points to this [commit](https://github.com/alimakki/streisand/commit/c9e95ff0105ba4ef4269d40aafc80f04ca4d9304) seems to work however.

I noticed this as well on a separate feature [branch](https://github.com/alimakki/streisand/commit/54b6d6ee9f4cd2efe1acab91036b9f5ac3473eb2) that I was working on that was recently merge with upstream/master also had the same issue.

cpu:
> Going back to my master branch which points to this commit seems to work however.

Have you tried with the current tip of master, e799e3e54ec3ef4525a40c2a0aad99e8a4e6dc41? Seems strange. I'm also testing with DigitalOcean and haven't run into that issue (yet?).

cpu:
> This seems like something shadowsocks-libev should be doing before dropping to nobody but I'll have to dig deeper. 

I think for this we'd have to drop the `User=nobody` from the upstream systemd job and instead use the `-a nobody` option to `ss-server` to let the daemon start as root, read its config, and then drop to nobody itself (haven't tested this yet).

alimakki:
Just tried the tip and I'm seeing the same thing:

```
Streisand will now set up your server. This process usually takes around ten minutes. Press Enter to begin setup...
: 

PLAY [Provision the DigitalOcean Server] ***************************************

TASK [setup] *******************************************************************
ok: [localhost]

TASK [genesis-digitalocean : Set the DigitalOcean Access Token fact to the value that was entered, or attempt to retrieve it from the environment if the entry is blank] ***
ok: [localhost]

TASK [genesis-digitalocean : Get the default SSH key] **************************
changed: [localhost]

TASK [genesis-digitalocean : Add the SSH key to DigitalOcean if it doesn't already exist] ***
fatal: [localhost]: FAILED! => {"changed": false, "failed": true, "msg": "dopy >= 0.3.2 required for this module"}

TASK [genesis-digitalocean : fail] *********************************************
fatal: [localhost]: FAILED! => {"changed": false, "failed": true, "msg": "* The API Access Token might be incorrect or missing the Write Scope. OR * The SSH key may already exist in the DigitalOcean Control Panel under a different name. OR * The dopy Python module might not be installed. Use `pip install dopy==0.3.5` to install this module. On the latest version of OS X, dopy > 0.3.5 is currently broken."}
	to retry, use: --limit @/Users/.../projects/streisand/playbooks/digitalocean.retry

localhost                  : ok=3    changed=1    unreachable=0    failed=2   
```

Is there a limit on how many droplets you can spin up/tear down on DO for a given interval?

cpu:
@alimakki Hmmm. Looking at this again, I'm wondering if you need to revert the inventory change from https://github.com/jlund/streisand/commit/89f40e80807d5275c04b26b30907543d65ab49b9#diff-7a1eabc3deb7fd02ceb1e16eafc41073

What is your host OS out of curiosity? Does adding `ansible_python_interpreter=python` back to the inventory fix this?

alimakki:
I'm on OSX El-Capitan.

Modified the `inventory` file with `localhost ansible_connection=local ansible_python_interpreter=python`.

You were correct, it's now running.

cpu:
@alimakki Great!

FYI: I just force pushed a rebase in 5c6e68e - I had put the group fix on the wrong block.

alimakki:
Spinning up a new droplet from your shadowsocks branch, going to test shortly.

cpu:
Merci beaucoup. I noticed I messed up the client mirroring and didn't see the failures. Should have that fixed in a jiffy. I always notice these things as soon as someone else is going to test!

alimakki:
@cpu De rien. 

I feel like I'm the bearer of bad news today, but shadowocks is still a no-go for me on OS X.

The server seems fine however, it's running and listening.  Tried with the latest beta client (1.5b2).

For one thing it wasn't able to scan the QR code on screen; it had to be manually inputted. I'm guessing that it's due to the beta nature.

Once it was running it was still refusing connections. This could be a quirk on my machine, I'm not sure. In any case, I added your SSH key again :)

 `ssh root@138.197.132.225`

`ECDSA key fingerprint is SHA256:J07J+dRoGzt+cfT8uYnqSSUDO55+FtglY+GOYx7gvZM.`

cpu:
> I feel like I'm the bearer of bad news today, but shadowocks is still a no-go for me on OS X.

Doh! Better to find out now before it's merged :-)

> For one thing it wasn't able to scan the QR code on screen; it had to be manually inputted. I'm guessing that it's due to the beta nature.

That's possible. I'll try to borrow a mac tomorrow to test with.

> In any case, I added your SSH key again :)

Great - I have to call it a night but I'll take a look tomorrow. Thanks again

CorbanR:
@cpu I also just grabbed the cpu-shadowsocks-libev-3.0.5-nodeb branch and couldn't seem to get GCE, nor DO to work. After reading backlog... I can confirm what  alimakki  said. Seems it's because you removed `ansible_python_interpreter=python` from `inventory`. 

FWIW I'm also on OSX

CorbanR:
I tested your latest branch `443ffd9` on GCE and Digital Ocean

OS: `OSX Sierra 10.12.4`
Ansible version: `2.2.1.0` <--- usually i'm on the latest, but had to downgrade for some reason.. and forgot to upgrade. Figured this was maybe important.

ShadowsocksX-NG: `1.5.0 (3)`

GCE:
- OSX client worked great! Scanning the QR code on screen worked great. After selecting "global" mode, I had to click on the Shadowsocks icon again, go to servers, and "select" the server. Besides that worked like a champ. 

Digital Ocean:
- OSX client. I had troubles Scanning the QR code (Error: Not valid found qrcode or shadowsocks profile). However after manually entering the info, it worked fine /shrug....

Notes:
I tried to import the GCE qrcode again, and it totally worked.. so something seems funky with the Digital Ocean qrcode (maybe just a fluke?).

Let me know if you want to troubleshoot any of these servers. I'm going to destroy them right now, but could easily spin up more servers in the morning.

cpu:
@CorbanR Thanks for the testing! Really appreciate it!

> I tried to import the GCE qrcode again, and it totally worked.. so something seems funky with the Digital Ocean qrcode (maybe just a fluke?).

Hmm. That's pretty odd! @alimakki had trouble with the QR code on OSX as well. My diff barely touched the QR code generation (just removed the OTA parameter), so I'm inclined to believe it was flaky to begin with (maybe that's optimistic!). It would be helpful if someone could save a copy of a QR code that failed to import so that we could decode it and try on a couple clients to see if its a generation bug or a parsing bug in a particular client implementation. The QR code has secret information so its best to share out of band and only for test instances.

@alimakki Any theories on why @CorbanR's OSX setup worked and yours did not?

> After reading backlog... I can confirm what alimakki said. Seems it's because you removed ansible_python_interpreter=python from inventory.

I'll put a PR out to restore that. It was clearly required for OSX hosts.

cpu:
> I'll put a PR out to restore that. It was clearly required for OSX hosts.

Fixed in master with 67d035a

alimakki:
@cpu I just spun up a new instance on DO from your branch, everything seems to be working fine.
It turns out Firefox doesn't play nice with Shadowsocks with this particular client on mac, but Chrome and Safari work flawlessly.

It still doesn't pick up the QR code from my screen, but no issues with manual input. Could be a bug in the client.

cpu:
@alimakki Thanks for the update! Glad to hear you had some success with OSX too.

> It still doesn't pick up the QR code from my screen, but no issues with manual input. Could be a bug in the client.

Do the other clients pick up the QR code?

jlund:
What does the OS X client call its cipher suite? If it isn't exactly `chacha20-ietf-poly1305` then that's probably the problem. The Cross client for iOS falls into this category as well. We can generate custom QR codes for different clients.

jlund:
I'm heavily leaning towards merging this in and iterating as we encounter edge cases. This implementation is _wildly_ better than the deprecated OTA functionality that is current in the master branch even though some clients are lagging behind in implementing AEAD support. We had to pull the trigger on the Ubuntu 16.04 upgrade even though some VPS providers couldn't support it and weren't ready to move on from 14.04 yet, but it resulted in a much better experience for the vast majority of people. I think similar logic applies here. Any objections?

alimakki:
@jlund Clients will have to update eventually, so I'm in favor of moving forward on this. We should always be pushing for more security.

cpu:
I'm not aware of any outstanding work required for this PR. Based on @jlund and @alimakki's comments I've removed the "revisions-needed" label.

I'll leave it up to other maintainers to +1 and merge when you're comfortable. Thanks!

jlund:
Thanks for all of your hard work on this, @cpu! This is a massive upgrade.

Issue #609, opened 2017-04-15
Workaround for Linode Wireguard Support Pre Ansible 2.4


cpu:
https://github.com/jlund/streisand/issues/500 identified an issue with Wireguard provisioning on Linode servers caused by the default Ansible profile using the wrong kernel option. Installation requires the "GRUB 2" kernel option.

A fix to specify GRUB2 by default was [provided by @rmcintosh](https://github.com/jlund/streisand/pull/574) but is tied to [an Ansible PR](https://github.com/ansible/ansible/pull/23100) which has only landed in devel, [and will not be available in a release until Ansible 2.4](https://github.com/jlund/streisand/issues/500#issuecomment-294304253). Ansible is on a ~4 month release schedule and so we need to evaluate interim solutions.

Brainstorming with @zx2c4 I think there are three to consider:

1. Add some kind of prompt/warning to the Wireguard role conditional to the Linode provider being used.
2. Monkey patch `api.avail_kernels()` from the Linode plugin to specify the required kernel id option
3. Update the Linode genesis role to create but not start the Linode server, issue a CURL based API call to set the kernel ID, and then start & wait on the Linode server.

**Number 1** is "easiest" but solves no real problems and punts to the user.

**Number 2** seems like it will introduce the least amount of short term role churn but needs to be tested. @zx2c4 suggests this monkey patch using the `|` yaml syntax:
```
monkey: |
	from linode import api as linode_api
	linode_api.avail_kernels = lambda: return [ { 'LABEL': 'Latest 64', 'KERNELID': 210 } ]
```

**Number 3** seems doable using the [Linode config.update API](https://www.linode.com/api/linode/linode.config.update), but the genesis role [currently creates & starts the server](https://github.com/jlund/streisand/blob/3894cbda36acd9b42a6605c99d01dd5f4d7495a2/playbooks/roles/genesis-linode/tasks/main.yml#L12:L21) as one action. This will need to be split into two using `state=present` and `wait=no` instead of `state=started` and `wait=yes`. After the API call is made the state can be expected to be started & we can wait on startup.

@zx2c4 points out that a similar curl task is already present in the role:
```
- name: Get the latest Ubuntu 16.04 image ID from the Linode API
[...]

Issue #500, opened 2017-02-04
Installation fails on Linode due to broken Linode configuration profile


vrms:
I am a novice with Streisand and exited to get it running. I hope I am not just missing anything obvious.

Trying to spin up Streisand on Linode. My local machine is a clean xubuntu 16.04 VM, prepared as per the Streisand README 

The installation fails at the same point towards the end (executing `~/playbooks/roles/wireguard/tasks/main.yml` as it seems) for several attempts in a row now

The error I am geting from the streisand install script

`TASK [wireguard : Enable the WireGuard service so it starts at boot, and bring up the WireGuard network interface] ***
fatal: [139.162.10.186]: FAILED! => {"changed": false, "failed": true, "msg": "Unable to start service wg-quick@wg0-server.service: Job for wg-quick@wg0-server.service failed because the control process exited with error code. See \"systemctl status wg-quick@wg0-server.service\" and \"journalctl -xe\" for details.\n"}`

on my linode machine's Terminal

`root@ubuntu:/# systemctl status wg-quick@wg0-server.service
● wg-quick@wg0-server.service
   Loaded: not-found (Reason: No such file or directory)
   Active: inactive (dead)`

pretty stuck with this but saw that the wireguard website issues a Warning which basically says not to "... rely on WireGuard at this stage"

did I just hit a bad moment where wireguard it's just not working? or is there any cure for this?

jlund:
Is this on a Xenial (16.04) server?

vrms:
on the linode remote server yes
`root@ubuntu:~# cat /etc/os-release 
NAME="Ubuntu"
VERSION="16.04.1 LTS (Xenial Xerus)"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 16.04.1 LTS"
VERSION_ID="16.04"
HOME_URL="http://www.ubuntu.com/"
SUPPORT_URL="http://help.ubuntu.com/"
BUG_REPORT_URL="http://bugs.launchpad.net/ubuntu/"
UBUNTU_CODENAME=xenial'

my local machine however is an xubuntu 16.04 (xfce4 Desktop)

vrms:
I have some friends @albertogalan & @rosscournoyer602  who will basically atempt the same thing today ... curiouos whether they hit the same problem

jlund:
Cool! What region? That could also be a factor. I will try to reproduce this later tonight. There weren't any issues with Linode earlier this week when I spun up some new servers there.

[...]

PR #612, opened 2017-04-16
linode: work around kernel limitation

zx2c4:
Only in the next ansible version will we have support for setting Linode's kernels directly from our playbooks. Until then, we need to monkey patch around this limitation.

Fixes: #500
Fixes: #609


cpu:
I was able to clone this branch locally & can confirm it created a fresh Linode instance with the correct kernel ID.

PR #615, opened 2017-04-16
Workaround missing Linode kernel headers

cpu:
Building on #612 3da8093 adds a workaround for Linode's grub2 instances
not having kernel headers installed. This manifests as the `wireguard-dkms`
package installing but generating no kernel modules. Later the systemctl task
will fail to bring the wg0-server interface up as a result.

This commit works around this by doing the equivalent of:
  `apt-get install linux-kernel-headers-$(uname -r)`
on the newly provisioned server, ensuring the kernel headers are there
for the wireguard role.

I've tested a from-scratch Linode provisioning and was able to configure a 
working wireguard client against a brand new Linode server once 3da8093
was applied. I believe this is the **last** hurdle for working Wireguard/Streisand
on Linode :tada:

### DIFF
--- a/playbooks/linode-headers.yml
+++ b/playbooks/linode-headers.yml
@@ -1,0 +1,23 @@
+#
+# Linode's grub2 image doesn't include kernel headers, which means the
+# wireguard-dkms package will fail to generate any kernel modules. We work
+# around this by doing the equivalent of a
+# `apt-get install linux-kernel-headers-$(uname -r)`.
+# It would be tempting to just install `linux-headers-generic` but in practice
+# this ends up installing different headers than the kernel that is running at
+# the time wireguard-dkms is installed following all the apt updates
+#
+- name: Install the Linode kernel headers
+  hosts: streisand-host
+  gather_facts: no
+  remote_user: "root"
+  become: true
+
+  tasks:
+    - name: Determine the running kernel version
+      command: uname -r
+      register: linode_kernel_version
+
+    - name: "Install linux-headers-{{ linode_kernel_version.stdout }} for Wireguard"
+      apt:
+        name: "linux-headers-{{ linode_kernel_version.stdout }}"
--- a/playbooks/linode.yml
+++ b/playbooks/linode.yml
@@ -49,5 +49,6 @@
   roles:
     - genesis-linode
 
+- include: linode-headers.yml
 
 - include: streisand.yml
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 generated-docs
 *.retry
+*.pyc


playbooks/linode-headers.yml
```diff
@@ -0,0 +1,23 @@
+#
+# Linode's grub2 image doesn't include kernel headers, which means the
```
zx2c4: "grub2 image" This doesn't make any sense. When you specify Grub2 to Linode, it's just specifying the code the CPU jumps to on startup, not some particular "OS image"

playbooks/linode-headers.yml
```diff
@@ -0,0 +1,23 @@
...
+# wireguard-dkms package will fail to generate any kernel modules. We work
+# around this by doing the equivalent of a
+# `apt-get install linux-kernel-headers-$(uname -r)`.
+# It would be tempting to just install `linux-headers-generic` but in practice
+# this ends up installing different headers than the kernel that is running at
+# the time wireguard-dkms is installed following all the apt updates
```
zx2c4: Add period to end.

playbooks/linode-headers.yml
```diff
@@ -0,0 +1,23 @@
...
+- name: Install the Linode kernel headers
+  hosts: streisand-host
+  gather_facts: no
+  remote_user: "root"
+  become: true
+
+  tasks:
+    - name: Determine the running kernel version
+      command: uname -r
+      register: linode_kernel_version
```
zx2c4: Since this is a generic playbook, no need to call it `linode_kernel_version`. A more accurate name would be `kernel_release`.

playbooks/linode-headers.yml
```diff
@@ -0,0 +1,23 @@
...
+
+    - name: "Install linux-headers-{{ linode_kernel_version.stdout }} for Wireguard"
```
zx2c4: Since this is a generic playbook, you can get rid of the ` for Wireguard` part.

(Also WireGuard, not Wireguard.)

playbooks/linode-headers.yml
```diff
@@ -0,0 +1,23 @@
+#
```
zx2c4: You should name this 'linux-headers' instead, because it might be applicable on more than just linode.

playbooks/linode.yml
```diff
@@ -49,5 +49,6 @@
   roles:
     - genesis-linode
 
+- include: linode-headers.yml
```
zx2c4: Rather than including this here, you should just include it unconditionally on the wireguard playbook. Sure some providers might already provide it, but others might not, so best to just have it in there anyway. More robust that way.

playbooks/linode-headers.yml
zx2c4: Alternatively, you could just fold it into wireguard.yml.

cpu:
Closing in favour of #616

PR #620, opened 2017-04-17
Fix the python inventory situation

zx2c4:
This patch cleans things up a bit and then dynamically figures out which python to use before launching ansible.


cpu:
@alimakki @CorbanR - Can you two test this PR on your OSX hosts? I think this is a better solution than the status quo I restored in 67d035a

alimakki:
Fix works for OS X hosts.

PR #622, opened 2017-04-17
Add basic Vagrantfile for local test instance.

cpu:
b1987b6 adds a bare bones Vagrant file that will create
a `streisand-host` VM and provision it with the project playbooks.
Nothing fancy yet!

To use, install Vagrant & `vagrant up`.

f280e0f moves the streisand-gateway tasks related to pausing
execution, prompting for continuation, and opening the generated docs in
the host browser into an ansible block. The block has a condition
attached such that it only executes when the `noninteractive` var is
anything other than `"true"`.

The Vagrantfile is updated to provide `noninteractive = "true"` as a
host inventory variable to exploit this change. Now its possible to
`vagrant up` and have the command return without needing to use
a keydown to continue through the pause.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 generated-docs
 *.retry
 *.pyc
+.vagrant/
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@ generated-docs
 *.retry
 *.pyc
 .vagrant/
+ubuntu-xenial-16.04-cloudimg-console.log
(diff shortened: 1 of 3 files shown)


cpu:
The `ansible_default_ipv4` fact that is used in a few roles (e.g. sslh) is being set incorrectly and preventing access of services outside of localhost on the vm. Trying to figure out the best approach to fixing that but until then I put a `revisions-needed` tag on here.

cpu:
Going to close this and resubmit as a larger functional testing suite on top of some CI work underway by @CorbanR.

PR #630, opened 2017-04-21
Streisand CI tests

CorbanR:
Streisand CI tests

Example Runs:
https://travis-ci.org/CorbanR/streisand


.travis.yml
```diff
@@ -0,0 +1,30 @@
...
+  - sudo apt-get install python-pip python-dev ca-certificates -qq
+
+install:
+  - pip install ansible==2.3.0.0
+  - pip install urllib3
+  - ansible --version
+  # Install additional packages, configure lxc, start streisand container
+  - ansible-playbook tests/development-setup.yml -i tests/inventory
+  # Ensure we get added to the lxd group
+  - sudo chmod 777 /var/lib/lxd/unix.socket
```
cpu: The comment makes a mention of an `lxd` group, but this `chmod` seems to be setting world read/write/execute for everyone to this socket. Should the comment be updated or the `chmod` replaced with something group related?

playbooks/roles/l2tp-ipsec/tasks/main.yml
```diff
@@ -120,12 +120,14 @@
...
 - name: Include the L2TP/IPsec sysctl value variables
   include_vars: sysctl-values.yml
+  when: ansible_virtualization_type != 'lxc'
 
 - name: Apply the sysctl values
   sysctl:
     name: "{{ item.key }}"
     value: "{{ item.value }}"
   with_items: "{{ l2tp_ipsec_sysctl_values }}"
+  when: ansible_virtualization_type != 'lxc'
```
cpu: Instead of putting two identical `when` conditions on the "Include" and "Apply" sysctl tasks could you make a block with both of these tasks inside of it and put the when condition on the block?

playbooks/roles/streisand-gateway/tasks/main.yml
```diff
@@ -101,63 +101,65 @@
...
+      local_action: command xdg-open ../{{ streisand_local_directory }}/{{ streisand_server_name }}.html
+      ignore_errors: yes
+      when: hostvars['127.0.0.1']['ansible_system'] == "Linux"
+      become: no
+
+    - name: Open the instructions on OS X (if applicable)
+      local_action: command open ../{{ streisand_local_directory }}/{{ streisand_server_name }}.html
+      when: hostvars['127.0.0.1']['ansible_system'] == "Darwin"
+      become: no
+  when: not streisand_ci
```
cpu: Is the reason that documentation can't be generated in CI that some of the roles have been removed and so the documentation fails on undefined variables? I believe that's likely the case but wanted to confirm. It would be nice to put a follow-up issue on the repo that would fix the documentation inter-dependencies since we will have to do that for the overall "modularity" work and it would also mean that CI could have docs generated. Testing docs in CI would be another great win!

playbooks/streisand.yml
```diff
@@ -31,7 +31,8 @@
     - shadowsocks
     - ssh
     - tinyproxy
-    - tor-bridge
+    - role: tor-bridge
+      when: not streisand_ci
```
cpu: I would maybe add a brief comment here about why we don't run this role under CI similar to what's in `tests/README.md`.

tests/README.md
```diff
@@ -0,0 +1,31 @@
+# Streisand-CI testing
+Streisand CI testing. Uses travis-ci to test Ansible syntax and kick off a full playbook run.
+Full playbook run uses LXC to execute.
+
+## Limitations
+The TOR playbook role is skipped. Turns out the travis-ci environment doesn't like being added to the tor network (even temporarily for testing purposes).
```
cpu: "TOR" -> "Tor"

tests/README.md
```diff
@@ -0,0 +1,31 @@
...
+Streisand doc generation is also skipped for now until some mock variables are added for the skipped TOR role.
+
+## Local Testing
+You can use this testing framework locally as well (Locally, the TOR playbook will be ran).
```
cpu: "TOR" -> "Tor"

tests/README.md
```diff
@@ -0,0 +1,31 @@
...
+
+Environment Expectations:
```
cpu: When you say Environment here you mean the host environment that will be running Ansible? Can you clarify that since experience shows people are easily confused about which machine you're talking about when you're dealing with Ansible, containers, and Streisand servers.

tests/README.md
```diff
@@ -0,0 +1,31 @@
...
+  - *Fresh* trusty or xenial ubuntu install (Vagrant or digital ocean server, etc.)
+
+Running the tests:
+Since LXC loads the hosts kernal modules, Libreswan is compiled and installed on the host. This allows the ipsec role to complete successfully
```
cpu: "the hosts" -> "the host's" ?

tests/README.md
```diff
@@ -0,0 +1,31 @@
...
+
+  Running all tests:
+    - Clone the Streisand repository
+    - Run test script
+      - `./tests/test.sh full`
+        - This will setup LXD/LXC and dependencies locally. Then run streisand against the streisand LXC container.
```
cpu: I think these two sentences should be combined?
"- This will setup LXD/LXC and dependencies locally and then run streisand against the LXC container."

tests/README.md
```diff
@@ -0,0 +1,31 @@
...
+  Sytanx only:
+    - `./tests/test.sh syntax` 
+
+# TODO
+- Vagrant file which can automate setting up a local test environment for use with development-setup.yml, and run.yml
+- Figure out a way to test the TOR role
```
cpu: TOR -> Tor

tests/README.md
```diff
@@ -0,0 +1,31 @@
...
+- Add some mock variables for the document generation task
+- Would be cool to automate tests with each provider( Example: Automatically spinning up an EC2 instance and testing)
```
cpu: "provider(" -> "provider ("

tests/development-setup.yml
```diff
@@ -0,0 +1,148 @@
...
+      pause:
+        seconds: 5
+
+    - name: Debug ipsec start
+      debug:
+        msg: "{{ ipsec_start }}"
+
+    - name: Ipsec verify
+      command: ipsec verify
+      ignore_errors: yes
```
cpu: Why do you need `ignore_errors` here?

tests/development-setup.yml
```diff
@@ -0,0 +1,148 @@
...
+  remote_user: root
+  become: yes
+  tasks:
+    - block:
+        - name: Install lxc ppa key
+          apt_key:
+            id: 7635B973
+            url: "http://keyserver.ubuntu.com:11371/pks/lookup?op=get&search=0xD5495F657635B973"
+            state: present
+          tags: deluge
```
cpu: Why are these tagged "deluge"?

tests/development-setup.yml
```diff
@@ -0,0 +1,148 @@
...
+            - "deb-src http://ppa.launchpad.net/ubuntu-lxc/lxd-stable/ubuntu {{ansible_distribution_release}} main"
+          tags: deluge
+
+        - name: Install the Libreswan dependencies that are required for compilation
+          apt:
+            name: "{{ item }}"
+          with_items: "{{ libreswan_compilation_dependencies_trusty }}"
+          when: ansible_distribution == 'Ubuntu' and ansible_distribution_release == 'trusty'
+
+      #Since LXC loads kernal modules from the host, we need to ensure libreswan is compiled and installed
```
cpu: "#Since" -> "# Since"

tests/development-setup.yml
```diff
@@ -0,0 +1,148 @@
...
+        name: ipsec
+        state: started
+        enabled: yes
+      register: ipsec_start
+
+    - name: Pause for a few seconds to let ipsec start
+      pause:
+        seconds: 5
+
+    - name: Debug ipsec start
```
cpu: Was this leftover from testing or included on purpose?

tests/development-setup.yml
```diff
@@ -0,0 +1,148 @@
...
+        name: lxd
+        state: present
+        update_cache: yes
+
+    - name: lxd new group
+      shell: newgrp lxd
+
+    - name: lxd init config
+      shell: lxd init --auto --storage-backend dir
+      ignore_errors: yes
```
cpu: Is ingore_errors required here?

tests/development-setup.yml
```diff
@@ -0,0 +1,148 @@
...
+
+    - name: lxd create network
+      shell: lxc network create testbr0
+      ignore_errors: yes
```
cpu: ditto RE: ignore_errors

tests/development-setup.yml
```diff
@@ -0,0 +1,148 @@
...
+
+    - name: lxd attach network to default profile
+      shell: lxc network attach-profile testbr0 default eth0
+      ignore_errors: yes
```
cpu: ditto RE: ignore_errors

tests/development-setup.yml
```diff
@@ -0,0 +1,148 @@
...
+        state: restarted
+
+    - name: pause
+      pause:
+        seconds: 10
+
+    - name: set permission on lxd daemon
+      file:
+        path: /var/lib/lxd/unix.socket
+        mode: "0777"
```
cpu: This is also done in the travis file for CI, does it need to happen there if it also happens here?

tests/run.yml
```diff
@@ -0,0 +1,52 @@
+---
+#Run main playbook
```
cpu: "#Run" -> "# Run"

tests/run.yml
```diff
@@ -0,0 +1,52 @@
...
+- hosts: streisand
+  gather_facts: no
+  remote_user: root
+  become: yes
+  pre_tasks:
+    - name:
```
cpu: Does this blank `name:` need to be here?

tests/syntax-check.yml
```diff
@@ -0,0 +1,66 @@
+---
+#Include each playbook in order to check syntax of each playbook
```
cpu: "#Include" -> "# Include"

tests/tests.sh
```diff
@@ -0,0 +1,28 @@
...
+  exit 1
+}
+
+required_ansible_version="2.3.0.0"
+
+if [[ "$(ansible --version | grep -oe '2\(.[0-9]\)*')" < $required_ansible_version ]]
+then
+    echo "Ansible $required_ansible_version or higher is required."
+    exit 1
+fi
```
cpu: Is it possible to abstract out the common parts of this and the main streisand shell script so that it can be sourced here? It would be nice to not have to keep the `required_ansible_version` in sync for one example.

tests/run.yml
```diff
@@ -0,0 +1,52 @@
...
+      raw: sudo apt update && apt install python -y
+
+- hosts: streisand
+  gather_facts: yes
+  remote_user: root
+  become: yes
+  tasks:
+    - name: Get the default SSH key
```
cpu: I'm slightly nervous about having a SSH private key checked into the repo that's used by default for the `tests/run.yml` setup. One day someone might end up doing something silly and leaving a container running that's internet exposed accessible with a published SSH private key.

Would it be possible to generate a unique SSH key per container & use that? Is there a reason you preferred this approach I'm overlooking?

.travis.yml
CorbanR: Ahh yeah, I should update the comment. The chmod 777 is a requirement to get this working on Travis.

playbooks/roles/l2tp-ipsec/tasks/main.yml
CorbanR: Agreed

playbooks/roles/streisand-gateway/tasks/main.yml
CorbanR: Yeah, I added some notes in the README.md about this, it's part of the TODO.

playbooks/streisand.yml
CorbanR: Agreed

tests/development-setup.yml
CorbanR: hahah woops., Used an example of a different playbook I have written ;)

tests/development-setup.yml
CorbanR: This was mostly for debugging purposes. It will almost always return with an error. I will take it out.

tests/development-setup.yml
CorbanR: if you were to run the development-setup.yml on a test machine more than once, this would error.. I figured it was ok to ignore it.

tests/development-setup.yml
CorbanR: if you were to run the development-setup.yml on a test machine more than once, this would error.. I figured it was ok to ignore it.

tests/development-setup.yml
CorbanR: if you were to run the development-setup.yml on a test machine more than once, this would error.. I figured it was ok to ignore it.

tests/development-setup.yml
CorbanR: nope, duplicate.

tests/run.yml
CorbanR: You are 100% right about generating a unique key. We totally could do that. I will update.

tests/tests.sh
CorbanR: Hmm ill have to think about the best way to do that.  I was considering doing something like

```- name: Validate ansible version
  assert:
    that:
      - ansible_version is defined
      - ansible_version.full|version_compare('2.3.0.0', '>=')
    msg: "You must have ansible >= 2.3.0.0 to run this specific playbook"
  run_once: yes```

and taking it out from both scripts.. but we would want the version check to happen before people run just about anything.

Comment (cpu) [inline tests/tests.sh]: I was thinking something simpler like putting the common lines in a .sh file that both the striesand script and the test script source directly.

File: tests/development-setup.yml (hunk 16 above)
Comment (cpu) [inline tests/development-setup.yml]: Ahh! Is there a way to make these tasks idempotent somehow so they won't run if they've already created whatever side-effect/artifact they are required for?

File: playbooks/roles/streisand-gateway/tasks/main.yml (hunk 3 above)
Comment (CorbanR) [inline playbooks/roles/streisand-gateway/tasks/main.yml]: https://github.com/jlund/streisand/issues/643

File: tests/development-setup.yml (hunk 16 above)
Comment (CorbanR) [inline tests/development-setup.yml]: Ya know, I did a quick search and as far as I can tell.. not really :(. We could run an lxd info command and based on the status skip or run a task.

File: tests/development-setup.yml (hunk 18 above)
Comment (CorbanR) [inline tests/development-setup.yml]: For the time being I at least put them in a block.

File: tests/group_vars/all/all [hunk 25]
```diff
@@ -0,0 +1,26 @@
...
+  - libcap-ng-dev
+  - libcap-ng-utils
+  - libcurl3-nss-dev
+  - libevent-dev
+  - libnspr4-dev
+  - libnss3-dev
+  - libnss3-tools
+  - libpam0g-dev
+  - libselinux1-dev
+  - libsystemd-daemon-dev
```
cpu: On a fresh 64bit Ubuntu 16.04 machine this fails for me:
> `"No package matching 'libsystemd-daemon-dev' is available"`

```
# apt-get install libsystemd-daemon-dev
Reading package lists... Done
Building dependency tree       
Reading state information... Done
Package libsystemd-daemon-dev is not available, but is referred to by another package.
This may mean that the package is missing, has been obsoleted, or
is only available from another source
However the following packages replace it:
  libsystemd-dev:i386 libsystemd-dev

E: Package 'libsystemd-daemon-dev' has no installation candidate
```

Would `libsystemd-dev` be the right replacement?

tests/development-setup.yml
```diff
@@ -0,0 +1,137 @@
...
+          type: image
+          mode: pull
+          server: https://images.linuxcontainers.org
+          protocol: lxd
+          alias: ubuntu/xenial/amd64
+        profiles: ["default"]
+        config:
+          security.privileged: "true"
+        wait_for_ipv4_addresses: true
+        timeout: 300
```
cpu: It might be worth adding a note to the task name here that this can take a long time. It was a pretty considerable wait (>60s? ) for me on initial run.

tests/run.yml
```diff
@@ -0,0 +1,57 @@
...
+            creates: "/tmp/id_rsa_insecure"
+
+        - name: Get the default SSH key
+          command: "cat /tmp/id_rsa_insecure.pub"
+          register: ssh_key
+
+        - name: Ensure permissions on insecure key are correct
+          file:
+            path: "/tmp/id_rsa_insecure"
+            mode: "0666"
```
cpu: I think this needs to be scoped to CI only - see my earlier comment.

tests/group_vars/all/all
CorbanR: Hmm that should be scoped to a trusty install only.

CorbanR: Found the issue and fixed :)! Nice catch!

cpu: Awesome!

cpu:
@alimakki gave this a +1 in a side-channel. Going to merge!

PR #632, opened 2017-04-22
wireguard: automatically set MTU from default route

zx2c4:
This is needed for some strange providers who give non-standard MTUs.
This is fairly rare on the client side, but on the server side, Google
Cloud Platform gives 1460, presumably due to vxlan or some other SDN
they deploy.

Also, while we're at it, since Streisand doesn't support IPv6 at the
moment (even though WireGuard does), we can get an extra 20 bytes in our
MTU by subtracting 60 instead of 80.

**I haven't tested this commit, so please make sure this actually works before merging.**


jlund:
I just tested this on a Ubuntu 16.04 client/server combo and also did not encounter any issues. This looks good to me too. Thanks for your help!

PR #634, opened 2017-04-22
Add GCE missing zones

x8x:


x8x:
https://cloud.google.com/compute/docs/regions-zones/regions-zones

CorbanR:
@x8x Thanks!

PR #647, opened 2017-04-25
Update torrc.j2

amsqube:
In this case "private" means that the bridge is configured with the option PublishServerDescriptor 0.  Without this option set, The Tor Project can learn about the bridge and may distribute its address to others.
https://www.whonix.org/wiki/Bridges#Additional_Information_and_Recommendations

### DIFF
--- a/playbooks/roles/tor-bridge/templates/torrc.j2
+++ b/playbooks/roles/tor-bridge/templates/torrc.j2
@@ -2,6 +2,7 @@ SocksPort 0
 ORPort {{ tor_orport }}
 ExtORPort auto
 BridgeRelay 1
+PublishServerDescriptor 0
 ExitPolicy reject *:*
 
 Nickname {{ tor_bridge_nickname.stdout }}


playbooks/roles/tor-bridge/templates/torrc.j2
```diff
@@ -2,6 +2,7 @@ SocksPort 0
 ORPort {{ tor_orport }}
 ExtORPort auto
 BridgeRelay 1
+PublishServerDescriptor 0
```
cpu: Can you make the value of this setting be populated from a variable in the [defaults.yml](https://github.com/jlund/streisand/blob/677d255849b4167b3390f77b1425bc24e476829a/playbooks/roles/tor-bridge/defaults/main.yml) for the `tor-bridge` recipe, similar to `{{ tor_orport }}` above? 

Keeping the value as `0` is great as the default but it would be nice to add a comment explaining that if you would like to opt-in to contributing a bridge to the Tor network that you should use the value `bridge` instead of `0`. Please also briefly describe what "0" does to save folks a manual read :-) 

Thanks!

cpu:
Created https://github.com/jlund/streisand/issues/655 as a potential follow-up.

cpu:
Hi @amsqube - Do you have any interest in addressing the feedback I left so far? If you're busy I'd be happy to adopt this PR and push the final touches onto it myself. I'll leave it be for now, please chime in with your preferences :-)

cpu:
@amsqube I wasn't able to push commits onto this existing PR so I cherry-picked your base commit into a new PR (https://github.com/jlund/streisand/pull/701) so that the credit remains with you but the PR can move foward.

I'm going to close this PR since it's been replaced by #701. Thanks again for the contribution!

PR #649, opened 2017-04-26
Remove explicit `brew update`

leppert:
Homebrew auto-updates on `brew install` now, and this change has had about a year to propagate.
https://github.com/Homebrew/brew/pull/679


alimakki:
Thank you @leppert :)

PR #660, opened 2017-04-30
Vagrant dev env support & basic end-to-end service tests

cpu:
This PR adds a `Vagrantfile` that will create & provision two VMs at fixed IP addresses: `streisand-host` and `streisand-client`.

The host machine is provisioned using the base Streisand playbooks with very little modification (primarily modified to allow a non-interactive promptless mode). One workaround is in place in `playbooks/vagrant.yml` to ensure that the correct interface is used for the `ansible_default_ipv4` var.

Along the way I also updated the SSH role to generate a known-hosts file with SSH fingerprints for the server. The fingerprints file is put in the webroot for clients to download instead of doing TOFU SSH fingerprint matching themselves.

The client machine is provisioned with a new `test-client` role. The goal of this role is to provision a machine that downloads resources from the Streisand host's HTTPS gateway and uses those resources to test the various services end-to-end.

Initially I have added client support for:
* testing parts of the gateway (HTTP auth, etc)
* testing the SSH forward user (SOCKS proxy, key auth, using the known hosts file) and
* testing Shadowsocks (installing the Go client from the host's mirror, connecting to the shadowsocks-libev server and proxying a request).

The tests are implemented as bash scripts instead of Ansible tasks largely because they involve setting up and tearing down proxies and the complexity merits having a more flexible environment (I think).


playbooks/roles/streisand-client/vars/main.yml
```diff
@@ -0,0 +1,48 @@
...
+gateway_test_url: "https://{{ streisand_ip }}"
+# User for accessing the streisand server HTTP gateway page
+gateway_test_user: "streisand"
+
+# The directory to copy test scripts to
+test_script_dir: "/usr/local/bin/"
+
+# Test scripts to install & run
+test_scripts:
+  - { template: 'streisand-gateway-test.sh.j2', cmd: 'streisand-gateway-test' }
```
jlund: I (maybe pointlessly?) format multi-line YAML dicts so the variable pairs are spaced to line up. Is it actually worthwhile? We can pretend.

```
  - { template: 'streisand-gateway-test.sh.j2',             cmd: 'streisand-gateway-test' }
  - { template: 'streisand-ssh-forward-test.sh.j2',         cmd: 'streisand-ssh-forward-test' }
  - { template: 'streisand-shadowsocks-forward-test.sh.j2', cmd: 'streisand-shadowsocks-forward-test' }
```

jlund: Another possible approach:

```
test_scripts:
  - streisand-gateway-test
  - streisand-ssh-forward-test
  - streisand-shadowsocks-forward-test
```


```
- name: "Install test scripts"
  template:
    src: "{{ item }}.sh.j2"
    dest: "{{ test_script_dir }}/{{ item }}"
    owner: root
    group: root
    mode: 0755
  with_items: "{{ test_scripts }}"
```

playbooks/roles/streisand-client/vars/main.yml
```diff
@@ -0,0 +1,48 @@
...
+  - { template: 'streisand-ssh-forward-test.sh.j2', cmd: 'streisand-ssh-forward-test' }
+  - { template: 'streisand-shadowsocks-forward-test.sh.j2', cmd: 'streisand-shadowsocks-forward-test' }
+
+# The gateway URL for clients to download the forward user's RSA key
+gateway_ssh_rsa_key: "{{ gateway_test_url }}/ssh/streisand_rsa"
+# The gateway URL for clients to download the SSH known hosts file
+gateway_ssh_hosts: "{{ gateway_test_url }}/ssh/streisand.known_hosts"
+
+# Files created on the streisand client test for SSH config:
```
jlund: streisand -> Streisand

playbooks/roles/streisand-client/vars/main.yml
```diff
@@ -0,0 +1,48 @@
...
+forward_ssh_config: "/root/.ssh/config"
+forward_ssh_hosts: "/root/.ssh/streisand.known_hosts"
+
+# Forwarding SSH username
+forward_ssh_user: "forward"
+
+# Port on the streisand-client to use for the SOCKS SSH proxy
+forward_socks_port: 9876
+
+# Path to the shadowsocks password on the Ansible host
```
jlund: shadowsocks -> Shadowsocks

playbooks/roles/streisand-client/vars/main.yml
```diff
@@ -0,0 +1,48 @@
...
+shadowsocks_password_localpath: "../{{ streisand_local_directory }}/shadowsocks-password.txt"
+
+# Path to the shadowsocks password file on the streisand client
```
jlund: shadowsocks -> Shadowsocks
streisand -> Streisand

playbooks/roles/streisand-client/vars/main.yml
```diff
@@ -0,0 +1,48 @@
...
+shadowsocks_password_file: "/root/shadowsocks-password.txt"
+
+# The gateway URL for clients to download the shadowsocks2 linux source
+gateway_shadowsocks_client: "{{ gateway_test_url }}/mirror/shadowsocks/shadowsocks2-linux-x64.gz"
+
+# Path to the shadowsocks2 client source on the streisand client
```
jlund: streisand -> Streisand

playbooks/roles/streisand-client/vars/main.yml
```diff
@@ -0,0 +1,48 @@
...
+# Port on the streisand-client to use for the SOCKS SSH proxy
+forward_socks_port: 9876
+
+# Path to the shadowsocks password on the Ansible host
+shadowsocks_password_localpath: "../{{ streisand_local_directory }}/shadowsocks-password.txt"
+
+# Path to the shadowsocks password file on the streisand client
+shadowsocks_password_file: "/root/shadowsocks-password.txt"
+
+# The gateway URL for clients to download the shadowsocks2 linux source
```
jlund: Hooray for the Go client!

playbooks/roles/streisand-gateway/tasks/main.yml
```diff
@@ -70,6 +70,13 @@
     - "{{ nginx_private_key }}"
     - "{{ streisand_gateway_password_file }}"
 
+- name: Copy the streisand gateway password file locally
```
jlund: streisand -> Streisand

jlund: The same question of whether or not this should be Vagrant-specific and applied using a conditional would also apply here (as it did for the Shadowsocks password).

playbooks/roles/streisand-gateway/vars/main.yml
```diff
@@ -3,6 +3,7 @@ streisand_gateway_url: "https://{{ streisand_ipv4_address }}"
 
 streisand_gateway_htpasswd_file: "/etc/nginx/htpasswd"
 streisand_gateway_password_file: "/etc/nginx/gateway-password.txt"
+streisand_gateway_password_localpath: "../{{ streisand_local_directory }}/gateway-password.txt"
```
jlund: If the local fetch isn't scoped to just Vagrant, we would want this file to have some kind of server identifier in its name as well (a la the Shadowsocks password localpath comment).

playbooks/streisand-client.yml
```diff
@@ -0,0 +1,20 @@
+---
```
jlund: Should we name these new playbook files something else? Maybe `vagrant-testing-streisand-client.yml` is too ugly/long, but I could see the current filename causing some confusion for people who run Streisand on non-native providers outside of the wrapper script--especially when tab complete will suggest this as a possibility given its similarity to `streisand.yml`.

playbooks/streisand-client.yml
```diff
@@ -0,0 +1,20 @@
...
+  remote_user: "root"
+  become: true
+
+  tasks:
+    - name: Install Python using a raw SSH command to enable the execution of Ansible modules
+      raw: apt update && apt install python -y
+      args:
+        executable: /bin/bash
+
+- name: Configure the client server as a streisand client
```
jlund: streisand -> Streisand

playbooks/roles/streisand-client/tasks/main.yml
```diff
@@ -0,0 +1,95 @@
...
+  copy:
+    src: "{{ shadowsocks_password_localpath }}"
+    dest: "{{ shadowsocks_password_file }}"
+    owner: root
+    group: root
+    mode: 0600
+
+- name: "Download the forward user SSH key to the client"
+  get_url:
+    url: "{{ gateway_ssh_rsa_key }}"
```
cpu: I'm game. Sounds clearer. Thanks!

playbooks/roles/shadowsocks/tasks/main.yml
```diff
@@ -91,6 +91,13 @@
   command: cat {{ shadowsocks_password_file }}
   register: shadowsocks_password
 
+- name: Copy the Shadowsocks password file locally
```
cpu: That's a good idea. I don't think there's any benefit outside of Vagrant for the time being. I'll have to add a var that gets set for vagrant runs as well but that will likely end up being useful anyway~!

playbooks/vagrant.yml
```diff
@@ -0,0 +1,20 @@
+- include: python.yml
+
+- name: Prepare the vagrant VM for Ansible
+# =========================================
+  hosts: streisand-host
+  remote_user: "root"
+  become: true
+
+  tasks:
+    # Ansible uses `ip -4 route get 8.8.8.8` to set the `ansible_default_ipv4`
```
cpu: Thanks! The workaround _seems_ stable on my machine, hopefully it won't bite us in the butt down the road.

playbooks/roles/streisand-client/templates/ssh-config.j2
```diff
@@ -0,0 +1,18 @@
+# Connection profile for a SOCKS proxy over a SSH connection to the forward user
```
cpu: Totally agree - I meant to create an issue for a follow-up to migrate this config to the server's SSH role. Created https://github.com/jlund/streisand/issues/675 for this & assigned it to myself with a blocked label pending this PR landing in master.

playbooks/roles/streisand-client/templates/streisand-shadowsocks-forward-test.sh.j2
```diff
@@ -0,0 +1,38 @@
+#!/bin/bash -xe
+#
+# NOTE: This test script relies on the -e argument to bash in the shebang above.
+#
+# Streisand Shadowsocks Client Test
+#   - Confirms the gateway mirror for the linux Shadowsocks client works
+#   - Confirms that the Linux shadowsocks client can connect & route a HTTP request
+#
+# TODO: This script makes an external request to google.com - this should
```
cpu: WFM! Good idea

playbooks/roles/streisand-client/vars/main.yml
cpu: I'm happy to reformat to align columns. I think the second approach will be more duplication since the current `test_scripts` var entries are used in a couple places.

playbooks/roles/streisand-gateway/tasks/main.yml
cpu: I'll put this behind the vagrant only var as well. No point doing it except for Vagrant (so far) since it's embedded into the generated docs index for human consumption.

playbooks/streisand-client.yml
cpu: I had intentions of changing the playbook/role name but wasn't able to think of anything great. I agree that it's likely to cause confusion. I'd like to get "test" in there somehow. I don't have time to tackle these revisions for another day or so, I'll strew on this and try to think of something.

cpu:
@jlund Thanks for the review notes - I don't think there's anything controversial there. I put a `status/revisions-needed` label on this for now. I'll incorporate the changes requested & ping you for a follow up in the next few days.

playbooks/streisand-client.yml
alimakki: `vagrant-client.yml`?

playbooks/roles/shadowsocks/tasks/main.yml
cpu: Made this conditional behind `streisand_client_test`

playbooks/roles/shadowsocks/vars/main.yml
```diff
@@ -30,6 +30,7 @@ shadowsocks_configure_flags: "--disable-documentation --prefix=/usr"
 
 shadowsocks_location: "/etc/shadowsocks-libev"
 shadowsocks_password_file: "{{ shadowsocks_location }}/shadowsocks-password.txt"
+shadowsocks_password_file_localpath: "../{{ streisand_local_directory }}/shadowsocks-password.txt"
```
cpu: > If we decide to make the creation of this file subject to a conditional so it is only present for Vagrant runs, then this won't really apply

:+1:  - the fetch is conditional now.

>  if we plan on pulling down more local files we can do server-specific subdirectories too.

I think that's probably a good follow up issue.

playbooks/roles/streisand-gateway/vars/main.yml
cpu: The local fetch is conditional now, like the shadowsocks file.

playbooks/streisand-client.yml
cpu: How about this: I use `test-client.yml` since its generic to end-to-end testing and not specific to vagrant. The vagrant specific pieces stay in `vagrant.yml`. 

The next immediate question is "If it's for testing, why isn't it in the `test/` directory with Corban's CI work". The answer to _that_ is that I don't want to use the inventory and other settings files that he has in the top level of that dir. I think a good follow-up issue would be to move some of that to `test/ci` and then make `test/vagrant` and put `test-client.yml` in there.

Sounds good to you folks?

cpu:
@jlund I believe this is ready for another :mag: when you have a chance. Thanks!

PR #661, opened 2017-04-30
Updates REQUIRED_ANSIBLE_VERSION to 2.3.0.0

cpu:
Having discussed this with other maintainers it sounds like bumping the REQUIRED_ANSIBLE_VERSION that the streisand wrapper checks is fairly painless since its generally easy for folks to update their Ansible installs to the latest via pip/etc.

With that in mind this PR bumps the REQUIRED_ANSIBLE_VERSION to 2.3.0.0 ahead of using some of the features it provides. It seems prudent to keep this process fluid by doing it frequently.

### DIFF
--- a/streisand
+++ b/streisand
@@ -3,7 +3,7 @@
 
 set -e
 
-REQUIRED_ANSIBLE_VERSION="2.2.1.0"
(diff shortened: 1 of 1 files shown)


jlund:
Working well on Linux too.

PR #670, opened 2017-05-07
Shorten required version so comparison succeeds.

GabeIsman:
I noticed that the `REQUIRED_ANSIBLE_VERSION` of `2.3.0.0` causes the ansible version check to fail with `2.3.0` installed. This fixes it.

### DIFF
--- a/streisand
+++ b/streisand
@@ -3,7 +3,7 @@
 
 set -e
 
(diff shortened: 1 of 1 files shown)


jlund:
Thanks!

PR #674, opened 2017-05-08
Update the Shadowsocks instructions to remove the Potatso AEAD warning.

jlund:
New versions of Potatso 2 now support AEAD ciphers.


cpu:
@jlund Glad to hear Potatso is updated! Thanks for the fix.

PR #677, opened 2017-05-09
Internationalization - Français/French 🇫🇷

alimakki:
Streisand parle français!

Initial commit for French.

### DIFF
--- a/playbooks/roles/l2tp-ipsec/templates/instructions-fr.md.j2
+++ b/playbooks/roles/l2tp-ipsec/templates/instructions-fr.md.j2
@@ -1,0 +1,121 @@
+L2TP/IPsec
+----------
+
+L2TP/IPsec est un excellent premier choix. Il est nativement pris en charge par Android, iOS, OS X et Windows. Il n'y a pas de logiciel supplémentaire à installer. L'installation ne devrait prendre que quelques minutes.
+
+Si la connexion semble instable, vérifiez d'abord que les mots de passe ont été entrés correctement. Si vous ne pouvez toujours pas vous connecter, il est possible que votre pays ou votre FAI bloque le protocole L2TP/IPsec. Ne vous inquiétez pas! les méthodes de connexion [OpenVPN (stunnel)](/openvpn/stunnel.html), [Shadowsocks](/shadowsocks/), [SSH](/ssh/) et [Tor](/tor/) devraient toujours vous servir.
+
+---
+* Plateformes
+  * [Windows](#windows)
+  * [OS X](#osx)
+  * [Linux](#linux)
+  * [Androïde](#android)
+  * [iOS](#ios)
+  * [Chromebook](#chromebook)
+
+<a name="windows"></a>
+### Windows ###
+1. Cliquez sur le menu *Démarrer* et sélectionnez le panneau de configuration.
+1. Accédez à la section *Réseau et Internet*.
+1. Cliquez sur *Afficher l'état du réseau*.
+1. Cliquez sur *Connexion à un réseau*.
+1. Sélectionnez *Connexion à votre espace de travail*.
+1. Cliquez *Utiliser ma connexion Internet (VPN)*.
+1. Saisissez `{{ streisand_ipv4_address }}` dans le champ *Adresse Internet*.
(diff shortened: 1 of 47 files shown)


playbooks/roles/genesis-digitalocean/tasks/main.yml
```diff
@@ -1,4 +1,4 @@
----
+---
```
cpu: spurious diff

tests/README.md
```diff
@@ -3,7 +3,7 @@ Streisand CI testing uses Travis CI to test Ansible syntax and kick off a full p
 The Full playbook run is executed against an LXC container.
 
 ## Testing Limitations
-There were some connectivity issues with the `tor-bridge` playbook in the Travis CI testing environment. Due to this, the role is curretnly skipped. 
+There were some connectivity issues with the `tor-bridge` playbook in the Travis CI testing environment. Due to this, the role is curretnly skipped.
```
cpu: I don't think this is your typo but can you fix "curretnly" -> "currently"?

playbooks/roles/streisand-mirror/vars/tor.yml
```diff
@@ -12,40 +12,92 @@ tor_base_download_url: "https://dist.torproject.org/torbrowser"
 # Windows
 tor_browser_bundle_windows_filename: "torbrowser-install-{{ tor_browser_bundle_version }}_en-US.exe"
 tor_browser_bundle_windows_sig_filename: "{{ tor_browser_bundle_windows_filename }}.asc"
+
+tor_browser_bundle_windows_filename_fr: "torbrowser-install-{{ tor_browser_bundle_version }}_fr.exe"
```
cpu: It seems like (at least for Tor) the difference in these URLs is just the suffix: `"_en-US"` or `"_fr"`.

Is there any way to make this diff a little bit smaller? With each of these vars having a `_fr` equiv I'm worried if we added three more languages then each var would be duplicated three more times.

playbooks/roles/l2tp-ipsec/templates/instructions-fr.md.j2
```diff
@@ -0,0 +1,127 @@
+- - -
+{% for key, value in streisand_languages.iteritems() %}
+  [{{ value.language_name }}](index{{ value.file_suffix }}.html)&nbsp;
+{% endfor %}
```
cpu: This header fragment is added in all of the .md template files for both languages. Is it possible to put it in its own `.md.j2` file and `{% include %}` it everywhere? I haven't tried it. If it's a huge pain its probably not worth it but it would be nice if it works easily!

playbooks/roles/streisand-gateway/tasks/main.yml
```diff
@@ -101,43 +101,75 @@
   command: base64 --wrap=0 {{ openssl_ca_certificate }}
   register: streisand_certificate_data_uri
 
-  # TODO:
-  #   Add to CI testing https://github.com/jlund/streisand/issues/643
```
cpu: Can you re-add this TODO comment? I think it still applies since this block is gated on the CI var.

playbooks/roles/l2tp-ipsec/templates/instructions-fr.md.j2
alimakki: That's a good suggestion; I think most can be, but there are some exceptions such as `streisand-gateway/templates/instructions*.md.j2`, where the filename doesn't start with `index`. I suspect those will be exceptions to the rule, unless a better approach is presented.

playbooks/roles/streisand-mirror/vars/tor.yml
alimakki: I'm gonna have to think about this one.

alimakki:
@cpu Thanks for the quick initial feedback 👍

playbooks/roles/streisand-mirror/vars/tor.yml
cpu: Ok. I'm happy to punt this to a separate issue for now since it's pretty theoretical when there's only one translation :-)

PR #687, opened 2017-05-18
tweaks the l2tp settings to work with iOS

designfrontier:
The builtin vpn on iOS didn't play well with the settings in the
l2tp playbook. This adds a couple settings to make that happen.


jlund:
Thanks!

PR #696, opened 2017-05-20
Work around `resolvconf` non-zero exit in wg client PostUp.

cpu:
The Wireguard `wg0-client.conf` template contains a `PostUp` hook that
uses `resolvconf` to set the DNS server for the client. On Ubuntu
clients, the `openresolv` package that the Wireguard PPA depends on to
provide the `resolvconf` command has an interesting "quirk" where it
will try to restart network services that may not be installed. Since
these attempted restarts fail, the overall `resolvconf` command returns
a non-zero exit status. The `wg-quick` function that invokes the
`PostUp` handler, `execute_hook()`, considers a non-zero exit code for
a PostUp as a fatal error preventing the Wireguard interface from being
created successfully.

Interestingly, a subsequent attempt to bring up the interface with
`wg-quick` succeeds, likely because `openresolv` doesn't need to update
the DNS servers or restart dependent services because it already did so
on the first non-zero exit code invocation.

This commit applies a dirty hack by appending `|| true` to the `PostUp`
in the `wg0-client.conf` template after the `resolvconf` invocation.
This ignores any failures from this command which isn't ideal, but
allows the interface to come up on the first try for Ubuntu clients.

### DIFF
--- a/playbooks/roles/wireguard/templates/wg0-client.conf.j2
+++ b/playbooks/roles/wireguard/templates/wg0-client.conf.j2
@@ -5,7 +5,12 @@ Address = 10.192.122.2/32
 #
 # The IP address of the DNS server that is available via the encrypted
 # WireGuard interface is {{ dnsmasq_wireguard_ip }}.
-PostUp = echo nameserver {{ dnsmasq_wireguard_ip }} | resolvconf -a %i -m 0 -x
+#
+# Note: You may see messages about "Failed to try-restart" services that you do
+# not have installed after the `resolvconf` PostUp. This is a side effect of the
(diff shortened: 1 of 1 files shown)


zx2c4:
This isn't a good way of going about things. I'll try to fix the DNS stuff within wg-quick itself for the next revision of WireGuard.

cpu:
> This isn't a good way of going about things.

A dirty hack that does what it says on the tin :-)

> I'll try to fix the DNS stuff within wg-quick itself for the next revision of WireGuard.

That would be great. I'll close this PR, the hack is here if it bothers Ubuntu users in the meantime.

PR #707, opened 2017-05-25
shadowsocks: listen on streisand_ipv4_address

yousong:


cpu:
Hi @yousong,

Thanks for the PR. Can you explain the motivation? Did you find that the existing `wait_for` was causing you problems? If so can you describe which hosting provider you're using and how you installed Streisand?

It seems like the fact that [`config.json.j2` uses `ansible_default_ipv4`](https://github.com/jlund/streisand/blob/77e4208ac2267e899a9d018150ae84e32368d3bd/playbooks/roles/shadowsocks/templates/config.json.j2#L2) is a bug/leftover artifact. Rather than switch the `wait_for` to use the same incorrect ansible var I expect we should change the `config.json` to use `streisand_ipv4_address`. Can you update your PR to do that?

The expectation is that all playbooks use `streisand_ipv4_address` for the external IPv4 address. It is populated using the `ansible_default_ipv4` if no provisioners override it and if that isn't happening correctly for your environment we should fix that instead :-)

yousong:
It's a cloud service provider in china https://www.mtyun.com/.  The KVM machine there can have 2 nics, one for internal tenant network and another for internet access.  The default route will point to the nic for internet nic.

I encountered the issue when trying out streisand through the internal nic from which I think `streisand_ipv4_address` will take value from.  But I searched the net and `ansible_default_ipv4.address` will take address from the interface of the route of command `ip route get 8.8.8.8`.  The 2 addresses did not match and the `wait_for` task failed.

cpu:
Hi @yousong,

Thanks for explaining! I think I understand now.

> I encountered the issue when trying out streisand through the internal nic from which I think streisand_ipv4_address will take value from

I believe that you can control the value the `streisand_ipv4_address` takes when you're using a custom VPS provider with your Ansible inventory file, because Streisand will populate it [based on the `streisand_ipv4_address` host var in the inventory, if present](https://github.com/jlund/streisand/blob/698cecbb774e6f429e64e7a1b3566bee641e4d50/playbooks/roles/common/tasks/set-default-variables.yml#L20:L28).

Can you try setting a `streisand_ipv4_address` host var in the inventory file for your `streisand-host` host and seeing if that accomplishes what you're after?

cpu:
Hi again @yousong - I'm going to merge this PR since it seems ready.

I recommend that if you try the [approach I suggest in my comment](https://github.com/jlund/streisand/pull/707#issuecomment-304038187) and run into the same problem again that you open a [new issue](https://github.com/jlund/streisand/issues/new). We can continue troubleshooting there.

Thanks!

yousong:
Hi @cpu , Thanks for the info.  After checking the code, I am fairly sure it will work as expected though I am new to the ansible world.

cpu:
@yousong Great! Feel free to ask for help if you get stuck :-)

PR #712, opened 2017-05-27
Removes README claim RE: client IP logging.

cpu:
The README has a feature line describing the Streisand server's logging
configuration w.r.t. trying to minimize client IP collection. My
understanding of the original intent was that this was considered from
the perspective of one sysadmin running a Streisand instance for many
clients, whose IPs would be ideally not recorded. This has caused some
some confusion[0] for folks thinking it was about the sysadmin/operator
themselves.

Overall I think this conversation requires a deeper conversation RE:
threat models and we should probably stay out of the weeds by saying
less in the README :-)

[0]: https://arstechnica.com/gadgets/2017/05/how-to-build-your-own-vpn-if-youre-rightfully-wary-of-commercial-options/?comments=1&post=33391773


jlund:
That was indeed the sentiment, but I agree that discussions about threat models need to be nuanced and that this line could lead to confusion. +1 from me.

PR #713, opened 2017-05-27
Revert "shadowsocks: listen on streisand_ipv4_address (#707)"

cpu:
This reverts commit 282d5eac6b12d33c89a434d0042c2e2b646b61d8
from https://github.com/jlund/streisand/pull/707.

It causes issues on AWS from the I think the primary interface address
having an internal 10.x VPC IP that doesn't match the external IP.


cpu:
Created https://github.com/jlund/streisand/issues/714 as a follow-up since this _should_ have worked.

Issue #468, opened 2017-01-06
Feature Request - Local Provisioning


wicknet:
Hi Joshua,

Your script continues to rock in the new year, working flawlessly to spin up new AIO anti-censorship servers for friends and family. 

I was wondering if support for local provisioning is still on the development road-map? The only “weak-ish” link in the project (that I’ve encountered with users) is their local system (which in many cases is a Mac clogged with other, conflicting apps/OS). It needs to be perfectly configured to run the installer packages properly, else the installation of the remote server fails with errors. 

However, if local provisioning could be supported (perhaps as an initial choice - local or remote install), I’m confident that your wonderful script would become almost “universally usable”, as many folks over here in the US are not trying to circumvent the Great Fire Wall but annoying workplace surfing restrictions via a self-hosted Streisand server at home, instead of one with a hosting provider, that also comes with a recurring monthly cost. 

Local provisioning would be of great help, especially to rookies who can't figure out how to edit the inventory file and direct the install to another, local server.

FWIW!

cpu:
Local provisioning has been added as of #724 being merged. Thanks!

Issue #122, opened 2015-01-31
How do i deploy streisand on local server, instead of deploying it on the cloud?


fnlctrl:
as the title says

zjx20:
+1

skeeve42:
This isn't currently possible, though @jlund has mentioned including this functionality at some point in #93 in a Dec. 18 post.

skeeve42:
Any update on this? It would really make things simpler for me to help others get streisand going who only have access to windows machines and the server they're putting it on.

soyuka:
[uncomment this](https://github.com/jlund/streisand/blob/master/inventory#L16 ) and set it to your server ip address, I remember that I used `localhost` once but docs had wrong adresses.

Add this:
```
host_key_checking = False // line in the ansible.cfg file. That setting
```
to the ansible.cfg file.


Launch the ansible playbook directly:
```
ansible-playbook playbooks/streisand.yml
```

skeeve42:
@soyuka Thanks I'll give it a try!

skeeve42:
@soyuka I finally got around to trying this, but it gave me an immediate error. (data could not be sent to the remote host). Ah well, I'll just install it via another VPS again. Silly, but all I can do!

kydan:
@skeeve42 I too have been able to get this to work on a local machine, and it did require the modifications that @soyuka suggested. I'll try to dig up which line items I had to monkey with to make it work, but I got it runing within a few hours.

aeibrahim:
you could try adding localhost to inventory and run ansible-playbook with --connection=local option

ma-tt:
+1

cpu:
Hi @ma-tt, please avoid +1 comments since they create notifications unneccesarily. Can you [use a Github reaction emoji](https://github.com/blog/2119-add-reactions-to-pull-requests-issues-and-comments) :+1: on the issue itself in the future? Thanks!

cpu:
This has been officially supported since https://github.com/jlund/streisand/pull/724 was merged. The answer to the original thread topic is now:
1) Clone master on the server
2) Follow the setup instructions (installing python, pip, ansible)
3) Running `./streisand`
4) Choosing the local server install.

Closing this issue. Thanks everyone.

PR #724, opened 2017-06-03
Add support for local provisioning.

### LINKED ISSUE #595: Local install option
Know that there is an effort to create a local (non-ansbile, presumably) install option... Is there any sort of ETA on its completion?

Trying to get Streisand installed via a OSX box has proven very frustrating, primarily with broken distribution packages for pip and pycurl.

Trying to avoid using one ec2 instance to just launch another one. :)

cpu:
This PR adds initial first-party support for "local provisioning",
that is, installing Streisand to the same machine that is running the
Ansible playbooks. This feature can be useful for individuals that can't
meet the Ansible system requirements on their client machine (e.g.
because it is Windows), or for users whose internet connection makes the
Ansible SSH connections unreliable.

The Streisand script is updated to pull distinct operations into bash
functions. A new test for the existence of the `~/.ssh/id_rsa.pub` SSH key
is added, since presently Streisand assumes it to exists and this is 
frequently a source of user error. 

To support local provisioning a new option is added to the initial
./streisand script menu. It is marked as an advanced feature and
requires that you enter a confirmation prompt to ensure you fully
understand that the Streisand playbooks will be overwriting configs on
the same machine that is running the script.

Similarly, an advanced option for running the Streisand playbook against
an existing server was added to the wrapper script. This allows testing the
SSH access for the root user (overidable with the `ANSIBLE_SSH_USER`
env var) before invoking the playbook, hopefully again helping to identify
user error early. The existing server option generates an inventory file on
the fly based on an IP the user enters and writes it to 
`inventories/inventory-existing`, avoiding the need to manually edit an
inventory file for existing servers & hopefully encouraging more people to
use the wrapper where we can detect common errors.

With several inventory files to `sed` patch for `python` vs `python2` and 
to avoid clutter this PR also moves the inventory files into a dedicated 
`inventories/` directory. The `ansible.cfg` file is updated to use 
`inventories/inventory` as the default inventory to maintain backwards
compatibility.

To help keep a growing pile of bash manageable `shellcheck` is added
to CI builds and run against the `streisand` wrapper.

For localhost installs the `streisand_ip4_address` var is set to
the Ansible default IPv4 address, which itself is calculated based on
the default IPv4 route to 8.8.8.8. This works for _most_ systems but
will likely fail for some more esoteric providers.

For the more complicated cases (e.g. GCE) a common 
`playbooks/provider-detect.yml` attempts to determine if the instance 
belongs to a known cloud provider (currently only GCE & AWS are 
supported) and will update the external IPv4 address according to the instance
metadata since it is not available as a property of any of the interfaces
on the server.

I've tested so far with:
* [x] Linode
* [x] Digital Ocean
* [x] GCE
* [x] AWS
* [ ] Others?

Additional testing welcome!

Resolves #595, #468, #122


cpu:
Before merging I'd like to try and revisit the "local provisioning" terminology. One point of feedback received so far is that "local" is a highly overloaded word and in practice confuses folks. Alternate suggestions: "headless mode", "direct mode", "self-contained mode". Something to think about!

pjrobertson:
Great to see this a WIP. I was just looking to do this a few minutes ago! One problem I've had in the past doing this is that SSL certificates etc. use the IP address `127.0.0.1` as opposed to the server's outside IP address.

Also, regarding the word 'local' - I think that is actually more meaningful than any of the other suggestions. Basically what this is doing is not putting out any calls across the network, so any chosen name should make this clear. "local install", "no network", "same machine". I don't think "direct mode" or "self-contained mode" are really very meaningful.

cpu:
>  One problem I've had in the past doing this is that SSL certificates etc. use the IP address 127.0.0.1 

I tried to handle this case by [the localhost.yml role setting the streisand_ipv4_address](https://github.com/cpu/streisand/blob/ef4a5688b72d1ca6523a8544219a5703cf4eb871/playbooks/localhost.yml#L8:L20). It should work OK provided Ansible detects it correctly.

cpu:
I was able to successfully provision an existing GCE instance both from "localhost mode" (running on-instance) and "existing server mode" (running from a client to the existing server). Since GCE doesn't have the external IP address assigned to any of the interfaces on the machine I introduced some workarounds in 3951598. This includes Yet Another Warning that if you're using one of these "advanced" modes against an existing GCE instance that you're responsible for making the network work (e.g. creating a Google cloud network, opening the correct ports, etc).

playbooks/provider-detect.yml
```diff
@@ -0,0 +1,46 @@
...
+        streisand_bios_name: "{{ streisand_localhost_bios_name.stdout }}"
+
+    # GCE specific work-arounds:
+    #   * None of the interfaces have the external IP bound, so this must be set
+    #     from the Google Platform medatadata service.
+    - block:
+        - name: Warn about manual provisioning of GCE instances
+          pause:
+            prompt: "You are running Streisand in an advanced mode against an existing GCE instance. Unlike the standard GCE provisioning mode this means Streisand *CAN NOT* open ports on your behalf. It is assumed you have manually opened all required ports. The Streisand maintainers aren't able to support you if you are unable to do this. Press [enter] to continue"
+
```
alimakki: Is there a file/url that could be referred to for a list of ports needed to be opened up? I think it would help the user when doing a local install, instead of hunting down numbers in separate places.

Not a blocker, but nice to have.

cpu: Hmmm. There is the firewall documentation that is generated after a full run. It would take a fair amount of work but perhaps that could be generated ahead-of-time and used as a reference.

What I did to test GCE was to assign the new VM to the network created from an older "standard" GCE provision.

I'm tempted to punt on this because I don't _really_ want folks using GCE for these advanced provisioning modes, they should probably use the standard provisioning that does all of this work for them,

cpu: I added a note referencing the `generated-docs/$HOSTNAME-firewall-information.html` file that is created at the end of a successful run. I think that's the best option for now.

playbooks/provider-detect.yml
```diff
@@ -0,0 +1,64 @@
...
+  remote_user: "root"
+  become: true
+
+  tasks:
+    - name: "Try to determine localhost Cloud provider name from BIOS version"
+      # Looking into ways to detect EC2/GCE instances without relying on
+      # connections to metadata IPs this solution[0] seemed the most
+      # straightforward & reliable.
+      #   https://serverfault.com/a/775063
+      command: dmidecode -s bios-version
```
jlund: Clever!

streisand
```diff
@@ -1,47 +1,168 @@
 #!/usr/bin/env bash
 # Streisand provisioning script
 
+# Set errexit option to exit immediately on any non-zero status return
```
jlund: Commenting the shell script :clap:

playbooks/existing-server.yml
```diff
@@ -0,0 +1,15 @@
+---
+# existing-server.yml is an advanced provisioning option that doesn't use a genesis
+# role to create a new server and instead applies Streisand to an existing
+# remote server.
+#
+# Try and detect the remote server's provider & apply required workarounds
+- include: provider-detect.yml
```
nickolasclarke: I vote for including `python.yml` up higher. We've got a perfectly working hammer, we use the pefectly working hammer elsewhere, just use the hammer. You'll need py eventually anyhow.

cpu: Good catch! I agree with both you and Nick, I can include `python.yml` in `localhost.yml` & `existing-server.yml` as the first statement before `provider-detect.yml` is included.

cpu: Resolved w/ cebb0dd

PR #727, opened 2017-06-05
Update mirrored Shadowsocks Android APK to 4.1.8.

cpu:
The `shadowsocks-android` project has tagged a `4.1.8` release today
that reintroduces the built-in QR code scanner. Some Streisand users
have found this feature helpful in the event they don't have the ZXing
QR code scanner and can't access Google Play to get it.

This commit updates the Streisand mirror's copy of the Shadowsocks
Android client to version 4.1.8.

Resolves https://github.com/jlund/streisand/issues/349


cpu:
Tested with an Android client using the mirrored APK connecting to a fresh GCE instance. Works as expected.

PR #733, opened 2017-06-08
Remove shadowsocks port from FR SSH docs.

cpu:
It looks like the `shadowsocks_local_port` snuck into the French SSH
instructions template where it should be `ssh_default_socks_port`. This
commit fixes that error.


alimakki:
Thanks @cpu!

cpu:
> :+1: All those .mds start to look alike after a while. Thanks for catching that.

N/P! I definitely missed it in review too :-) Thanks for the quick review.

Issue #741, opened 2017-06-16
Support `simple-obfs` for Streisand shadowsocks-libev service


cpu:
[One observation](https://github.com/jlund/streisand/issues/501#issuecomment-308970827) from the longer discussion in https://github.com/jlund/streisand/issues/501 is that using `simple-obfs` with the existing `shadowsocks-libev` service in Streisand can improve conditions for users relying on SS from a censored network:

> With shadowsocks without simple-obfs, the latency increases to 290-300 ms while I browse the internet.
> 
> With shadowsocksR having obfs configured, the latency increases to 250-270 ms.
> 
> However, with SS simple obfs, I am writing this comment now and browsing for almost 30 minutes while my latency is back to 204-208 ms.

It would be interesting to evaluate how we could integrate `simple-obfs` support into the existing `shadowsocks-libev` service.

denoza:
1. In my opinion, the simplest way is to install `simple-obfs` plugin while the `shadowsocks-libev` installation task is carried out.

2. The `simple-obfs` plugin can be enabled by default on the server side (in the server config JSON). As even if the client doesn't wish to have obfs feature, he/she has a choice to enable/disable it on his/her side. In my experience, it doesn't matter if I don't enable it on the client side, I am still able to use SS without it, however on the server side I have enabled it by default.

3. Addition of a couple of lines in the documentation section of shadowsocks, mentioning to add couple of flags on client config JSON and restarting the shadowsocks service.

ortonomy:
I'd be happy to try this out and let you know how it goes, but I can't find any client-side settings in Shadowsocks-NG for Mac.

On iOS I'm using Shadowrocket which has the option for ``http`` or ``tls``.

As I have a pretty stable Streisand server already spun up on Digital Ocean using the latest version with AEAD, I'm in a pretty good place to A/B/C/... with different configs.

denoza:
> I'd be happy to try this out and let you know how it goes, but I can't find any client-side settings in Shadowsocks-NG for Mac.

@ortonomy Any success with Shadowsocks-NG for Mac? Since 'simple-obfs' is now part of master, it would be good to know about your experience while using it on Mac and iOS.

ortonomy:
I'm sorry, I didn't see that you've added a PR / merge. I'll roll a new server and let you know.

cpu:
[...]

PR #743, opened 2017-06-18
Add shadowsocks 'simple-obfs' plugin installation task and update server config json

denoza:
This PR addresses the solutions discussed in #741. The changes committed include:

1.  The unused variable `shadowsocks_checksum` has been removed. Since, `shadowsocks-libev` is compiled directly via git repository therefore it is not needed anymore.

2. The `simple-obfs` installation task has been added under the role of `shadowsocks-libev`. Git repo is used to compile the `simple-obfs` plugin and its installation. 

3. `config.json` on the server side has been edited to include the changes required to turn the plugin ON. By turning it ON, the server enables the use of simple-obfs, however user has a choice to turn it off/on at his end. (Tested on `shadowsocks-libev` on client side, testing with further clients and platforms is needed)

4. The english instructions have been edited to include the information for Linux based `shadowsocks` clients, which can use `simple-obfs` plugin as an added layer of security in the case of QOS throttling.

Future tasks include:

- Testing on clients based upon platforms other than Linux. (Help will highly be appreciated for Mac and iOS based clients)

- Addition of instructions for other clients in English and French.


cpu:
@denoza Thanks for contributing this PR! I just wanted to leave a note that I will review this in the next couple of days.

playbooks/roles/shadowsocks/defaults/main.yml
```diff
@@ -4,3 +4,6 @@ shadowsocks_local_port: "1080"
 shadowsocks_timeout: "600"
 shadowsocks_encryption_method: "chacha20-ietf-poly1305"
 shadowsocks_tcp_fast_open: "true"
+shadowsocks_obfs_plugin: "obfs-server"
+shadowsocks_obfs_plugin_opts: "obfs=http"
+shadowsocks_obfs_client_plugin_opts: "obfs=http;obfs-host=www.xxx.com"
```
cpu: What is the significance of the `obfs-host=www.xxx.com` in the `shadowsocks_obfs_client_plugin_opts`?

playbooks/roles/shadowsocks/tasks/main.yml
```diff
@@ -100,6 +100,48 @@
     flat: yes
   when: streisand_client_test
 
+#simple-obfs begins here
```
cpu: Small nit: can you put a space between the `#` and the comment text? Thanks!

playbooks/roles/shadowsocks/tasks/main.yml
```diff
@@ -100,6 +100,48 @@
...
+    chdir: "{{ simple_obfs_compilation_directory }}"
+    creates: "{{ simple_obfs_compilation_directory }}/src/obfs-server"
+
+- name: Install simple-obfs {{ simple_obfs_version }} binaries
+  command: make install
+  args:
+    chdir: "{{ simple_obfs_compilation_directory }}"
+    creates: "/usr/local/bin/obfs-server"
+
+#simple-obfs ends here
```
cpu: Ditto on the comment nit from above.

playbooks/roles/shadowsocks/templates/instructions.md.j2
```diff
@@ -64,6 +64,17 @@ For reference, this is the configuration specified:
         Password: {{ shadowsocks_password.stdout }}
         Encryption Method: {{ shadowsocks_encryption_method }}
 
+For users behind stricter surveillance (esp. experiencing QOS throttling in PRC), further configuration of shadowsocks along with simple-obfs plugin can be carried out via following configuration:
```
cpu: I think the "behind stricter surveillance" part would be better expressed as "For users on unreliable or hostile networks (esp. experiencing quality-of-service (QOS) throttling) ..."

"further configuration of shadowsocks along with simple-obfs" -> "further configuration of Shadowsocks to use the `simple-obfs` plugin ..."

playbooks/roles/shadowsocks/templates/instructions.md.j2
```diff
@@ -64,6 +64,17 @@ For reference, this is the configuration specified:
...
+
+        Server: {{ streisand_ipv4_address }}
+        Port: {{ shadowsocks_server_port }}
+        Password: {{ shadowsocks_password.stdout }}
+        Encryption Method: {{ shadowsocks_encryption_method }}
+        Plugin: {{ shadowsocks_obfs_plugin }}
+        Plugin_Options: {{ shadowsocks_obfs_client_plugin_opts }}
+
+Note: www.xxx.com should be replaced by a known host name like, www.cloudfront.com or www.baidu.com etc.
```
cpu: See my earlier question about this config setting. Should it be replaced server-side? If so then I think this will be problematic. Ideally we can come up with a solution that can be used without manual configuration file editing on the Streisand instance.

playbooks/roles/shadowsocks/vars/main.yml
```diff
@@ -17,8 +17,8 @@ shadowsocks_dependencies:
   - libudns-dev
   # Shadowsocks may complain about lack of entropy without rng-tools installed
   - rng-tools
+  - libssl-dev
 
-shadowsocks_checksum: "sha256:7d9b43b0235a57c115bfe160efd54abef96bffcbfff61c5496e7c2800f0734ca"
```
cpu: Good catch on this stale var! Thanks!

playbooks/roles/shadowsocks/vars/main.yml
```diff
@@ -32,6 +32,11 @@ shadowsocks_location: "/etc/shadowsocks-libev"
 shadowsocks_password_file: "{{ shadowsocks_location }}/shadowsocks-password.txt"
 shadowsocks_password_file_localpath: "../{{ streisand_local_directory }}/shadowsocks-password.txt"
 
+#simple-obfs vars
```
cpu: Same comment about `#` and a space.

playbooks/roles/shadowsocks/vars/main.yml
```diff
@@ -32,6 +32,11 @@ shadowsocks_location: "/etc/shadowsocks-libev"
...
+simple_obfs_version: "0.0.2"
```
cpu: There is actually a newer "0.0.3" release. Can you update this PR to use that release instead?

I opened [a PR with upstream](https://github.com/shadowsocks/simple-obfs/pull/75) to fix the README, which still claims that 0.0.2 is the latest release.

playbooks/roles/shadowsocks/defaults/main.yml
denoza: @cpu Setting this variable on client side, obfuscates its shadowsocks traffic as the host name mentioned here. I used `x` here as in mathematics for a variable hostname, however if it is confusing then I can change it to a more suitable value. I have added this variable `shadowsocks_obfs_client_plugin_opts` to be used in the instructions for Linux based clients esp. SS-libev clients. I haven't checked it with the mirrored SS client for linux by streisand, however  I am on it.

playbooks/roles/shadowsocks/tasks/main.yml
denoza: NP

playbooks/roles/shadowsocks/templates/instructions.md.j2
denoza: Yes it looks better. Thanks!

playbooks/roles/shadowsocks/templates/instructions.md.j2
denoza: No it isn't replaced server side. It is only mentioned for the client side. Even without setting it, shadowsocks traffic works fine in my tests (client side). However, I have access to Linux based SS-libev client, I checked windows version but it doesn't support this option as yet. That is why I only edited the Linux section. Android client also has an option of disabling/enabling `simple-obfs` as per requirements.

So this setting is **optional** for client side only. Server side `simple-obfs` can be enabled by default and it doesn't affect the non-obfs traffic from clients.

playbooks/roles/shadowsocks/vars/main.yml
denoza: Thanks for this. I followed upstream and seems like it wasn't updated either.

playbooks/roles/shadowsocks/defaults/main.yml
denoza: I'll change it to `www.cloudfront.net`. Didn't know that there is a website with this host name which is highly NSFW.

atulsian89:
@denoza, thank you for taking this. I was wondering if this PR supports both tls and http? I see in android client we have option to select http or tls wrapper. I believe Tls will be more effective towards QOS.

playbooks/roles/shadowsocks/defaults/main.yml
cpu: Aha! Thanks for explaining. Can you use `www.github.com` instead of `www.cloudfront.com`? We rely on this external URL elsewhere and since the user cloned Streisand from Github it's ~relatively unlikely to be blocked.

playbooks/roles/shadowsocks/templates/instructions.md.j2
cpu: Ahhhh! That makes sense and I follow now. Thanks again for explaining. I thought the var was being put into the ss-server config when it's not. Disregard my comments :-)

playbooks/roles/shadowsocks/vars/main.yml
cpu: All fixed upstream!

playbooks/roles/shadowsocks/tasks/main.yml
```diff
@@ -100,6 +100,48 @@
     flat: yes
   when: streisand_client_test
 
+# simple-obfs begins here
```
cpu: Apologies for not flagging this in my first pass, but instead of adding these `# begins here` and `# ends here` comments, how about creating a separate tasks file under `playbooks/roles/shadowsocks/tasks/` called `simple-obfs.yml` - then in `playbooks/roles/shadowsocks/tasks/main.yml` you can include the `simple-obfs.yml` tasks without needing any "begin"/"ends" comments. The simple-obfs stuff will be self-contained in its own tasks file.

Does that sound OK?

playbooks/roles/shadowsocks/templates/instructions.md.j2
```diff
@@ -64,6 +64,17 @@ For reference, this is the configuration specified:
         Password: {{ shadowsocks_password.stdout }}
         Encryption Method: {{ shadowsocks_encryption_method }}
 
+For users on unreliable or hostile networks (esp. experiencing quality-of-service (QOS) throttling), further configuration of Shadowsocks to use the simple-obfs plugin can be carried out via following configuration:
```
cpu: small typo: "via following configuration" -> "via the following configuration"

playbooks/roles/shadowsocks/defaults/main.yml
```diff
@@ -4,3 +4,7 @@ shadowsocks_local_port: "1080"
 shadowsocks_timeout: "600"
 shadowsocks_encryption_method: "chacha20-ietf-poly1305"
 shadowsocks_tcp_fast_open: "true"
+shadowsocks_obfs_plugin: "obfs-server"
+shadowsocks_client_obfs_plugin: "obfs-local"
+shadowsocks_obfs_plugin_opts: "obfs=http"
+shadowsocks_obfs_client_plugin_opts: "obfs=http;obfs-host=www.cloudfront.com"
```
cpu: When you replace `www.cloudfront.com` can you put it into its own var? e.g. `shadowsocks_obfs_cover_domain: "www.github.com"` and then set `shadowsocks_obfs_client_plugin_opts: "obfs=http;obfs-host={{ shadowsocks_obfs_cover_domain }}". That lets us use the var in the documentation instead of copy/pasting.

playbooks/roles/shadowsocks/templates/instructions.md.j2
```diff
@@ -64,6 +64,17 @@ For reference, this is the configuration specified:
...
+
+        Server: {{ streisand_ipv4_address }}
+        Port: {{ shadowsocks_server_port }}
+        Password: {{ shadowsocks_password.stdout }}
+        Encryption Method: {{ shadowsocks_encryption_method }}
+        Plugin: {{ shadowsocks_client_obfs_plugin }}
+        Plugin_Options: {{ shadowsocks_obfs_client_plugin_opts }}
+
+Your Shadowsocks traffic will now be obfuscated as HTTP traffic from `www.cloudfront.com`.
```
cpu: Can you replace "`www.cloudfront.com`" with "`{{ shadowsocks_obfs_cover_domain }}`" (see my earlier comment on `playbooks/roles/shadowsocks/defaults/main.yml`)

denoza:
@atulsian89 This PR supports `http` for now. However, `tls` flag can be set if `http` doesn't serve the purpose well. Right now, it doesn't support both of them. My tests with `http` are successfull though. However, I saw this issue https://github.com/shadowsocks/simple-obfs/issues/73 on the upstream repository. I haven't tested it yet but will definitely see if I am able to reproduce it.

playbooks/roles/shadowsocks/tasks/main.yml
denoza: That sounds good, will clean up the main task!

playbooks/roles/shadowsocks/templates/instructions.md.j2
denoza: yes that would be simple

cpu:
@denoza, @atulsian89 - My preference would be to land the `http` version in master and then consider the `tls` improvement in a subsequent follow-up.

@denoza I apologize but a commit from one of my PRs that landed yesterday has caused a conflict on this branch. Can you pull in master and resolve the conflict in `playbooks/roles/shadowsocks/tasks/main.yml`?

denoza:
@cpu I have resolved the conflict with upstream and added the new commits.

playbooks/roles/shadowsocks/templates/instructions.md.j2
```diff
@@ -64,6 +64,17 @@ For reference, this is the configuration specified:
         Password: {{ shadowsocks_password.stdout }}
         Encryption Method: {{ shadowsocks_encryption_method }}
 
+For users on unreliable or hostile networks (esp. experiencing quality-of-service (QOS) throttling), further configuration of Shadowsocks to use the simple-obfs plugin can be carried out via the following configuration:
```
cpu: What do you think about adding this text at the very bottom of the file under its own heading instead of under the `Linux` heading. Something like:

```
 <a name="simple-obfs"></a>
 ### simple-obfs for unreliable/hostile networks ###
```

I think this information might be useful to folks other than Linux users.

playbooks/roles/shadowsocks/templates/instructions.md.j2
```diff
@@ -64,6 +64,17 @@ For reference, this is the configuration specified:
...
+
+        Server: {{ streisand_ipv4_address }}
+        Port: {{ shadowsocks_server_port }}
+        Password: {{ shadowsocks_password.stdout }}
+        Encryption Method: {{ shadowsocks_encryption_method }}
+        Plugin: {{ shadowsocks_client_obfs_plugin }}
+        Plugin_Options: {{ shadowsocks_obfs_client_plugin_opts }}
+
+Your Shadowsocks traffic will now be obfuscated as HTTP traffic from `{{ shadowsocks_obfs_cover_domain }}`.
```
cpu: I think this should actually read "will now be obfuscated as HTTP traffic **to** `{{ shadowsocks_obfs_cover_domain }}`" no?

cpu:
As a follow up to my [other testing comment]() I also tried with the Android Shadowsocks app and the Android simple-obfs plugin. The results were the same. Without the plugin enabled/configured the connection worked fine. With the plugin enabled/configured the connection did not work.

denoza:
@cpu I am wondering how your `ss-local` works the first time when you don't use `obfs-local` plugin. Because you mention your local port as `-l9999` while on server it is set as `1080`. I think there is an issue here.

On the other hand, I tested my client on an existing server and a newly spun server. Both of them worked fine for me. I am mentioning my results as under:

1. Server config is same as yours.

2. The presence of running `ss-server` process with `obfs-plugin` is confirmed by:
```~# systemctl status shadowsocks-libev.service
● shadowsocks-libev.service - Shadowsocks-libev Default Server Service
   Loaded: loaded (/lib/systemd/system/shadowsocks-libev.service; enabled; vendor preset: enabled)
   Active: active (running) since Sat 2017-06-24 03:49:56 UTC; 9min ago
     Docs: man:shadowsocks-libev(8)
 Main PID: 31224 (ss-server)
    Tasks: 2
   Memory: 1.3M
      CPU: 41ms
   CGroup: /system.slice/shadowsocks-libev.service
           ├─31224 /usr/bin/ss-server -c /etc/shadowsocks-libev/config.json -u
           └─31227 obfs-server
```
3. On client machine, I get following:
```
~ # ss-local --plugin obfs-local --plugin-opts "obfs=http;obfs-host=www.github.com" -sXX.XX.XX.XX -p8530 -l1080 -k"XXX" -mchacha20-ietf-poly1305 -v
 2017-06-24 11:54:26 INFO: plugin "obfs-local" enabled
 2017-06-24 11:54:26 INFO: initializing ciphers... chacha20-ietf-poly1305
 2017-06-24 11:54:26 INFO: listening at 127.0.0.1:1080
 2017-06-24 11:54:26 [simple-obfs] INFO: obfuscating enabled
 2017-06-24 11:54:26 [simple-obfs] INFO: obfuscating hostname: www.github.com
 2017-06-24 11:54:26 [simple-obfs] INFO: tcp port reuse enabled
 2017-06-24 11:54:26 [simple-obfs] INFO: listening at 127.0.0.1:41921
```
4. Server side logs:
```
~# journalctl -u shadowsocks-libev.service 
-- Logs begin at Sat 2017-06-24 03:43:32 UTC, end at Sat 2017-06-24 04:02:30 UTC. --
Jun 24 03:49:56 test systemd[1]: Started Shadowsocks-libev Default Server Service.
Jun 24 03:49:56 test ss-server[31224]:  2017-06-24 03:49:56 INFO: using tcp fast open
Jun 24 03:49:56 test ss-server[31224]:  2017-06-24 03:49:56 INFO: plugin "obfs-server" enabled
Jun 24 03:49:56 test ss-server[31224]:  2017-06-24 03:49:56 INFO: UDP relay enabled
Jun 24 03:49:56 test ss-server[31224]:  2017-06-24 03:49:56 INFO: initializing ciphers... chacha20-ietf-poly1305
Jun 24 03:49:56 test ss-server[31224]:  2017-06-24 03:49:56 INFO: tcp server listening at 127.0.0.1:38036
Jun 24 03:49:56 test ss-server[31224]:  2017-06-24 03:49:56 INFO: udp server listening at XX.XX.XX.XX:8530
Jun 24 03:49:56 test ss-server[31224]:  2017-06-24 03:49:56 [simple-obfs] INFO: obfuscating enabled
Jun 24 03:49:56 test ss-server[31224]:  2017-06-24 03:49:56 [simple-obfs] INFO: tcp port reuse enabled
Jun 24 03:49:56 test ss-server[31224]:  2017-06-24 03:49:56 [simple-obfs] INFO: listening at XX.XX.XX.XX:8530
```
5. Finally Android client should be enabled to use the `simple-obfs` app. While editing the server configuration, scrolling all the way down reveals the `Plugin` option where you select `simple-obfs` as plugin and configure it accordingly.

Hope this helps.

Cheers

playbooks/roles/shadowsocks/templates/instructions.md.j2
denoza: Thanks for correction

playbooks/roles/shadowsocks/templates/instructions.md.j2
denoza: Yes, that will be fine

cpu:
> @cpu I am wondering how your ss-local works the first time when you don't use obfs-local plugin. Because you mention your local port as -l9999 while on server it is set as 1080. I think there is an issue here.

@denoza Good question! I hadn't noticed the mismatch and assumed it wasn't a problem since my understanding of the local port was that it only specified the port that the SOCKS proxy was made available on my localhost. I've never been 100% clear what the "local port" does on the server or if its necessary to match with the client. I will retry with matching ports to remove one variable.

> Finally Android client should be enabled to use the simple-obfs app. While editing the server configuration, scrolling all the way down reveals the Plugin option where you select simple-obfs as plugin and configure it accordingly.

That's how I configured my Android client. Very strange!

I will try to re-test in the next day or so. Thanks for your help/patience!

cpu:
@denoza Update: I was able to get the Android client & simple-obfs plugin to work with a fresh Streisand server provisioned with this branch :tada: - I'm still struggling to get a Ubuntu 16.04 VM to do the same with `shadowsocks-libev` as a client w/ simple-obfs. I tried changing the local port on my `ss-local` machine to `1080` to match the server config but it didn't seem to help.

> On client machine, I get following:

Are you able to `curl -I --socks5 localhost:1080 www.github.com` from the client machine? I have the same output from my client machine up until I actually try to send a request through the socks proxy that `ss-local` creates.

cpu:
@denoza I figured it out! It was a firewall issue on my host machine. Sorry about that, I should have noticed much sooner!

I was able to verify a `shadowsocks-libev` client works correctly with and without the obfuscation configured as described in your addition to the docs :tada:.

I think this is ready to merge. I will work on https://github.com/jlund/streisand/issues/762 over the next few days as a follow-up to make sure we don't regress support in the future.

denoza:
> @denoza I figured it out! It was a firewall issue on my host machine. Sorry about that, I should have noticed much sooner!

@cpu I am glad you were able to fix that.

> This is an awesome contribution @denoza - Thanks so much for sticking with it & integrating all of the review feedback so quickly. It was a pleasure to work with you!

It was a pleasure to work with you as well @cpu. Your guidance meant a lot, esp. since it was my first ever pull request. I am looking forward to contributing more.

PR #744, opened 2017-06-18
Adds support for mirroring localized versions of the Tor Browser Bundle

alimakki:
In order to support mirroring localized versions of Tor, steps taken from the `download-and-verify` task were unrolled and modified into three distinct steps specifically for Tor:

1. `tor-common`&nbsp;&nbsp;&nbsp;&nbsp;- Determining the recommended TBB version and registering the GPG keys
2. `tor-download`&nbsp;- Downloads TBB for the locales defined in `streisand_languages`
3. `tor-verify`&nbsp;&nbsp;&nbsp;&nbsp;- Integrity verification 

Furthermore, the download task is gated only to available locales that TBB supports defined as `tor-available-locales`.

### DIFF
--- a/playbooks/roles/streisand-mirror/tasks/tor-common.yml
+++ b/playbooks/roles/streisand-mirror/tasks/tor-common.yml
@@ -1,0 +1,42 @@
+---
+
+- name: Make the directory where the Tor Project's mirrored files will be stored
+  file:
+    path: "{{ tor_mirror_location }}"
+    owner: www-data
+    group: www-data
+    mode: 0755
+    state: directory
+
+- name: Discover the latest stable version of the Tor Browser Bundle
+  # The Python code below can be cleaned up once the Tor Project has
+  # finished transitioning away from putting the OS suffix in the
+  # RecommendedTBBVersions file. This check should work with both
+  # formats though.
+  #
+  # https://trac.torproject.org/projects/tor/ticket/8940#comment:28
+  shell: curl -s 'https://www.torproject.org/projects/torbrowser/RecommendedTBBVersions' | python -c 'import json; import re; import sys; j = json.load(sys.stdin); print [re.sub(r"-.*$", "", tbb) for tbb in j if "a" not in tbb and "b" not in tbb][-1];'
+  args:
+    warn: no
+  register: tor_latest_recommended_check
+
+- name: Set the target Tor Browser Bundle version
+  set_fact:
+    tor_browser_bundle_version: "{{ tor_latest_recommended_check.stdout }}"
+
+- name: "Import GPG key ID {{ tor_browser_developers_key_id }} for verifying {{ tor_project_name }} downloads"
+  command: gpg --keyserver {{ gpg_key_server_address }} --recv-keys {{ tor_browser_developers_key_id }}
+  register: gpg_recv_keys_result
+  until: gpg_recv_keys_result | success
+  retries: 10
+  delay: 5
+
+- name: Register the GPG Key fingerprint output
+  command: gpg --fingerprint {{ tor_browser_developers_key_id }}
+  register: retrieved_fingerprint
+
+- name: Make sure the retrieved fingerprint perfectly matches the expected fingerprint
+  assert: { that: "tor_browser_developers_expected_fingerprint in retrieved_fingerprint.stdout" }
+
+- name: Include the mirror variables for Tor Browser Bundle
+  include_vars: tor-mirror.yml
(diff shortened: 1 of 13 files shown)


playbooks/roles/streisand-mirror/tasks/tor-common.yml
```diff
@@ -0,0 +1,42 @@
...
+    state: directory
+
+- name: Discover the latest stable version of the Tor Browser Bundle
+  # The Python code below can be cleaned up once the Tor Project has
+  # finished transitioning away from putting the OS suffix in the
+  # RecommendedTBBVersions file. This check should work with both
+  # formats though.
+  #
+  # https://trac.torproject.org/projects/tor/ticket/8940#comment:28
+  shell: curl -s 'https://www.torproject.org/projects/torbrowser/RecommendedTBBVersions' | python -c 'import json; import re; import sys; j = json.load(sys.stdin); print [re.sub(r"-.*$", "", tbb) for tbb in j if "a" not in tbb and "b" not in tbb][-1];'
```
cpu: What do you think about refactoring this into a script file copied to & invoked from the host with a `cmd` task? I think it will improve readability and let you do error handling in a more flexible way than as this one-liner within the task.

playbooks/roles/streisand-mirror/tasks/tor-common.yml
```diff
@@ -0,0 +1,42 @@
...
+  args:
+    warn: no
+  register: tor_latest_recommended_check
+
+- name: Set the target Tor Browser Bundle version
+  set_fact:
+    tor_browser_bundle_version: "{{ tor_latest_recommended_check.stdout }}"
+
+- name: "Tor Browser Developers signs the {{ tor_project_name }} downloads. Import the correct GPG key."
```
cpu: Personally I find the "XXX signed the YYYY" wording/output strange though I know it's an established paradigm in Streisand. Would you be open to changing this `name` line to something like:
`"Import the {{ tor_browser_developers_key_id }} GPG key for verifying {{ tor_project _name }} downloads"`?

playbooks/roles/streisand-mirror/tasks/tor-common.yml
```diff
@@ -0,0 +1,42 @@
...
+  command: gpg --keyserver x-hkp://pool.sks-keyservers.net --recv-keys {{ tor_browser_developers_key_id }}
```
cpu: Is this the same `--keyserver` we use elsewhere? Could you put it in a var?

playbooks/roles/streisand-mirror/tasks/tor.yml
```diff
@@ -1,40 +1,17 @@
...
-      args:
-        warn: no
-      register: tor_latest_recommended_check
+    - include: tor-common.yml
 
-    - name: Set the target Tor Browser Bundle version
-      set_fact:
-        tor_browser_bundle_version: "{{ tor_latest_recommended_check.stdout }}"
+    - include: tor-download.yml
+      with_items: "{{ streisand_languages.keys() | map('regex_replace', 'en', 'en-US') | list }}"
```
cpu: Would it be worthwhile renaming the Streisand "en" to "en-US"? IIRC we don't rely on it being "en" and that would be one less replace to run.

playbooks/roles/streisand-mirror/tasks/tor-common.yml
alimakki: Done!

playbooks/roles/streisand-mirror/tasks/tor-common.yml
alimakki: Done!

playbooks/roles/streisand-mirror/tasks/tor.yml
alimakki: I took a slightly different approach, instead creating a new value in the `streisand-languages` as `tor-locale`. Any thoughts?

playbooks/roles/streisand-mirror/tasks/tor-common.yml
alimakki: I agree that it should be refactored; I'll be working on another PR to perform a check against Tor for the available TBB locales, and refactor version check along with that.

cpu: Awesome! Thanks!

playbooks/roles/streisand-mirror/tasks/tor.yml
cpu: that works for me

tests/group_vars/all/all
```diff
@@ -27,3 +27,5 @@ libreswan_compilation_dependencies_trusty:
   - libunbound-dev
   - make
   - pkg-config
+
+gpg_key_server_address: "x-hkp://pool.sks-keyservers.net"
```
cpu: I opened https://github.com/jlund/streisand/issues/750 to remember that we should figure out a way to not have to do this duplication going forward.

PR #745, opened 2017-06-18
Verify LibreSwan source code with GPG signatures

### LINKED ISSUE #642: Validate Libreswan source code with GPG keys
Libreswan provides .asc signature files for integrity verification; Streisand should use those instead of the file SHA256 checksum.

alimakki:
Since LibreSwan provides signature files for integrity verification, we should elect to use them whenever possible.

To accomplish this, the PR refactors the `download-and-verify` task previously part of the `streisand-mirror` into its own separate role, making it generic and re-usable; an example invocation being:

```
- name: Retrieve foo
      include_role:
        name: download-and-verify
      vars:
        project_name: "Project Foo"
        project_signer: "John Hancock"
        project_signing_key: "{{ expected_signing_key_id }}"
        project_expected_fingerprint: "{{ expected_fingerprint }}"
        project_download_location: "{{ download_path }}"
        project_download_urls: "{{ list_of_url_strings }}"
``` 

The main `download-and-verify` task remains largely unchanged, the exception being that the variable `project-mirror-location` has been renamed to `project-download-location` to reflect its generalized nature.

Resolves #642


playbooks/roles/l2tp-ipsec/tasks/main.yml
```diff
@@ -14,16 +14,15 @@
     install_recommends: no
 
 - name: Retrieve the Libreswan source code
-  get_url:
-    # This CDN mirrors the Libreswan source code, and helps
-    # mitigate connection errors that were occurring when
-    # using the project's official download location.
-    url: https://d25kfp60e9u1dw.cloudfront.net/{{ libreswan_source_filename }}
```
cpu: Can you preserve this comment about the CDN mirrors? It seems like it could be moved to `playbooks/roles/l2tp-ipsec/vars/main.yml` above the `libreswan_download_urls`.

alimakki: and done!

PR #749, opened 2017-06-20
Remove lxd and snapd from default installation

### LINKED ISSUE #703: Purge snapd and lxd from server
It would be nice to strip down on running services that launched by default. Ubuntu 16.04 comes with `snapd` and `lxd` installed and launched on boot. It's safe to remove them to lower the moving bits on server.

z0rc:
These services are part of default Ubuntu installation. Since Streisand
doesn't use any of those, it's safe to remove them completely.

Closes #703

### DIFF
--- a/playbooks/roles/common/tasks/main.yml
+++ b/playbooks/roles/common/tasks/main.yml
@@ -17,6 +17,14 @@
     name: "{{ item }}"
   with_items: "{{ streisand_common_packages }}"
 
+- name: Purge unneeded services
+  apt:
+    name: "{{ item }}"
+    state: "absent"
+    purge: yes
+    autoremove: yes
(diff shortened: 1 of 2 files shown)


playbooks/roles/common/tasks/main.yml
```diff
@@ -17,6 +17,14 @@
     name: "{{ item }}"
   with_items: "{{ streisand_common_packages }}"
 
+- name: Purge unneed services
+  apt:
```
alimakki: unneed -> unneeded

z0rc: Fixed.

alimakki:
Thank you @z0rc !

PR #779, opened 2017-07-03
Fix tor daemon & obfs4proxy apparmor profile.

### LINKED ISSUE #778: Tor fails to start, apparmor policy omission
Started from scratch today with a fresh git pull and Amazon as the provider. Everything seems to be going fine until tor-bridge:

```
TASK [tor-bridge : Copy the fixed AppArmor abstraction into place] *************
changed: [35.182.166.2]

TASK [tor-bridge : Restart AppArmor so the abstraction takes effect] ***********
changed: [35.182.166.2]

TASK [tor-bridge : Restart Tor so the server fingerprint will be available in the state file, and the hidden service for the Gateway will start running] ***
changed: [35.182.166.2]

TASK [tor-bridge : Wait until obfs4proxy information has shown up in its state file] ***
fatal: [35.182.166.2]: FAILED! => {"changed": false, "elapsed": 300, "failed": true, "msg": "Timeout when waiting for search string node-id in /var/lib/tor/pt_state/obfs4_state.json"}

RUNNING HANDLER [rsyslog : Restart rsyslog] ************************************

RUNNING HANDLER [openconnect : Restart ocserv] *********************************

RUNNING HANDLER [l2tp-ipsec : Restart Libreswan] *******************************

RUNNING HANDLER [l2tp-ipsec : Restart xl2tpd] **********************************

RUNNING HANDLER [dnsmasq : Restart dnsmasq] ************************************

RUNNING HANDLER [openvpn : Restart OpenVPN] ************************************

RUNNING HANDLER [stunnel : Restart stunnel] ************************************

RUNNING HANDLER [ssh : Restart SSH] ********************************************

RUNNING HANDLER [tinyproxy : Restart Tinyproxy] ********************************
    to retry, use: --limit @/home/serge/repos/streisand/playbooks/amazon.retry

35.182.166.2               : ok=183  changed=172  unreachable=0    failed=1  
localhost                  : ok=17   changed=8    unreachable=0    failed=0   
```

Thus never getting to the generated-docs folder. Retried 3 times so far.

This did work successfully on this same host in April. Ubuntu 16.04 with all latest updates installed.

```
ansible 2.3.1.0
  config file = /home/serge/repos/streisand/ansible.cfg
  configured module search path = Default w/o overrides
  python version = 2.7.12 (default, Nov 19 2016, 06:48:10) [GCC 5.4.0 20160609]
```

cpu:
There is [an outstanding bug](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=862993) with the Tor apparmor profile that causes the tor daemon to fail to start when configured with a hidden service directory.

To temporarily work-around this issue this commit adds a new `/etc/apparmor.d/local/system_tor` override file that adds the missing capability. This can be removed when upstream has fixed the problem.

Additionally, while debugging this issue I removed the abstractions apparmor file that Streisand previously used to replace the distro provided profile. We only need to change two lines so rather than keep an entire profile up to date this commit adds a `replace` task to update the two obfsproxy binaries to have the correct `ix` permissions. We should look into fixing this in a better way in the future.

Resolves #778


jlund:
Looks good to me. Excellent work. Thanks for pulling this together.

PR #785, opened 2017-07-06
Run shellcheck against shell scripts

pgporada:


pgporada:
Just curious, do you want to do the same thing with flake8/autopep8 for python scripts and yamllint for yaml files? Like woodworking, I like my end product to be as refined as possible. I think that software craftsmanship is important and having as many files conform to a specific standard as possible is a step in the right direction. Let me know what you think!

cpu:
> Just curious, do you want to do the same thing with flake8/autopep8 for python scripts and yamllint for yaml files?

That would be great! In addition to the craftsmanship reasons you list (agreed!) I think having more linting & a common set of standards makes it easier for new contributors to submit PRs.

Issue #746, opened 2017-06-20
Support pick-and-choose modular services


cpu:
This issue is an "omnibus" issue to track first party support for pick-and-choose ("modular") Streisand services. This issue will track the proposed design & the overall progress made to convert individual services/roles.

## Motivation

Historically Streisand's anti-censorship goal has been the driving force behind providing many services pre-configured out-of-box. For users on censored & hostile networks it can be difficult to know what will work ahead of time and often the "best" or "most secure" choices are flat out unusable. This userbase continues to need a service like Streisand provides.

On the other hand, the recent Internet climate w.r.t Internet Service Provider surveillance has driven increased demand for a simple do-it-yourself VPN service usable on traditional uncensored networks, leading many to evaluate Streisand. From the perspective of this use-case Streisand's attack surface & complexity are a disadvantage and a risk.

How can a balance be struck? By allowing services to be enabled/disabled, both use-cases can be addressed. For the user's already served by Streisand who need to balance what works with the "academically pure" solution, all services can be enabled/included. For users with a more narrow & straightforward use-case just a bare minimum of the services they require can be included. 

## Challenge

The difficulty of providing this modularity lies in the interconnections between roles that exists today. Simply disabling the Shadowsocks role (for example) cascades out into other roles that have implicit dependencies on variables & files that the Shadowsocks role populates & creates. The documentation makes reference to Shadowsocks variables. The mirror role assumes that Shadowsocks clients are required. The firewall role requires knowledge of Shadowsocks to enable/disable port mappings.

The most immediate & straight forward way to address this challenge is to pepper the various roles with `if` statements evaluating whether or not a role was included. This approach (or variations of it) have been proposed in community contributed PRs in the past, but the complexity and overall brittleness of this approach make it hard to manage.

For this Issue we intend to exploit the relative similarity of the needs of a Streisand service to encourage isolated role design. Put simply, each service's role will be required to do its own firewall updates, its own documentation generation, and its own client mirroring. Base firewall/documentation roles will fill in the gaps by providing required system-wide dependencies (the `ufw` package being installed for instance) and stitch together index documentation based on which services were enabled/disabled.

## Base System

I consider the following to be the "base system" (e.g. these are not things you can pick-and-choose). These pieces are either integral to the goal of Streisand (things like a documentation web portal) or required for generic system access/administration (e.g. ufw, ssh).
* ufw
* nginx
* monit (likely to be replaced with better `systemd` units)
* ssh (the base sshd & config)
[...]

PR #794, opened 2017-07-09
Modular Streisand: Optional Shadowsocks

cpu:
This PR updates the Shadowsocks role to be optional, allowing it to be enabled/disabled by changing the `playbooks/group_vars/all` variable `streisand_shadowsocks_enabled` var. By default it is `true` and the base Streisand system remains unchanged. 

`documentation/modular_roles.md` describes the approach taken and more information can be found in #746.

Once this PR is merged I have a similar branch that makes Wireguard optional that I will submit. I'm deliberately converting the roles piece-wise and trying to keep the review pipeline relatively unclogged. Thanks!

_Edit: Apologies for the unrelated commits at the head of this PR, conflicts with master have been resolved so the diffs shown should be only from the modular Shadowsocks work. I've had this branch around for a while_

Updates https://github.com/jlund/streisand/issues/746


documentation/modular_roles.md
```diff
@@ -0,0 +1,162 @@
...
+Similarly, initially the `streisand-mirror` role downloaded all of the
+Shadowsocks clients for the mirror using a distinct tasks `.yml` with its own
+vars `.yml`. This made it fairly easy to move the mirror subtasks from
+`streisand-mirror/tasks/shadowsocks.yml` to
+`shadowsocks/tasks/mirror.yml` and `streisand-mirror/vars/shadowsocks.yml` to
+`shadowsocks/vars/mirror.yml`. Now the `shadowsocks` role can be responsible for
+preparing its own client mirror.
+
+Gateway & Mirror Indexes
+------------------------------
```
alimakki: Indexes -> Indices?

documentation/modular_roles.md
```diff
@@ -0,0 +1,162 @@
...
+
+While the Gateway & Mirror documentation can be largely agnostic of which
+services are included in a given build by outsourcing the management of
+subfolders of `/var/www/streisand` and `/var/www/streisand/mirror` eventually
+a consistent set of docs to navigate must be created based on the included
+services. This is done by conditionally adding links to the connection
+instruction & mirror indexes of each subservice using the enable bool vars in
```
alimakki: indexes -> indices?

alimakki: vars -> variables
(again, your call, not trying to bike-shed)

cpu:
> This should probably be a Phase Two thing, but one gap to consider is that we're still opening up the Shadowsocks port on the Azure/EC2/GCE side. I don't really have a super strong opinion on this, but it seems like it would be cleaner to continue to keep those tasks in separate roles (e.g. azure-security-group) and add some conditionals in those focused roles rather than bringing provider-specific logic (and lots of associated conditionals) into every role. What do you think?

Oops! I think I missed this chunk on this branch and fixed it in my wireguard branch. I intended to have conditionals here for the Shadowsocks port. I agree with your rationale that its best to suffer the conditionals in the provider specific firewall roles than try to move provider specific networking into each of the service roles. Unfortunately I couldn't think of something more clever like with dnsmasq.

cpu:
@alimakki @jlund Thank you for the detailed reviews! I believe all of your feedback has been resolved by fe4480a.

PR #795, opened 2017-07-09
Apply Shellcheck to all *.sh files, yamllint to all *.yml files in CI

cpu:
Previously we explicitly ran `shellcheck` against a fixed number of *.sh scripts from the Streisand repo as part of `.travis.yml` for CI. This PR adds a convenience wrapper script (`tests/shellcheck.sh`) that uses `find` to call `shellcheck` against *all* script files in the project directory without having to explicitly list them in the travis file. Note: this approach also includes the wrapper script too so it checks itself!

This PR also adds a similar wrapper script for [`yamllint`](https://github.com/adrienverge/yamllint) to lint all `*.yml` files in the project directory. A custom yamllint config is provided to allow for some legacy cruft (long lines) and @jlund's preference to visually align elements in lists.


cpu:
Blah! `realpath` doesn't exist on the Trusty CI environment :roll_eyes: - I will work around that and add a fix commit.

cpu:
Fixed using the same approach the `streisand` wrapper uses.

cpu:
This will break in CI until https://github.com/jlund/streisand/pull/796 is merged - there's a stale copy of the ansible go-role that was included by mistake (oops! :blush: ) that has shell scrpts that don't pass shellcheck. Turns out this PR was a useful way to catch mistakenly included crud!

cpu:
Add `yaml-lint` to this PR & updated the description.

cpu:
Everything is passing in CI now. Ready for a review!

PR #796, opened 2017-07-09
Removes ansible-go role.

cpu:
While adding end-to-end wireguard tests I accidentally committed the ansible-go galaxy role to the source tree.

This PR removes it outright since it isn't being used presently and was from a WIP branch not yet submitted for a PR.

Apologies!


cpu:
This will break in CI until https://github.com/jlund/streisand/pull/796 is merged - there's a stale copy of the ansible go-role that was included by mistake (oops! :blush: ) that has shell scrpts that don't pass shellcheck. Turns out this PR was a useful way to catch mistakenly included crud!

cpu:
Thanks @alimakki !

PR #802, opened 2017-07-13
Modular Streisand: Optional Wireguard

cpu:
This PR updates the Wireguard role to be optional, allowing it to be enabled/disabled by changing the `playbooks/group_vars/all` variable `streisand_wireguard_enabled` var. By default it is true and the base Streisand system remains unchanged.

Compared to https://github.com/jlund/streisand/pull/794 this was a more involved conversion so I've included some notes on required changes this PR adds along the way:

#### Support modular dnsmasq config.

Prior to this PR the `dnsmasq` role had an implicit dependency on both the `openvpn` role and the `wireguard` role. Listen address variables were referenced from the centralized `/etc/dnsmasq.conf` file in a single comma separated `listen-address` config parameter.

This PR changes the `/etc/dnsmasq.conf` template to only specify `listen-address=127.0.0.1` for the localhost bind. Subsequently the `openvpn` and `wireguard` roles are updated to include tasks that write
their own individual `openvpn.conf` and `wireguard.conf` files to `/etc/dnsmasq.d/` where they will be read in and amalgamated into one config at runtime. This allows roles to augment the base `dnsmasq` listeners when they need to without the `dnsmasq` role having to know about it.

#### Remove rc-local role in favour of modular approach.

Similarly, to avoid the one `rc-local` role having cross dependencies this PR removes it entirely. Instead individual roles can write their own service description to /etc/init.d/ and enable it using the standard mechanisms.

The `ip-forwarding`, `l2tp-ipsec`, `openconnect`, `openvpn` and `wireguard` roles are updated accordingly.

#### Documentation 

This PR updates `documentation/modular_roles.md` to describe the approach taken above. 

Updates #746


playbooks/roles/ip-forwarding/tasks/main.yml
```diff
@@ -1,6 +1,15 @@
...
   when: ansible_virtualization_type != 'lxc'
+
+- name: "Add IPv4 traffic forwarding persistence service to init"
+  copy:
+    src: streisand-ipforward.sh
+    dest: /etc/init.d/streisand-ipforward
+    mode: 0755
+
+- name: "Enable the streisand-ipforward init service"
+  command: "update-rc.d streisand-ipforward defaults"
```
jlund: We should be able to use the service module for all of these update-rc.d tasks:

```
- name: "Enable the streisand-ipforward init service"
  service:
    name: streisand-ipforward
    enabled: yes
```

documentation/modular_roles.md
```diff
@@ -160,3 +160,51 @@ Similarly, the index can conditionally generate a link to the service client mir
...
+    The most common convention adopted was to permit including a directory
+    full of configuration files, where anything dropped into that directory
+    would become active and part of that configuration.
+
+A concrete example of this common daemon problem is the `dnsmasq` service/role
+and its interactions with the `wireguard` and `openvpn` roles. Prior to
+modularization the `dnsmasq` role made sure the daemon listened on all required
+intefaces by having one `/etc/dnsmasq.conf` config file generated by the
+`dnsmasq` role's `dnsmasq.conf.j2` template. The template had one
+`listen-address` configuration parameter that listed all of the inferface
```
jlund: inferface -> interface

documentation/modular_roles.md
```diff
@@ -160,3 +160,51 @@ Similarly, the index can conditionally generate a link to the service client mir
...
+addresses in a comma-separated form:
+
+```
+listen-address=127.0.0.1,{{ dnsmasq_openvpn_tcp_ip }},{{ dnsmasq_openvpn_udp_ip }},{{ dnsmasq_wireguard_ip }}
+```
+
+This meant that if the `openvpn` or `wireguard` roles were disabled/removed the
+template for the `dnsmasq` config would fail from the missing variables.
```
jlund: s/from/because of/ or s/from/due to the/

playbooks/roles/ip-forwarding/tasks/main.yml
cpu: Ah cool! I didn't realize the `service` module did sysv init :+1:

cpu:
@jlund Thanks for the review! All of your feedback has been addressed. Ready for another :mag:

cpu:
Also noticed I had dropped the `+x` mode on the `streisand` wrapper script accidentally. Fixed in https://github.com/jlund/streisand/pull/802/commits/26cb490ef7f61e8b304eef51dfc109111e26c8a0

playbooks/roles/ip-forwarding/tasks/main.yml
cpu: FWIW, this generates a warning:
```
TASK [openconnect : Enable the streisand-openconnect init service] ******************************************************************
 [WARNING]: The service (streisand-openconnect) is actually an init script but the system is managed by systemd
```
I'm happy keeping it this way w/ the warning. It's a good reminder to write some systemd units at some point.

PR #816, opened 2017-07-19
Modular Streisand: Optional OpenVPN and stunnel

alimakki:
This PR modifies Streisand to make the OpenVPN role optional, which can be configured by modifying the `streisand_openvpn_enabled` to `no`, located in `playbooks/group_vars/all`.

~~Tested successfully on a DigitalOcean droplet with no apparent issues.~~

Updates #746.

Edit: Updated typo

Edit 2:  Since stunnel is entirely dependent on OpenVPN, I felt it was justified and make stunnel optional as well within the context of OpenVPN; by setting `streisand_stunnel_enabled: no` located in `playbooks/group_vars/all`, OpenVPN will be configured without stunnel profiles and the entire stunnel role will be skipped.

## Providers tested

Tested with the following settings in `playbooks/group_vars/all`

```
streisand_shadowsocks_enabled: yes
streisand_wireguard_enabled: yes
streisand_openvpn_enabled: yes
streisand_stunnel_enabled: no
```

and 

```
streisand_shadowsocks_enabled: yes
streisand_wireguard_enabled: yes
streisand_openvpn_enabled: no
streisand_stunnel_enabled: no # doesn't matter if `streisand_openvpn_enabled` is `no`
```

* [x] - Amazon 
* [ ] - Azure 
* [x] - DigitalOcean 
* [x] - Google 
* [x] - Linode 
* [ ] - Rackspace ??


cpu:
Hi @alimakki - I haven't missed this PR. Thank you! I will give it a review in the next couple of days.

P.s. The PR description says `streisand_tor_enabled` when I think you mean `streisand_openvpn_enabled`.

alimakki:
@cpu Thanks! Edited the message.

alimakki:
After some more playing around it made some sense to make stunnel a part of the OpenVPN, and conditional within it, that is you are now able to have and OpenVPN installation without stunnel.

playbooks/roles/openvpn/tasks/main.yml
```diff
@@ -239,38 +210,18 @@
...
+- include: mirror.yml
 
-
-- name: Copy the client files to the OpenVPN Gateway directory
-  command: cp --recursive {{ openvpn_path }}/{{ item }} {{ openvpn_gateway_location }}
-  args:
-    creates: "{{ openvpn_gateway_location }}/{{ item }}/ca.crt"
-  with_items: "{{ openvpn_clients }}"
+# Install stunnel
+# @alimakki - must be done after OpenVPN's Gateway
```
cpu: nit: I think it's clearer to use the `// NOTE(@alimakki) - ` syntax.

playbooks/roles/openvpn/vars/main.yml
```diff
@@ -24,3 +24,9 @@ openvpn_firewall_rules:
   - iptables -t nat -A POSTROUTING -s 10.9.0.0/24 -o {{ ansible_default_ipv4.interface }} -j MASQUERADE
 
 openvpn_gateway_location: "{{ streisand_gateway_location }}/openvpn"
+
+# If stunnel is enabled generating
```
cpu: nit: I think this comment could be fewer lines and still <80 chars a line.

playbooks/roles/openvpn/vars/mirror.yml
```diff
@@ -0,0 +1,39 @@
+---
+# OpenVPN Download variables
+# --------------------------
+openvpn_mirror_location:  "{{ streisand_mirror_location }}/openvpn"
+openvpn_mirror_href_base: "/mirror/openvpn"
+
+openvpn_samuli_seppanen_key_id: "0x29584D9F40864578"
```
cpu: This is unrelated to your PR, but would you mind updating these vars to not use "Samuli Seppanen" in the name? 

I think we should call it the `openvpn_gpg_release_key` or something like that and add a comment to the documentation page on the OpenVPN site that mentions that it's Samuli's key and he signs the releases. 

If you'd rather tag this as a follow-up issue that's fine with me too :-)

playbooks/roles/openvpn/vars/main.yml
cpu: Apologies - meant to leave these two comments as part of a review but hit the wrong buttons in my substandard coffee haze. I'll leave them as-is for now

playbooks/roles/sslh/templates/sslh.j2
```diff
@@ -14,4 +14,4 @@ RUN=yes
 # binary to use: forked (sslh) or single-thread (sslh-select) version
 DAEMON=/usr/sbin/sslh
 
-DAEMON_OPTS="-u sslh -p {{ ansible_default_ipv4.address }}:443 --ssh 127.0.0.1:{{ ssh_port }} --openvpn 127.0.0.1:{{ openvpn_port }} --ssl 127.0.0.1:{{ nginx_port }} -P /var/run/sslh/sslh.pid"
+DAEMON_OPTS="-u sslh -p {{ ansible_default_ipv4.address }}:443 --ssh 127.0.0.1:{{ ssh_port }} {% if streisand_openvpn_enabled %} --openvpn 127.0.0.1:{{ openvpn_port }} {% endif %} --ssl 127.0.0.1:{{ nginx_port }} -P /var/run/sslh/sslh.pid"
```
cpu: I think this is fine for now but it would be a lot less messy to read if we used the SSLH config file instead of the systemd defaults and the `DAEMON_OPTS`.

I opened https://github.com/jlund/streisand/issues/827 to remember that we can clean this up a little bit in the future. I don't see an easy way to avoid having `{% if  %}` blocks based on module vars but I think having it as a config file instead of a set of CLI parameters will help.

playbooks/roles/stunnel/meta/main.yml
```diff
@@ -1,3 +1,4 @@
 ---
 dependencies:
+  - { role: ufw }
```
cpu: Should there also be a dependency on the `openvpn` role? Maybe it doesn't matter since it's only being used from an `include_role` in the `openvpn` role? Thoughts?

cpu:
> Edit 2: Since stunnel is entirely dependent on OpenVPN, I felt it was justified and make stunnel optional as well within the context of OpenVPN; by setting streisand_stunnel_enabled: no located in playbooks/group_vars/all, OpenVPN will be configured without stunnel profiles and the entire stunnel role will be skipped.

Another thought: I totally agree with you that this is the right way to go. Maybe something that we should also add is a check (somewhere) that if the user sets `streisand_stunnel_enabled` to true but not `streisand_openvpn_enabled` there should be an error message & early exit.

alimakki:
@cpu, thanks for the review!
I applied all the requested changes you have mentioned, as well as created a validation role that executes before common in the `streisand.yml` which will fail if the openvpn optionals aren't rational.

playbooks/roles/sslh/templates/sslh.j2
alimakki: Agreed that this needs a cleanup.

alimakki:
@cpu My pleasure! My main concerns are Azure and GCE because of the optional firewall rules. I've never touched Rackspace before.

playbooks/roles/gce-network/tasks/main.yml
```diff
@@ -21,11 +21,11 @@
     credentials_file: "{{ gce_json_file_location }}"
     project_id: "{{ gce_project_id }}"
 
-- name: Open necessary TCP ports across most services in the GCE firewall
+- name: Open the nginx port across the GCE firewall
```
jlund: Let's change this to "Open the Nginx port in the GCE firewall"

alimakki: Done!

PR #829, opened 2017-07-23
Set GCE `ansible_user` to `root` to match SSH key metadata.

cpu:
Prior to this commit the GCE genesis role was setting the `ansible_user`
to `ubuntu`. This appears to be a mistake since in the earlier `gce`
task it adds the SSH key under the `root` user.

In testing a GCE provision recently I noticed that the cloud-init task
would fail with an SSH "Permission denied (publickey)" error. Similar
`raw` tasks would fail if the cloud-init include was commented out.

In practice using `root` as the `ansible_user` for GCE allows
provisioning to complete without error.

It's not clear to me why this didn't break before since the
`ansible_user` has been static for some time... Mysterious!

### DIFF
--- a/playbooks/roles/genesis-google/tasks/main.yml
+++ b/playbooks/roles/genesis-google/tasks/main.yml
@@ -28,7 +28,6 @@
   add_host:
     name: "{{ streisand_server.instance_data[0].public_ip }}"
     groups: streisand-host
-    ansible_become: yes
 
 - name: Set the streisand_ipv4_address variable
   set_fact:
--- a/playbooks/roles/genesis-google/tasks/main.yml
+++ b/playbooks/roles/genesis-google/tasks/main.yml
@@ -28,7 +28,6 @@
   add_host:
     name: "{{ streisand_server.instance_data[0].public_ip }}"
     groups: streisand-host
-    ansible_user: root
     ansible_become: yes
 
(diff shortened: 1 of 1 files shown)


playbooks/roles/genesis-google/tasks/main.yml
```diff
@@ -28,7 +28,7 @@
   add_host:
     name: "{{ streisand_server.instance_data[0].public_ip }}"
     groups: streisand-host
-    ansible_user: ubuntu
+    ansible_user: root
```
jlund: I think we can just omit the `ansible_user` and `ansible_become` lines entirely as we do for other providers that use `root` (e.g. DigitalOcean). The `streisand.yml` playbook [specifies default values for these variables](https://github.com/jlund/streisand/blob/master/playbooks/streisand.yml#L8) and it looks like we don't need to override them anymore.

My speculative guess is that GCE has moved away from the `ubuntu` user in their images in an attempt to be more consistent, though they've now diverged from Azure and EC2 who still don't use `root` for new installs. I will double-check to make sure I am seeing this behavior in my GCE account as well.

cpu: > I think we can just omit the ansible_user and ansible_become lines entirely as we do for other providers that use root

Good point! fixed in 6de38f7

PR #835, opened 2017-07-26
Update OpenVPN mirror GPG key ID and fingerprint

alimakki:


cpu:
I'm in favour of trying to find a longer term fix since upstream is flip-flopping between the same two keys. Thoughts?

alimakki:
@cpu, I agree there should be a longer term solution for this, and will follow up in a separate PR.

PR #845, opened 2017-07-30
Optional tinyproxy

alimakki:
This PR modifies Streisand to make the tinyproxy role optional, which can be configured by modifying the `streisand_tinyproxy_enabled` to `no`, located in `global_vars/vars.yml`.

Updates #746.

### DIFF
--- a/global_vars/vars.yml
+++ b/global_vars/vars.yml
@@ -11,5 +11,6 @@ streisand_openvpn_enabled: yes
 streisand_stunnel_enabled: yes
 streisand_tor_enabled: yes
 streisand_openconnect_enabled: yes
+streisand_tinyproxy_enabled: yes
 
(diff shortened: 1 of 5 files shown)


jlund:
Tinyproxy is tied to SSH tunneling support and isn't currently used in any other areas. There are sections of the SSH documentation that refer to port 8888 (Tinyproxy's default) that should be wrapped in `{% if streisand_tinyproxy_enabled %}` conditionals because they won't be relevant or functional if this role is disabled.

This role could arguably become completely subservient to the SSH role and pulled in with an `include_role` task (a la stunnel and OpenVPN).

We can cover all of the above as two separate PRs. I just wanted to get it documented somewhere. I'm fine merging this one as-is if that's easiest.

alimakki:
@jlund, thanks for the review.

In the interest of Do The Right Thing&#8482;, I'll take another pass at this and handle the SSH documentation as well.

PR #846, opened 2017-07-30
Variables missing for modular Streisand

ridercz:
The new modular Streisand expects some configuration variables to be set and they aren't.


alimakki:
Hi @ridercz, all of these variables are defined in the [global_vars](https://github.com/jlund/streisand/blob/master/global_vars/vars.yml) files.

Is there a specific issue that you are facing?

PR #873, opened 2017-08-06
Refactor the Nginx restart tasks to support modularity

jlund:
Previously, the Nginx restart handler was located in the Nginx role. The `tor-bridge` and `streisand-gateway` roles declared a dependency on the Nginx role and notified the same handler for their respective vhost template tasks.

The issue with this approach in a modular world is that the Tor role runs prior to the Gateway role. The Nginx role (and its handler) are skipped as declared dependencies of the Tor role if the `streisand_tor_enabled` variable is set to 'no'. Later on, the Nginx tasks run successfully when the Gateway role also declares a dependency on Nginx. However, Ansible still treats the handler as having been skipped and won't execute the restart task because handlers only run once and the first skip takes precedence.

This PR moves the handlers into each role with unique names (which also helps sidestep another issue as well: ansible/ansible#18594).

### DIFF
--- a/playbooks/roles/nginx/tasks/main.yml
+++ b/playbooks/roles/nginx/tasks/main.yml
@@ -26,7 +26,6 @@
     owner: root
     group: root
     mode: 0644
-  notify: Restart Nginx
 
 - name: Set up Nginx vhost directories
   file:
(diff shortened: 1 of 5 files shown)


nopdotcom:
From what I understand of the system, this is a good way to modularize. I think somebody else should review it though, since even the symptom is beyond my Ansible knowledge.

PR #877, opened 2017-08-08
Always do check_python()

wzyboy:
On Arch Linux `/usr/bin/python` is Python 3. When executing the playbook
on an Arch Linux laptop, there are actually some modules executed
locally, e.g. `ec2_*` ones. Without proper `ansible_python_interpreter` set
these tasks would fail.


wzyboy:
Sure, I have deleted redundant invocations.

PR #881, opened 2017-08-09
shadowsocks: Make installation idempotent

wzyboy:
When configuring with `--prefix=/usr`, the binaries are installed under
the directory `/usr/bin` instead of `/usr/local/bin`.

### DIFF
--- a/playbooks/roles/shadowsocks/tasks/main.yml
+++ b/playbooks/roles/shadowsocks/tasks/main.yml
@@ -48,7 +48,7 @@
   command: make install
   args:
     chdir: "{{ shadowsocks_compilation_directory }}"
-    creates: "/usr/local/bin/ss-server"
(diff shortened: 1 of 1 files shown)


nopdotcom:
Should this be in `/usr/local` anyway?

wzyboy:
@nopdotcom Maybe. Considering all other components compiled from source are installed under `/usr/local/bin`. I'll make the changes.

wzyboy:
I see where the problem is: the systemd service file of shadowsocks is not generated with `--prefix`. It has hard-coded its executable path to be under `/usr/bin`.

Now I see why the original version of this role uses `--prefix=/usr` flag. So I shall revert to the original PR: keep the executables under `/usr/bin` and let Ansible check for `/usr/bin/ss-server` instead of `/usr/local/bin/ss-server`.

PR #883, opened 2017-08-10
ssh-forward: Allow port forwarding only

wzyboy:
By default one can log into the Streisand server with the SSH private
key, which is intended for port forwarding. This poses a security issue
if a careless sysadmin sets open permissions on critical files.
Prepending a force command to authorized_key is a simple way to
prevent SSH login with that key.

### DIFF
--- a/playbooks/roles/ssh-forward/templates/instructions.md.j2
+++ b/playbooks/roles/ssh-forward/templates/instructions.md.j2
@@ -109,7 +109,7 @@ Sshuttle offers significantly faster performance and is easier to set up, but th
 
 1. SSH into the server and forward a dynamic SOCKS port:
 
-   `ssh -D {{ ssh_default_socks_port }} forward@{{ streisand_server_name }}`
+   `ssh -ND {{ ssh_default_socks_port }} forward@{{ streisand_server_name }}`
 1. Verify that the fingerprint matches one of these:
 
    `{{ ssh_server_fingerprints.results[0].stdout }}`
--- a/playbooks/roles/ssh-forward/templates/instructions.md.j2
+++ b/playbooks/roles/ssh-forward/templates/instructions.md.j2
@@ -8,6 +8,7 @@ SSH Tunnel
 {% if streisand_tinyproxy_enabled %}
   * [Android](#android)
 {% endif %}
+  * [Notes](#notes)
 
 <a name="windows"></a>
 ### Windows ###
@@ -136,3 +137,6 @@ Some applications allow you to make these settings persistent for all networks. 
 6. Set the value of *network.proxy.socks\_remote\_dns* to `true`.
 7. Set the value of *network.proxy.type* to `1`.
 {% endif %}
+
+<a name="notes"></a>
+1. If you see the message "This account is for port forwarding only", make sure you configure your SSH client to not execute shell commands on remote server (`-N`).
--- a/playbooks/roles/ssh-forward/templates/instructions.md.j2
+++ b/playbooks/roles/ssh-forward/templates/instructions.md.j2
@@ -5,7 +5,6 @@ SSH Tunnel
 * Platforms
   * [Windows](#windows)
   * [Linux and OS X](#linux-and-osx)
-  * [Linux and OS X (alternate)](#linux-and-osx-alternate)
 {% if streisand_tinyproxy_enabled %}
   * [Android](#android)
 {% endif %}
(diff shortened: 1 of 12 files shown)


cpu:
@wzyboy Thanks for the PR!! I appreciate the intention of this work but I'm not sure it's a complete unit yet.

> This poses a security issue if a careless sysadmin sets open permissions on critical files.

That's true but I think your argument would be stronger by focusing on the way it increases the threat surface of the Streisand instance considerably with respect to local root exploits & accessing management interfaces that are bound to localhost only. Streisand isn't designed to be modified by sysadmins (careless or careful!) after install. If the base install has open permissions on critical files we should definitely fix that!

@jlund Can you speak to the original intention of the forward user? E.g. was it a conscious decision to offer a shell account for the forward user? I'm not sure if you made this decision purposefully to allow access to `ssh`, `curl`, other shell niceties or if it was just the way things worked out at the time. If so then this PR would be removing that feature.

@wzyboy If it wasn't intentional to have a shell user then I still think this is missing a little bit of docs & due diligence. E.g. Linux `ssh` users would need to add a `-N` to keep the connection open without creating a remote shell that would be closed by the authorized keys restriction. What do Windows/Putty users need to do? How does this interact with TinyProxy?

I think answering the questions above would be a prerequisite for this landing in master.

wzyboy:
> accessing management interfaces that are bound to localhost only

Good point. This reminds me that even a forward only account could always reach localhost ports by `ssh -L`. To make Streisand server a strong one, we must be confident even an SSH account (without root privilege) is compromised, the bad guy cannot get hold of the information of other users on the same server. I can see Streisand is do good here, like not logging anything on disk.

> If it wasn't intentional to have a shell user then I still think this is missing a little bit of docs & due diligence. E.g. Linux ssh users would need to add a -N to keep the connection open without creating a remote shell that would be closed by the authorized keys restriction. What do Windows/Putty users need to do? How does this interact with TinyProxy?

When I do an dynamic port forwarding, I always use `-N` in combination with `-D`. I could add this to docs. As of Windows SSH software, I am not very sure. But shell access is definitely not needed for pure port forwarding.

cpu:
> As of Windows SSH software, I am not very sure. But shell access is definitely not needed for pure port forwarding.

I'm still waiting on @jlund to express whether there was the intention to provide more than just port forwarding from this user.

I don't think this is mergeable until all platforms have working instructions and any potential negative interaction with TinyProxy is ruled out.

jlund:
The main reason that the `forward` user got a shell is because [sshuttle](https://github.com/sshuttle/sshuttle) didn't work without one. Perhaps that limitation has been addressed now though? We would also need to test the clients on other platforms as well.

Otherwise, I'm not opposed to this at all! We previously had dropped support for sshuttle too, and I would be open to considering that again if we think that the security improvements are worth that tradeoff.

cpu:
@wzyboy Are you interested in addressing the Windows client/documentation questions and testing that this doesn't affect `sshuttle`? If not that's totally fine but I would like to close the PR. Thanks!

wzyboy:
@cpu My test results:

### Putty
Additional configuration options (equivalent to `-N`) are required if we disable shell access. I have added how-to in docs: https://github.com/jlund/streisand/pull/883/commits/cdb3dfa85c29ed370ae4be9e98a623fe906314b1

But I am not sure if the French translation is correct since I cannot speak French (thanks Google Translate).

### sshuttle
The software cannot work without shell access, because it tries to invoke `/bin/sh -c python ...` on the remote server.

### OpenSSH client
Adding `-N` is sufficient. I have already addressed this issue in [a previous commit](https://github.com/jlund/streisand/pull/883/commits/700dd10f4e92386ae79c579403b88865fed29763)

### Tinyproxy
Disabling shell access does not effect Tinyproxy.

---

My suggestion is that we drop support on sshuttle entirely. I do not see any advantage of sshuttle over native OpenSSH client on Linux / OS X. What do you think @jlund ?

nopdotcom:
Has anybody benchmarked sshuttle vs pure SSH tunneling over a vaguely lossy, slow network? If the advertising is as it says, sshuttle avoids the worst of the TCP-over-TCP problems. (That one has bitten me before, although it was a long time ago.) 

(I wonder how hard it would be to set up a mosh session as a jump terminal for non-admins, i.e., you don’t get a streisand shell, but you can connect onward from there.)

wzyboy:
> Has anybody benchmarked sshuttle vs pure SSH tunneling over a vaguely lossy, slow network? If the advertising is as it says, sshuttle avoids the worst of the TCP-over-TCP problems. (That one has bitten me before, although it was a long time ago.)

SSH port forwarding was very popular in China ~10 years ago (due to its low cost and easy to set up), before it was blocked by the national firewall (GFW) ~5 years ago. Based on my own experience during that time, native OpenSSH client (`ssh -D`) does not scale well if the number of concurrent connections is high. However, thanks to the wider and wider deployment of HTTP/2 nowadays, the number of concurrent connections when surfing the web (duh...) is now much lower compared to ~10 years ago.

I have not been using SSH port forwarding to bypass the censorship for years. But even back then, one can optimize the performance of SSH port forwarding by simply using a different client. I remember that [Bitvise](https://www.bitvise.com/ssh-client) was popular back then. It's a closed-sourced free software for Windows and it scales well when the number of concurrent connections is high.

A side note: Per an original post by the author of Shadowsocks, the software was originally written because the author cannot bear the low performance of native OpenSSH client when the number of concurrent connections is high.

> (I wonder how hard it would be to set up a mosh session as a jump terminal for non-admins, i.e., you don’t get a streisand shell, but you can connect onward from there.)

Sorry I don't get it. Mosh does not support port forwarding; an Streisand server admin could always log in as `root` or `ubuntu`, instead of `forward`.

cpu:
> My suggestion is that we drop support on sshuttle entirely. I do not see any advantage of sshuttle over native OpenSSH client on Linux / OS X. What do you think @jlund ?

I'm a +1 to dropping `sshuttle`. I seem to remember that @jlund said it was included to help Android users proxy apps like the twitter client that (at the time) did not support configuring a SOCKS proxy. I _think_ that use-case isn't relevant anymore and there are better options available on Android these days.

playbooks/roles/ssh-forward/templates/instructions.md.j2
```diff
@@ -19,6 +19,8 @@ SSH Tunnel
    * Port `443` is available as a fallback option if you are on a network that restricts access to the default SSH port.
 1. Go to Connection --> Data.
 1. Enter `forward` in the *Auto-login username* field.
+1. Go to Connection --> SSH.
+1. Check `Don't start a shell or command at all`.
```
cpu: Could you also add a small section on what the error looks like if you forget this step? Something like "If you see the message "this account is for port forwarding only" ensure you have selected "Don't start a shell or command at all" in your Putty configuration". ?

cpu: Maybe this should be at the very bottom of the instructions so that it can reference the Linux solution as well. "If you see xxx then make sure you are providing -N in your ssh command line`.

wzyboy: @cpu I have added some clarification at the end of the documentation. The French part is machine translated.

wzyboy:
sshuttle support has been dropped in  https://github.com/jlund/streisand/pull/883/commits/7ac4259a327e804ce100ac3c68870cefcf663b00 .

cpu:
@jlund Can you +1 this? Primarily want to run dropping `sshuttle` by you before merging.

nickolasclarke:
@nopdotcom @wzyboy hit this on the head prettey well.  My experience last year was that while performance was greatly improved over normal ssh tunneling, it is also quickly detected by the GFW and connections appear to be dropped or severely throttled. In China, at least, it is not a viable solution in all but the most desperate cases.

wzyboy:
@nickolasclarke Yes, like I said, SSH port forwarding was blocked by GFW ~5 years ago. GFW has been able to tell the difference between a normal SSH session and a port forwarding session by the different traffic pattern. As soon as GFW detects an SSH port forwarding session, the (ip_addr, ssh_port) tuple would be blocked sooner or later. Changing the SSH port will give you a couple more days but it's just a cat and mouse game.

cpu:
@jlund was a +1 out-of-band for dropping `sshuttle`. I'm going to merge this PR and we can always come back around if it turns out there's a significant userbase that was relying on `sshuttle` and an interactive `forward` user.

Issue #639, opened 2017-04-23
issue SSL cert via Lets Encrypt


chipitsine:
we can issue SSL cert using lets encrypt (actually, we need to renew it every 2-3 months later)

cpu:
Let's Encrypt does not presently issue for IP addresses, this feature request would require support for configuring a domain name.

cpu:
Related: https://github.com/jlund/streisand/issues/60

wzyboy:
I have submitted a PR which implements this function: https://github.com/jlund/streisand/pull/911

PR #885, opened 2017-08-10
[WIP] New role: nghttpx as an HTTP/2 proxy

wzyboy:
This is a proof of concept implemention of what is described in https://github.com/jlund/streisand-discussions/issues/37 .

It's quick and dirty. There are no docs or tests for now. And as mentioned in above linked discussion thread, nghttpx requires a valid certificate to work (self-signed cert does not work). In this POC I just copied a snakeoil cert for demonstration purposes. You may wish to replace `/etc/nghttpx/server-{cert,key}.pem` with your own (valid) ones and `systemctl restart nghttpx` to test it.

I'll write detailed testing steps later.


wzyboy:
Testing steps:

1. You need a valid certificate (could be bought for a few bucks online);
2. Use a simple playbook `streisand/playbooks/nghttpx.yaml` to run only the nghttpx role:
```
- hosts: your-server-here
  roles:
    - role: nghttpx
```
3. Log into the server and replace `/etc/nghttpx/server-{cert,key}.pem` with your real certificate and key before restarting nghttpx daemon with `systemctl restart nghttpx`;
4. Point your certificate domain to the server IP address. You could actually change your domain records, or you could do this by editing your `/etc/hosts` file of your client computer;
5. Set your browser proxy to `https://your-domain:10443`. If you are using Chromium / Google Chrome, you could follow steps [here](https://www.chromium.org/developers/design-documents/secure-web-proxy) by simply launching your Cr with `--proxy-server=https://your-domain:10443`. Note it's `https://`, not `http://`. And `curl` does not supports "HTTPS in HTTPS" proxy so you have to use a modern browser;
6. [Test](https://www.google.com/search?q=my+ip+address) if your IP address is now the public address of your server.

Side notes:
- You could use WireShark, `tcpdump`, or similar tools to inspect the traffic, it looks just like you are browsing an HTTPS website. But under the hood, the server is fetching and delivering the content for your through encyprted HTTPS connection. It's a forward proxy implemented in HTTP/2.
- This nghttpx does not require any authentication for now. Authentication could be done during TLS handshake (client certificate), or via HTTP auth ([Squid](https://wiki.squid-cache.org/Features/Authentication)), or both. If I have time I will implement authentication.

playbooks/roles/nghttpx/tasks/main.yml
```diff
@@ -0,0 +1,107 @@
...
+#
+#- name: "Convert the key and certificate into PKCS #12 format"
+#  expect:
+#    command: certtool --to-p12 --load-privkey {{ nghttpx_client_key_file }} --pkcs-cipher 3des-pkcs12 --load-certificate {{ nghttpx_client_certificate_file }} --outfile {{ nghttpx_client_pkcs12_file }} --outder
+#    responses:
+#      "Enter a name for the key": "{{ nghttpx_client_name }}"
+#      "Enter password": "{{ nghttpx_pkcs12_password.stdout }}"
+#      "Confirm password": "{{ nghttpx_pkcs12_password.stdout }}"
+#    creates: "{{ nghttpx_client_pkcs12_file }}"
+
```
alimakki: Commented out sections should be deleted if not needed

playbooks/roles/nghttpx/tasks/main.yml
```diff
@@ -0,0 +1,107 @@
...
+# Set up firewall rules
+- include: firewall.yml
+
+# Generate Gateway documentation
+#- include: docs.yml
+
+# nghttpx has only source code tarballs hosted on GitHub.
+# There is no need to mirror them.
+#- include: mirror.yml
```
alimakki: Same as above

playbooks/roles/nghttpx/vars/main.yml
```diff
@@ -0,0 +1,64 @@
...
+  #- libjansson-dev
+  - libc-ares-dev
+  - libjemalloc-dev
+  - libsystemd-dev
+  #- libspdylay-dev
+  #- cython
+  #- python3-dev
+  #- python-setuptools
+  - gnutls-bin  # certtool
+
```
alimakki: Commented out code

playbooks/roles/nghttpx/tasks/main.yml
wzyboy: As the title suggests, this is a WIP. There is no authentication for now. These certificate generation code are adapted from OpenConnect role. They are commented out because I would like to implement them later.

playbooks/roles/nghttpx/vars/main.yml
wzyboy: IMHO the Streisand project itself is a good-written cookbook of Ansible, isn't it? People write comments in code, so in the future it's easier to get the idea what does the code do. These lines are commented out by intention.

wzyboy:
@alimakki I have included the role in `global_vars` but keep it disabled by default. I could not agree with your other suggestions though.

playbooks/roles/nghttpx/tasks/main.yml
wzyboy: After I have implemented all the features (when I remove "[WIP]" prefix from the title), I would do a clean up.

playbooks/roles/nghttpx/tasks/main.yml
cpu: It makes it considerably easier for us to review your work if you only include the parts that are done and leave the note on future work for the PR description or a brief non-code comment :-)

playbooks/roles/nghttpx/vars/main.yml
cpu: I might be misunderstanding and this is a byproduct of being a WIP not intended to be merged but we wouldn't merge commented out code to master. Comments should be written for humans not disabled code :-)

wzyboy:
Based on feedbacks from @alimakki and @cpu , there are multiple issues with this PR:
- [X] Too many unused code. Clean up is needed.
- [ ] The service is not authenticated. Authentication is needed.
- [ ] The service requires a user to provide a valid certificate. Automation should be used instead.

I have cleaned up the code. I'll add HTTP basic auth with Squid to the service this weekend.

The certificate issue would be resolved once the role of Let's Encrypt is done. (https://github.com/jlund/streisand/issues/639)

wzyboy:
The role has now documentation. The rest of work could only be finished after the Let's Encrypt role is finished, as discussed [here](https://github.com/jlund/streisand-discussions/issues/37).

playbooks/roles/nghttpx/files/snakeoil.pem
```diff
@@ -0,0 +1,46 @@
+-----BEGIN PRIVATE KEY-----
```
cpu: There shouldn't be a hardcoded private key used here. If you need a self-signed certificate generate it with a fresh key pair on the instance.

playbooks/roles/nghttpx/tasks/install.yml
```diff
@@ -0,0 +1,28 @@
...
+  unarchive:
+    remote_src: yes
+    src: "{{ nghttpx_source_location }}"
+    dest: /usr/local/src
+    owner: root
+    group: root
+    creates: "{{ nghttpx_compilation_directory }}/README"
+
+- name: Compile and install nghttpx
+  shell: ./configure && make -j {{ ansible_processor_cores }} && make install && install -Dm755 contrib/nghttpx.service /lib/systemd/system/nghttpx.service
```
cpu: nghttpx isn't packaged on 16.04? It must be compiled from source?

playbooks/roles/nghttpx/templates/nghttpx.conf.j2
```diff
@@ -0,0 +1,12 @@
+frontend=0.0.0.0,{{ nghttpx_port }}
+backend=127.0.0.1,{{ squid_port }}
+http2-proxy=yes
+certificate-file={{ nghttpx_server_certificate_file }}
+private-key-file={{ nghttpx_server_key_file }}
+workers={{ ansible_processor_cores }}
+add-x-forwarded-for=no
+no-via=yes
+no-ocsp=yes
```
cpu: Why is this required?

playbooks/roles/nghttpx/files/snakeoil.pem
wzyboy: Self-signed certifcates do not work at all, as well as the hardcoded keys. But nghttpx won't start without a key. So I just put a fake one here to make `service nghttpx start` work. Once the work on Let's Encrypt is done, this would be removed.

playbooks/roles/nghttpx/tasks/install.yml
wzyboy: No, it can be installed from repository. I just see software in other roles are compiled from the source and followed the convention. If you think installing from repository is better, I can change it.

playbooks/roles/nghttpx/templates/nghttpx.conf.j2
wzyboy: Actually this is for nghttpx client side. In heavyly-censored networks, querying OCSP often results in timed out.

playbooks/roles/nghttpx/files/snakeoil.pem
cpu: What I'm saying is that you can generate this private key & certificate on the fly versus hardcoding them.

wzyboy:
> My vote is that we do not include nghttpx at this time. I appreciate the work you've put into this, apologies that it isn't the outcome you hoped for.

I tought we have already agreed on not merging this at this time? I have said that nghttpx would only be merged after the work on Let's Encrypt is done.

I may be misunderstanding, but: are you saying the PR should not be merged *now* (I already agreed with that in previous comments), or the PR should be dropped entirely even if Let's Encrypt role is finished?

playbooks/roles/nghttpx/tasks/install.yml
cpu: Yes, much better to install from package where possible because that way the unattended upgrades configuration ensures they get security updates automatically. With source compiled software if you install Streisand today, you will be on that version forever. Nothing updates the source built components down the road for new patches/fixes.

cpu:
> I may be misunderstanding, but: are you saying the PR should not be merged now (I already agreed with that in previous comments), or the PR should be dropped entirely even if Let's Encrypt role is finished?

I think there are concerns I have that would remain even after the HTTPS questions are resolved.

As is I'm feeling that there's just too much software in Streisand to manage and not enough time. I would prefer to try and do more with what we have than add new software.

wzyboy:
Okay.

PR #887, opened 2017-08-11
RFC: ocserv Android backup passphrase instructions

nopdotcom:
I just deployed another server and got AnyConnect asking for a username and password. I don’t see any pattern to this bug. (Password length??)

Until we get ocserv figured out, here are instructions for a workaround. It would need translation.

Note that the passphrase is already in the file.

(Ahh, git mismanagement.)


playbooks/roles/openconnect/templates/instructions.md.j2
```diff
@@ -106,6 +106,11 @@ OpenConnect is an extremely high-performance and lightweight VPN server that als
 1. If this is your first time using AnyConnect, you will need to accept the Connection Request dialog that Android displays.
 1. You should be good to go! You can verify that your traffic is being routed properly by [looking up your IP address on Google](https://encrypted.google.com/search?hl=en&q=ip%20address). It should say *Your public IP address is {{ streisand_ipv4_address }}*.
 
+Some users have reported that AnyConnect prompts for a username and password. This is a bug; please report it. If this happens to you, use these credentials as a workaround:
```
cpu: Who should the bug be reported to? Streisand? AnyConnect? Can you make the text a link to the correct repositories's new issue link?

nopdotcom:
When in doubt, write more. Uses different parts of the brain than debugging, I guess.

nopdotcom:
I think this is complete. Take another glance, @cpu?

PR #892, opened 2017-08-18
russian tranlation

levleontiev:
I have translated README to Russian, as I really think that taking into account recent internet blocking issues in Russia, this gonna be very useful.


levleontiev:
Thanks , fixed.

alimakki:
Thank you @levleontiev!

PR #895, opened 2017-08-19
Eliminate the EC2 90-second pause for ssh

nopdotcom:
There's a 90-second pause in EC2 provisioning where we wait for SSH to
start working. The bug this worked around was not reproducible recently. Delete the pause.

----

Original message:

My guess was that /dev/random is starved for entropy. Use haveged at cloud-init to start gathering some entropy.

Do we trust/like haveged?

I’m leaving the 90-second pause commented out (instead of deleted), in case somebody suddenly runs into this bug again. I’m open to opinion on that.

### DIFF
--- a/playbooks/roles/genesis-amazon/tasks/main.yml
+++ b/playbooks/roles/genesis-amazon/tasks/main.yml
@@ -100,9 +100,3 @@
 - name: Set the streisand_server_name variable
   set_fact:
     streisand_server_name: "{{ aws_instance_name | regex_replace('\\s', '_') }}"
-
-# This does not seem to be necessary if /dev/random has been filled by haveged.
-#
-#- name: New EC2 servers are occasionally slow to process incoming SSH connections e#ven after the OpenSSH daemon has started up. Pause for 90 seconds.
-#  pause:
-#    seconds: 90
--- a/playbooks/roles/genesis-amazon/tasks/main.yml
+++ b/playbooks/roles/genesis-amazon/tasks/main.yml
@@ -47,11 +47,6 @@
     instance_tags:
       Name: "{{ aws_instance_name }}"
     wait: yes
-    # This is gross, but it seems like sshd is starved for entropy on boot. We can't
-    # ssh in to fix this...
-    user_data: |
-      #!/bin/sh
-      sudo apt-get install haveged -y
   register: streisand_server
 
 - name: Create CloudWatch alarm to auto-recover instance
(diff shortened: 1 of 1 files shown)


cpu:
> My guess is that /dev/random is starved for entropy. 

Is there a way to verify this guess?

> Do we trust/like haveged?

If the problem is a lack of entropy at boot I think its as good a solution as is available.

> I’m leaving the 90-second pause commented out (instead of deleted), in case somebody suddenly runs into this bug again. I’m open to opinion on that.

In general I think if we aren't confident this fix will work in 100% of cases then I'd rather not merge it and leave the pause. It may not be the most elegant solution but it seems to work :-)

nopdotcom:
I wasn’t around for the 90-second bug being squashed, and I haven’t been able to replicate it. So I can’t really tell what is blocking SSH from accepting incoming connections. This PR just waves a dead chicken over it and say “more entropy guud”. You’re right to be skeptical.

cpu:
> I wasn’t around for the 90-second bug being squashed, and I haven’t been able to replicate it. 

same! @jlund Do you remember the dark days that spawned this workaround?

> This PR just waves a dead chicken over it and says “more entropy guud”. You’re right to be skeptical.

:laughing:

jlund:
> same! @jlund Do you remember the dark days that spawned this workaround?

1. Users started reporting it one day out of nowhere; everything had been fine for a long time.
2. I was able to duplicate it consistently.
3. Adding an arbitrary 90-second delay seemed to work.
4. Users stopped reporting that they were experiencing the problem.

This fix from the dark days was also very dead chicken-esque :)

cpu:
@jlund Thanks for weighing in with some history!

@nopdotcom Have you been able to replicate the error removing the 90 second pause and not including haveged?

nopdotcom:
No, I’ve never been able to replicate the losing condition, even without haveged. I just deployed from 15920bf1cce7f6c45c6bbf95273bd452aecd5fba with the pause commented out; it worked fine for EC2 Ohio->Singapore, Ohio->Seoul. It also worked fine on a local OS X->Seoul.

I bet some combination of Ansible and AWS has fixed the “SSH starts slowly” bug. If we take it out, the good news is that if the bug reappears, it would be really obvious from the log.

cpu:
> No, I’ve never been able to replicate the losing condition, even without haveged. 

Ok! I think my preference would be to remove the pause but not add haveged. If a poor soul finds their way to the repo with a problem smelling like this we can revisit haveged. 

@jlund @nopdotcom Sounds reasonable to you folks?

cpu:
@nopdotcom Can you update the title & description of this PR now that it only removes the pause and doesn't involve haveaged at all?

PR #896, opened 2017-08-20
Modular Streisand: optional l2tp/ipsec.

cpu:
This commit introduces modular support for l2tp/ipsec. The role is
included/skipped based on the var `streisand_l2tp_enabled`, which by
default is 'true'.

The `rsyslog` role is now removed since both `openconnect` and `l2tp`
modify the configuration as required individually. I moved the restart
handler into each role to avoid the need for a centralized `rsyslog`
role.

Updates #746 

Tested:
* [x] - Amazon 
* [ ] - Azure 
* [x] - DigitalOcean 
* [x] - Google 
* [x] - Linode 
* [ ] - Rackspace


alimakki:
@cpu, great stuff! 
My only concern is with one particular task in `playbooks/l2tp-ipsec/tasks/main.yml`:

```
- block:
    - name: Include the L2TP/IPsec sysctl value variables
      include_vars: sysctl-values.yml

    - name: Apply the sysctl values
      sysctl:
        name: "{{ item.key }}"
        value: "{{ item.value }}"
      with_items: "{{ l2tp_ipsec_sysctl_values }}"
  when: ansible_virtualization_type != 'lxc'
```

Wouldn't it be better if this was refactored into the common role?

cpu:
> Wouldn't it be better if this was refactored into the common role?

@alimakki Which common role were you thinking? There's a fair number of sysctl's being set by the l2tp role and I'm not sure which should be considered common and which are specific to libreswan. Guidance welcome!

alimakki:
@cpu, I'll go through the list later tonight and follow up  :)

cpu:
>I think the sysctl refactoring can go safely go into another PR.

Sounds good to me. 

> The one issue that I found is that skipping the OpenConnect role will cause the handler for restarting rsyslog to be skipped. This is similar to #873.

Good catch! Thanks. Fixed with a0fb0f5

cpu:
Tested on/off in GCE and Linode. :white_check_mark:

cpu:
I filed https://github.com/jlund/streisand/issues/908 for the sysctl follow-up.

PR #897, opened 2017-08-20
Multi-clients

alimakki:
This PR refactors Streisand to better help allows the various VPN roles in generating multiple clients.

Rather than having a user define how many clients are generated by way of appending elements to a list, a more simplified approach can be achieve by simply specifying the number of clients desired, and having the Streisand script to generate randomly named clients.

In order to adjust this value, simply edit the `global_vars\vars.yml#vpn_clients` variable to a desired number.

A validation has been added to the `validation` role to ensure that the number of VPN clients is set to a reasonable number (can't be less than 1), and an arbitrarily imposed maximum of 20 clients.

As OpenVPN has already been generating multiple clients, a simple refactor was needed to take into account the 'new state of the universe'.

The OpenConnect role is now able to generate multiple client certificates which can be used on various platforms (Android, iOS, etc)

### Tested ###
- [x] Amazon
- [ ] Azure
- [x] DigitalOcean
- [x] Google
- [x] Linode
- [ ] Rackspace


alimakki:
Also thanks to @nopdotcom for helping craft some language for the modified OpenConnect role :)

nopdotcom:
I’m wondering if we want to prefix the names of files: `client-Pynchon-fabulosity.p12`. I know I will misplace them if they are just random words in ~/Downloads.

alimakki:
@nopdotcom Not a bad suggestion, should be an easy enough change. To take it a step further, how would you feel about additionally prefixing with `client-$protocol-Pynchon-fabulosity.p12`, so for openconnect we'd have `client-oc-Pynchon-fabulosity.p12`? Might be a bit redundant for OpenVPN files but others stand to benefit.

playbooks/roles/openconnect/templates/config.j2
```diff
@@ -33,6 +33,8 @@ ipv4-network = 192.168.1.0
 ipv4-netmask = 255.255.255.0
 ping-leases = false
 cisco-client-compat = true
+max-clients = {{ vpn_clients + 1 }}
+max-same-clients = {{ vpn_clients + 1 }}
```
cpu: What is `max-same-clients`? Can you add a comment describing this var since it might not be as self-explanatory as `max-clients`?

playbooks/roles/openconnect/templates/instructions.md.j2
```diff
@@ -1,18 +1,32 @@
...
   * [Windows](#windows)
   * [OS X](#osx)
   * [Linux](#linux)
   * [Android](#android)
   * [iOS](#ios)
 
+<a name="clientcerts"></a>
+### Server and client Certifcates (OS X, Android, iOS) ###
+
+Certificates are a mechanism by which through [public key crypytography](https://en.wikipedia.org/wiki/Public-key_cryptography) allow for secure authentication to a given resource. OS X, Andoid and iOS OpenConnect/AnyConnect support this form of authentication, and is the recommended approach to connect to your Streisand OpenConnect VPN.
```
cpu: I think this initial sentence could be simplified somewhat. I would probably skip the link to the wiki on asymmetric encryption all together and stick with:
> Client certificates are a mechanism by which clients can authenticate themselves securely with the server.

playbooks/roles/openconnect/templates/instructions.md.j2
```diff
@@ -1,18 +1,32 @@
...
+
+1. Your OpenConnect server issues its own __public certificate__. This is used by your device's client software (such as AnyConnect for iOS) to securely identify the VPN server. Download this server's certificate.
```
cpu: I would say "issues its own __server certificate__" here. All certificates are generally considered public material.

playbooks/roles/openconnect/templates/instructions.md.j2
```diff
@@ -1,18 +1,32 @@
...
+   * [ca-cert.pem](/openconnect/ca-cert.pem)
+1. Each device you wish to configure needs a __client certificate__ in addition to the public certificate above. A client certificate is used to securely identify and authenticate your device to the VPN server. Two devices can't use the same client certifcate and be logged in at the same time (one client certificate per device). Each client certificate is protected by a password, which will be needed to unlock it once you import it into your device.
```
cpu: same diff here "public certificate above" -> "server certificate above".

playbooks/roles/openconnect/templates/instructions.md.j2
```diff
@@ -41,23 +55,24 @@ OpenConnect is an extremely high-performance and lightweight VPN server that als
...
 1. Install OpenConnect using Homebrew:
 
    `brew install openconnect`
 
    * If installing Homebrew is not an option, you can also download and compile the [OpenConnect source code](/mirror/#openconnect).
-1. Download the CA and client certificate files:
-   * [ca-cert.pem](/openconnect/ca-cert.pem)
-   * [client.p12](/openconnect/client.p12)
-1. Place the downloaded files into a separate folder (e.g. `{{ streisand_server_name }}-openconnect`), open your Terminal, and `cd` to that directory.
+1. Download the [ca-cert.pem](/openconnect/ca-cert.pem) file, and a [client certificate file](#clientcerts) from the list above.
```
cpu: Instead of using the file name here (`ca-cert.pem`) I think you should use the same term as the docs (`server certificate file`). That matches the client certificate file link.

playbooks/roles/openconnect/templates/instructions.md.j2
```diff
@@ -41,23 +55,24 @@ OpenConnect is an extremely high-performance and lightweight VPN server that als
...
+1. Place the downloaded CA certificate a selected client certificate into a separate folder (e.g. `{{ streisand_server_name }}-openconnect`), open your Terminal, and `cd` to that directory.
```
cpu: Typo? "certificate a selected" -> "certificate and selected"?

playbooks/roles/openconnect/templates/instructions.md.j2
```diff
@@ -117,12 +132,14 @@ If you're affected, you can use this workaround:
...
 <a name="ios"></a>
 ### iOS ###
-1. Download and install the [client profile](/openconnect/client.p12) on your iOS device.
-   * The profile can be emailed to your device (simply tap the attachment), transferred via the [Apple Configurator](https://itunes.apple.com/us/app/apple-configurator/id434433123?mt=12) utility, or downloaded directly by viewing these instructions on the device itself and tapping [here](/openconnect/client.p12).
-   * Follow the on-screen instructions.
-   * You will be prompted to enter your device password or pin.
-   * Enter `{{ ocserv_pkcs12_password.stdout }}` when the *ENTER THE PASSWORD FOR THE CERTIFICATE...* prompt is displayed.
-   * Once the profile is installed, you are ready to continue.
+
+1. Transfer a [client certificate](#clientcerts) for each device you wish to configure:
```
cpu: "for each device" -> "to each device"

playbooks/roles/openconnect/vars/main.yml
```diff
@@ -26,14 +26,14 @@ ocserv_compilation_directory:  "/usr/local/src/ocserv-{{ ocserv_version }}"
 ocserv_source_filename:        "ocserv-{{ ocserv_version }}.tar.xz"
 ocserv_sig_filename:           "{{ ocserv_source_filename }}.{{ ocserv_sig_extension }}"
 
-mit_gpg_keyserver_address:    "x-hkp://pgp.mit.edu"
+mit_gpg_keyserver_address:     "x-hkp://pgp.mit.edu"
```
cpu: spurious diff?

playbooks/roles/validation/tasks/main.yml
```diff
@@ -8,3 +8,8 @@
   fail:
     msg: "tinyproxy cannot be enabled if ssh forward user is disabled"
   when: not streisand_ssh_forward_enabled and streisand_tinyproxy_enabled
+
+- name: Validate that the maximum number of clients is set to a reasonable amount
+  fail:
+    msg: "Unreasonable number of VPN clients sepcified"
```
cpu: I think this might be clearer as "Too many VPN clients specified. vpn_clients must be greater than 0 and less than 20".

alimakki:
@cpu, thanks for the feedback!
I've made the necessary changes based on your feedback.
There's a bit more testing I'd like to do on this, specifically with regards to the `max-same-clients`; setting it to 1 only allowed me access to only a single session. 2 is listed in the [same conf](https://ocserv.gitlab.io/www/manual.html) and seems to work.

Otherwise I hope this helps with rolling out multiple WireGuard and IKEv2 clients :)

playbooks/roles/openconnect/vars/main.yml
alimakki: Just so that it lines up with with the variable above it.

playbooks/roles/openvpn/templates/instructions.md.j2
```diff
@@ -28,8 +28,8 @@ For security reasons, OpenVPN clients typically have their own unique certificat
 1. Go to the *Compatibility* tab and click the *Run this program as an administrator* checkbox in the *Privilege Level* section.
 1. Double-click the OpenVPN GUI desktop icon to launch the application.
 1. Download one of these unified OpenVPN profiles:
-{% for client in openvpn_clients %}
-   * [{{ client }}](/openvpn/{{ client }}/{{ openvpn_direct_profile_filename }})
+{% for client in vpn_client_names.results %}
```
nopdotcom: I can table-ize these after this gets committed.

alimakki:
@cpu glad to hear it!

I wanted to point out as well that I changed the OpenConnect instructions to reference Legacy AnyConnect for iOS for the meantime due to changes in the newer version where imported certificates aren't visible to the app and a username/password is prompted.

cpu:
Obviously haven't gotten to the testing I promised :'( Apologies!

alimakki:
@cpu, no worries!

I did some additional testing and I had to bump ocserv to version 0.11.8, since the previous one simply just doesn't cooperate with multiple certificates.

I also had to restore the `max-same-clients` configuration to match `max-clients`; my conclusion during my short round of testing tells me that ocserv treats username/pass as one client, and all certificate based clients as one singular block as far as I can tell, though I would like to be proven wrong.

PR #900, opened 2017-08-25
docs: Fix undefined variable

wzyboy:
Variable `stunnel_remote_port` is only available when we have run the stunnel role. There are two references of `stunnel_remote_port` in the docs, but only one of them is surrounded by `{% if
streisand_stunnel_enabled %}{% endif %}`. This causes "AnsibleUndefinedVariable" error when we disabled stunnel role and run gateway role.

 This commit fixes the bug by surrounding the other reference with `{% if %}{% endif %}`, as well.

### DIFF
--- a/playbooks/roles/streisand-gateway/templates/firewall-fr.md.j2
+++ b/playbooks/roles/streisand-gateway/templates/firewall-fr.md.j2
@@ -42,8 +42,6 @@ Si vous installez Streisand sur un serveur existant ou un fournisseur alternatif
   * TCP - {{ shadowsocks_server_port }}
   * UDP - {{ shadowsocks_server_port }}
 {% endif %}
-* stunnel
-  * TCP - {{ stunnel_remote_port }}
 {% if streisand_tor_enabled %}
 * Tor
   * TCP - {{ tor_orport }} - Pont
--- a/playbooks/roles/streisand-gateway/templates/firewall.md.j2
+++ b/playbooks/roles/streisand-gateway/templates/firewall.md.j2
@@ -42,8 +42,6 @@ If you are installing Streisand on an existing server or alternate provider then
   * TCP - {{ shadowsocks_server_port }}
   * UDP - {{ shadowsocks_server_port }}
 {% endif %}
-* stunnel
-  * TCP - {{ stunnel_remote_port }}
 {% if streisand_tor_enabled %}
 * Tor
   * TCP - {{ tor_orport }} - Bridge


playbooks/roles/streisand-gateway/templates/firewall-fr.md.j2
```diff
@@ -42,8 +42,10 @@ Si vous installez Streisand sur un serveur existant ou un fournisseur alternatif
   * TCP - {{ shadowsocks_server_port }}
   * UDP - {{ shadowsocks_server_port }}
 {% endif %}
+{% if streisand_stunnel_enabled %}
 * stunnel
   * TCP - {{ stunnel_remote_port }}
+{% endif %}
```
alimakki: This block of code should be deleted, the stunnel port is wrapped in the OpenVPN if block (line 28-39)

playbooks/roles/streisand-gateway/templates/firewall.md.j2
```diff
@@ -42,8 +42,10 @@ If you are installing Streisand on an existing server or alternate provider then
   * TCP - {{ shadowsocks_server_port }}
   * UDP - {{ shadowsocks_server_port }}
 {% endif %}
+{% if streisand_stunnel_enabled %}
 * stunnel
   * TCP - {{ stunnel_remote_port }}
+{% endif %}
 {% if streisand_tor_enabled %}
```
alimakki: Same as above.

wzyboy:
I don't understand. I was testing a new role when I encountered this bug. I set all `streisand_XXX_enabled` to `no` in global vars to speed up the tests. But I got `AnsibleUndefinedVariable` error. In my case, neither stunnel nor openvpn role was run (I disabled both of them):

```
failed: [vagrant] (item={'key': u'en', 'value': {u'file_suffix': u'', u'tor_locale': u'en-US', u'language_name': u'English'}}) => {"failed": true, "item": {"key": "en", "value": {"file_suffix": "", "language_name": "English", "tor_locale": "en-US"}}, "msg": "AnsibleUndefinedVariable: 'stunnel_remote_port' is undefined"}   
```

alimakki:
Hi @wzyboy,

As en example, let's take a look at a part of firewall-fr.md.j2:

```
{% if streisand_openvpn_enabled %} #OpenVPN block starts here
* OpenVPN
  * UDP - 53 de/à `{{ dnsmasq_openvpn_tcp_ip }}`
  * UDP - 53 de/à `{{ dnsmasq_openvpn_udp_ip }}`
    * Dnsmasq écoute le trafic DNS sur ces adresses IP et répond aux requêtes des clients OpenVPN connectés.
  * TCP - {{ openvpn_port }}
  * UDP - {{ openvpn_port_udp }}
{% if streisand_stunnel_enabled %} # inner stunnel block here
* stunnel
  * TCP - {{ stunnel_remote_port }}
{% endif %}
{% endif %}
{% if streisand_shadowsocks_enabled %}
* Shadowsocks
  * TCP - {{ shadowsocks_server_port }}
  * UDP - {{ shadowsocks_server_port }}
{% endif %}
{% if streisand_stunnel_enabled %} # this stunnel block is redundant, it should be deleted
* stunnel
  * TCP - {{ stunnel_remote_port }}
{% endif %}
```

If you observe, stunnel is shown twice; it should be only in one place, and only if both openvpn and stunnel are set to yes.

Please let me know if that helps clarify!

wzyboy:
Hi @alimakki ,

thanks! I did not notice that stunnel cannot be installed without openvpn before I read "validation" role. Now I see.

I have replaced the commit with a new one.

alimakki:
@wzyboy Thanks for the quick turnaround!

I'll merge as soon as CI completes.

Cheers!

PR #910, opened 2017-08-29
Added failover to Nginx for SS-simple-obfs plugin. Fixes #843

### LINKED ISSUE #843: Add `--failover` to nginx for shadowsocks simple-obfs plugin
Presently Streisand configures the simple-obfs Shadowsocks plugin [without --failover](https://github.com/jlund/streisand/blob/2bd2a007dc6230685b1117527037f49c2dbe606b/playbooks/roles/shadowsocks/defaults/main.yml#L12). We should consider using the `--failover` option to route non-Shadowsocks requests to the NGINX instance that is already running. This has the advantage of better masking the purpose of the port. See: https://github.com/shadowsocks/simple-obfs/issues/81

denoza:
This PR adds the support for `failover` to Nginx web server running at port 443. It addresses the issue #843.


cpu:
Hi @denoza 

This looks good! I'd like to try running it on a few test instances before merging. I wanted to comment first to let you know that it hasn't gone unnoticed :-)

Thanks!

nickolasclarke:
@cpu do you still need testing on this? I can run this since I use OBFS daily here.

cpu:
@nickolasclarke That would be great! I have family in town and haven't been able to get away to a computer as much as I hoped.

denoza:
I have added a new commit [`74e37a7`](https://github.com/jlund/streisand/pull/910/commits/74e37a71076d56b835f138aaba07aab0bc6f135c) to this branch. Changing the obfs-parameter from `http` to `tls` is more reasonable as configured Nginx web server resides behind port `443` and uses `HTTPS`. I have been using these settings for last one month or so and getting pretty stable performance.

A quick experiment for checking `failover` gives following output:
```
$_ curl -v https://xxx
* Rebuilt URL to: https://xxx
*   Trying xxx...
* Connected to xxx (xxx) port 443 (#0)
* found 173 certificates in /etc/ssl/certs/ca-certificates.crt
* found 704 certificates in /etc/ssl/certs
* ALPN, offering http/1.1
* SSL connection using TLS1.2 / ECDHE_RSA_AES_128_GCM_SHA256
* server certificate verification failed. CAfile: /etc/ssl/certs/ca-certificates.crt CRLfile: none
* Closing connection 0
curl: (60) server certificate verification failed. CAfile: /etc/ssl/certs/ca-certificates.crt CRLfile: none
More details here: http://curl.haxx.se/docs/sslcerts.html

curl performs SSL certificate verification by default, using a "bundle"
 of Certificate Authority (CA) public keys (CA certs). If the default
 bundle file isn't adequate, you can specify an alternate file
 using the --cacert option.
If this HTTPS server uses a certificate signed by a CA represented in
 the bundle, the certificate verification probably failed due to a
 problem with the certificate (it might be expired, or the name might
 not match the domain name in the URL).
If you'd like to turn off curl's verification of the certificate, use
 the -k (or --insecure) option
```

which shows that non-SS traffic is getting routed to Nginx.

denoza:
> I did notice however that before your PR a shadowsocks-android client that did not configure the obfuscation worked fine. After applying your PR the shadowsocks-android client does not work at all unless obfuscation is configured properly. I tried this against master and your PR twice and both times was consistent.

@cpu you are right. It almost slipped my mind that initial support for obfuscation was meant as an option. I had a check with my Android client and I had the same problem. I think adding the `failover` tag in server config makes it mandatory to use obfuscation. Additionally I also checked the SS-Libev as client on my Linux machine and duplicated the issue. I will figure out if this problem is with our configuration or simply `failover` doesn't support SS traffic without OBFS. Until then it is safe to not merge.

denoza:
Seems like [failover](https://github.com/shadowsocks/simple-obfs/issues/101#issuecomment-327107559) **must** use the obfuscation. We can't leave it optional.

cpu:
@denoza Oh bummer! Thanks for checking with upstream. I didn't realize this limitation of failover :-(

I think that its probably best to keep obfuscation optional and not use failover for the time being. What do you think? I feel that there are probably a number of easy ways to fingerprint a Streisand server that would let you guess at the presence of Shadowsocks even with failover in place.

nickolasclarke:
I believe that the windows client does not currently support this naively, and its likely that many other clients do not either. I would vote to keep this optional, due to this limitation, unfortunately.

denoza:
> I think that its probably best to keep obfuscation optional and not use failover for the time being. What do you think?

I agree, as @nickolasclarke mentioned not all clients have support for obfuscation, therefore it is better not to make it enforcing.

cpu:
Ok. Regretfully I'll close this PR for now :'( Hopefully more clients will add obfuscation support in the future. I think its best to avoid prompting for "enforced obfuscation" during the shadowsocks install for now. Thanks for the hard work @denoza, I'm sorry it didn't work out but hopefully you find it useful for your own shadowsocks deployments :-)

cpu:
@hybtoy No plans.

PR #911, opened 2017-08-29
New role: Let's Encrypt

wzyboy:
This role implements Let's Encrypt support for Streisand server. It is intended to resolve https://github.com/jlund/streisand/issues/639.

The role adds two more questions during playbook run. Whether the role is run is based on user input. If user skips the questions, the role will not be run at all.

The role itself has a fail-safe. If it successfully gets the certificates, it sets `le_ok` true, which has the following consequences:
- OpenSSL self-signed certificates generation process will be skipped;
- Let's Encrypt certificates will be used in Nginx, instead of self-signed ones;
- Instructions on how to import self-signed certificates in generated_docs will be removed.

If the roles fails to get the certificates for any reason (Let's Encrypt server down, user error, etc), `le_ok` would be set false. The consequences above will not happen, the playbook continues to run as if the role does not exist at all.


nopdotcom:
The install should fail if the user has requested LE configuration, and Streisand is unable to comply. At least failure should be the default. `Fall back to self-signed certificates instead? (y/N)`

This appears to need port 80 opened inbound. Are we a) OK with that, and b) ready to doctor the security groups/firewall rule instructions?

Sadly, LE doesn’t sound like it will issue certs for stuff in amazonaws.com. https://community.letsencrypt.org/t/policy-forbids-issuing-for-name-on-amazon-ec2-domain/12692

wzyboy:
> The install should fail if the user has requested LE configuration, and Streisand is unable to comply. At least failure should be the default. Fall back to self-signed certificates instead? (y/N)

I don't see the reason to fail at this point, considering it's one of the last roles to run. Re-running the entire playbook is time consuming.

> This appears to need port 80 opened inbound. Are we a) OK with that, and b) ready to doctor the security groups/firewall rule instructions?

Valid point. Thank you. When writing the role, I always test it with "8. Existing server" but have never tried to provision a new cloud server. I'll need to open these ports up in EC2/GCE/whatever security groups as well. The problem is, `lets-encrypt.yml` (which asks questions) is included in `streisand.yml` so when creating a new cloud server, the playbook does not know about whether the user would like to set up Let's Encrypt, yet. Maybe I should asks the questions in earlier stages.

> Sadly, LE doesn’t sound like it will issue certs for stuff in amazonaws.com. https://community.letsencrypt.org/t/policy-forbids-issuing-for-name-on-amazon-ec2-domain/12692

No, of course we are not requesting certificates of Amazon's domain. Most cloud server providers --
 AFAIK like Linode, DigitalOcean and Rackspace -- do not provide a public domain for the cloud server anyway. Users are required to use their own domains to request certificates.

nopdotcom:
Oh, I’m sorry, I just skimmed your code--I assumed the user interview was going to show up first, when it would be cheap to test/abort. I can’t read Ansible!

I feel like this PR is a really good capability to have; I just don’t understand the source of my feelings yet. :‐) I think some of this is because I don’t understand where other people are going with planned Streisand lifecycle improvements.

playbooks/lets-encrypt.yml
```diff
@@ -0,0 +1,58 @@
+---
+# lets-encrypt.yml asks questions about Streisand domain name, which is used to
```
cpu: typo: "about Streisand domain name" -> "about the Streisand instance domain name" maybe?

playbooks/lets-encrypt.yml
```diff
@@ -0,0 +1,58 @@
...
+# request TLS certificates from Let's Encrypt server. If user fails to answer
```
cpu: Nit: Can you drop "server" and just say "from Let's Encrypt"?

playbooks/lets-encrypt.yml
```diff
@@ -0,0 +1,58 @@
...
+# these questions, certain roles are disabled.
```
cpu: Can you expand on which roles are disabled?

playbooks/lets-encrypt.yml
```diff
@@ -0,0 +1,58 @@
...
+- name: Collect information about the Streisand domain
+# =========================================
+  hosts: streisand-host
+  gather_facts: yes
```
cpu: Do you need `gather_facts` here? I suspect it might not be necessary

cpu: Do you need `gather_facts` here? I suspect it might not be necessary

playbooks/roles/azure-security-group/tasks/main.yml
```diff
@@ -45,6 +45,15 @@
         access: Allow
         direction: Inbound
         priority: 108
+      # Let's Encrypt
+      # ---
+      - name: "AllowLetsEncrypt"
```
cpu: I think I would prefer we call the rule "AllowHTTP" and then comment that it's for Let's Encrypt. Does that make sense to you?

playbooks/roles/ec2-security-group/tasks/main.yml
```diff
@@ -33,6 +33,12 @@
         from_port: "{{ ssh_port }}"
         to_port: "{{ ssh_port }}"
         cidr_ip: 0.0.0.0/0
+      # Let's Encrypt
```
cpu: Ditto for calling it HTTP and mentioning Let's Encrypt afterwards.

playbooks/roles/ec2-security-group/tasks/main.yml
```diff
@@ -33,6 +33,12 @@
...
+      # ---
+      - proto: tcp
+        from_port: "{{ le_port }}"
+        to_port: "{{ le_port }}"
+        cidr_ip: "{{ le_cidr }}"
```
cpu: I think this can be `0.0.0.0/0` in all cases no?

cpu: I think you can remove the `le_cidr` var and use that range directly like the other rules.

playbooks/roles/gce-network/tasks/main.yml
```diff
@@ -33,6 +33,18 @@
     credentials_file: "{{ gce_json_file_location }}"
     project_id: "{{ gce_project_id }}"
 
+- name: Open the Let's Encrypt port in the GCE firewall
```
cpu: same nit RE: calling it the HTTP rule.

playbooks/roles/lets-encrypt/tasks/main.yml
```diff
@@ -0,0 +1,58 @@
...
+- name: Perform initial configuration of acmetool
+  command: acmetool quickstart
+  args:
+    creates: "{{ le_base }}/conf/target"
+
+# acmetool should now run in redirector mode listening on 80/tcp for HTTP
+# requests (including ACME chanllenges). All HTTP requests are simply
+# redirected to their HTTPS counterparts.
+
+- name: Ensure incoming HTTP traffic is allowed
```
cpu: Can you put this in its own `firewall.yml` task file that's included in here?

playbooks/roles/lets-encrypt/tasks/main.yml
```diff
@@ -0,0 +1,58 @@
...
+  apt_key:
+    id: 9862409EF124EC763B84972FF5AC9651EDB58DFA
+    data: "{{ item }}"
+  with_file: acmetool_ppa.pem
+
+- name: Add the official acmetool repository
+  apt_repository:
+    repo: "deb http://ppa.launchpad.net/hlandau/rhea/{{ ansible_distribution|lower }} {{ ansible_lsb.codename }} main"
+
+- name: Install acmetool
```
cpu: Can you put the first three tasks (adding the apt key, adding the repository, installing the apt package) into their own `install.yml` task file that's included here? Thanks!

playbooks/roles/lets-encrypt/templates/response.yaml.j2
```diff
@@ -0,0 +1,13 @@
+"acme-enter-email": "{{ streisand_admin_email }}"
+"acme-agreement:https://letsencrypt.org/documents/LE-SA-v1.1.1-August-1-2016.pdf": true
```
cpu: I would prefer to not see this URL hardcoded. Let's Encrypt can change it at any time and new registrations will fail until this line is updated. Is it possible to prompt for it? I can think of some alternative workarounds if it will be tricky.

playbooks/roles/lets-encrypt/templates/response.yaml.j2
```diff
@@ -0,0 +1,13 @@
...
+"acmetool-quickstart-choose-server": https://acme-v01.api.letsencrypt.org/directory
+"acmetool-quickstart-choose-method": redirector
+# This is only used if "acmetool-quickstart-choose-method" is "webroot".
+"acmetool-quickstart-webroot-path": "/var/www/foo/bar/.well-known/acme-challenge"
```
cpu: If we don't use this mode can we remove this line? It seems strange to have a `/var/www/foo/bar` config line if the folder doesn't exist.

playbooks/roles/lets-encrypt/templates/response.yaml.j2
```diff
@@ -0,0 +1,13 @@
...
+"acmetool-quickstart-complete": true
+"acmetool-quickstart-install-cronjob": true
+"acmetool-quickstart-install-haproxy-script": true
```
cpu: I think this can be false - we don't install `haproxy`.

playbooks/roles/lets-encrypt/templates/response.yaml.j2
```diff
@@ -0,0 +1,13 @@
...
+"acmetool-quickstart-install-redirector-systemd": true
+"acmetool-quickstart-key-type": ecdsa
+"acmetool-quickstart-rsa-key-size": 4096
```
cpu: I would prefer to just use `2048` here.

playbooks/roles/lets-encrypt/vars/main.yml
```diff
@@ -0,0 +1,16 @@
+---
+le_base: /var/lib/acme
+le_port: 80
+# Currently Let's Encrypt does not provide a range of IP addresses they are
```
cpu: Let's Encrypt won't provide a range of IP addresses as a matter of policy. I think you can delete this entire comment :-)

playbooks/roles/streisand-gateway/templates/firewall-fr.md.j2
```diff
@@ -20,6 +20,10 @@ Si vous installez Streisand sur un serveur existant ou un fournisseur alternatif
 {% endif %}
 * Nginx (Passerelle Streisand)
   * TCP - {{ nginx_port }}
+{% if streisand_le_enabled %}
+* Let's Encrypt
```
cpu: HTTP for Let's Encrypt

playbooks/roles/streisand-gateway/templates/firewall.md.j2
```diff
@@ -20,6 +20,10 @@ If you are installing Streisand on an existing server or alternate provider then
 {% endif %}
 * Nginx (Streisand Gateway)
   * TCP - {{ nginx_port }}
+{% if streisand_le_enabled %}
+* Let's Encrypt
```
cpu: HTTP for Let's Encrypt

tests/tests.sh
```diff
@@ -32,7 +32,7 @@ function dev_setup {
 }
 
 function run_tests {
-  run_playbook "$DIR/run.yml" -e streisand_ci=true "$1"
+  run_playbook "$DIR/run.yml" --extra-vars=@"$DIR/vars_ci.yml" "$1"
```
cpu: Can you explain this diff?

tests/group_vars/all/all
```diff
@@ -1,5 +1,4 @@
 ---
-streisand_ci: yes
```
cpu: Why is this being removed?

tests/vars_ci.yml
```diff
@@ -0,0 +1,5 @@
+streisand_ci: yes
+
+# Answer questions in playbooks/lets-encrypt.yml
```
cpu: I think it would be easier to check for the `streisand_noninteractive` var that we already use to skip prompting during CI/automated runs.

playbooks/lets-encrypt.yml
```diff
@@ -0,0 +1,58 @@
...
+        of your Let's Encrypt account.
+
+        Please type your email below. Press enter to use hostmaster email of Streisand domain.
+      default: "hostmaster@{{ streisand_domain_var }}"
+      private: no
+
+  pre_tasks:
+    - name: Set Streisand domain
+      set_fact:
+        streisand_domain: "{{ streisand_domain_var }}"
```
cpu: I would love to see some basic checks here that the domain resolves to an IP (and maybe even that it's == the streisand ipv4 address we know the server is using). I think that will cut down on the most obvious failures right out of the box.

playbooks/roles/lets-encrypt/tasks/main.yml
```diff
@@ -0,0 +1,58 @@
...
+        creates: "{{ le_certificate }}"
+
+    - set_fact:
+        le_ok: True
+
+  rescue:
+    - set_fact:
+        le_ok: False
+
+    - name: Failed to get Let's Encrypt certificates. We are falling back to use self-signed ones. You could re-run the playbook to try again.
```
cpu: I would remove the "You could re-run the playbook to try again". That's likely to lead to the user tripping one of the Let's Encrypt rate limits (in particular the # of failed validations per hour rate limit).

I think it might make sense to ask the user if they want to abort here. If you're intending to use Let's Encrypt and it fails you may want to start over instead of getting a server with self-signed certificates.

cpu:
I think there are also some CI failures to address. The build is red on this branch presently.

playbooks/lets-encrypt.yml
wzyboy: @cpu May I ask if you are worrying about reaching Let's Encrypt's rate limit? According to acmetool's [user guide](https://hlandau.github.io/acme/userguide#challenge-completion-philosophy), self-tests are already done for HTTP-based validation before sending requests to Let's Encrypt for real.

playbooks/roles/lets-encrypt/tasks/main.yml
wzyboy: @cpu I am worried about user experience. Other non-critical roles in the playbook, like streisand-mirror, can fail without aborting the entire playbook. Maybe only a warning would be enough? Anyway, self-signed certificates are still secure if user checks for fingerprints carefully...

playbooks/roles/lets-encrypt/templates/response.yaml.j2
wzyboy: @cpu Oh yes. I'll fix that.

playbooks/roles/lets-encrypt/templates/response.yaml.j2
wzyboy: @cpu I'll remove that line.

playbooks/roles/lets-encrypt/templates/response.yaml.j2
wzyboy: @cpu Hmm this value is not actually used, as we are issuing ECC certificates anyway. I'll remove this line.

playbooks/roles/lets-encrypt/templates/response.yaml.j2
wzyboy: @cpu I took the response file from acmetool's repo: https://github.com/hlandau/acme/blob/master/_doc/response-file.yaml .

I'm also a little uncomfortable on this, but I cannot think of a better way :-(

The agreement URL is subject to change, yes. But it's not like we can predict the future URL. So the only hope is that we watch out for announcements from Let's Encrypt and update the URL here when they change it.

tests/tests.sh
wzyboy: @cpu I would like to pass more vars to `run.yml` and enumerating them here looks ugly. So I put all (three) vars in a file and pass the file.

tests/group_vars/all/all
wzyboy: @cpu Because it's already passed in `vars_ci.yml`. Vars passed by `-e / --extra-vars` have the highest priority: http://docs.ansible.com/ansible/latest/playbooks_variables.html#variable-precedence-where-should-i-put-a-variable

tests/vars_ci.yml
wzyboy: @cpu That's the tricky part. In Ansible, `vars_prompt` has never supported `when:` statement. So we cannot skip a prompt based on the value of a variable. The only way to prevent Ansible from asking questions is to define the variable in advance.

I found this out the hard way:

```
$ cat a.yaml 
- hosts: localhost
  connection: local
  gather_facts: no

- include: b.yaml
  when: bbb is not defined

$ cat b.yaml 
- hosts: localhost
  connection: local
  gather_facts: no

  vars_prompt:
    - name: "var_1"
      prompt: "111"

  tasks:
    - debug:
        var: var_1

$ ansible-playbook a.yaml -e bbb=defined

PLAY [localhost] ***************************************************************
111: 

```

wzyboy:
@cpu Thanks for reviewing. I have fixed most of the issues you commented. The rest needs some discussion before committing.

As of CI failures, I think there is something wrong with the CI itself. It always stuck at:

```
PLAY [Prepare the new server for Ansible] **************************************
TASK [Install Python using a raw SSH command to enable the execution of Ansible modules] ***
changed: [10.205.188.102]
PLAY [Collect information about the Streisand domain] **************************
TASK [Gathering Facts] *********************************************************
fatal: [10.205.188.102]: UNREACHABLE! => {"changed": false, "msg": "SSH Error: data could not be sent to remote host \"10.205.188.102\". Make sure this host can be reached over ssh", "unreachable": true}
	to retry, use: --limit @/home/travis/build/jlund/streisand/tests/run.retry
10.205.188.102             : ok=1    changed=1    unreachable=1    failed=0   
streisand                  : ok=10   changed=6    unreachable=0    failed=0   
```

It looks like something wrong with the LXC container.

tests/vars_ci.yml
wzyboy: A related post: https://stackoverflow.com/a/41429660/1324308

wzyboy:
I have no idea why CI fails. I tried with `-vvv` verbosity (with a [temporary commit](https://github.com/jlund/streisand/pull/911/commits/a11cb05cb99a9ba1ca289004e6a7791da4835610)) but I still cannot see why it fails: https://travis-ci.org/jlund/streisand/builds/270295313

tests/vars_ci.yml
nopdotcom: We might put the LE interview question (or questions) in the `streisand` script. It sounds like it’s hard to do the user-friendly thing if it’s in a role.

wzyboy: What do you think @cpu? Are there some pitfalls I need to take care of if I move the questions from an Ansible role to shell input?

wzyboy: @nopdotcom The user experience of answering questions with `vars_prompt` is fine. The problem is that the CI system cannot answer questions interactively. So we have to answer these questions in advance, if we are running in a CI environment.

playbooks/lets-encrypt.yml
cpu: A little bit about rate limits and a little bit about user experience. It's always better to find out something is broken sooner than later.

> According to acmetool's user guide, self-tests are already done for HTTP-based validation before sending requests to Let's Encrypt for real.

Ok, let's call that sufficient for now. :-)

playbooks/roles/lets-encrypt/tasks/main.yml
cpu: Hmm. I understand what you're saying. Let's leave this as-is for now and we can revisit if it becomes something people are confused/surprised by. Sound good?

tests/tests.sh
cpu: I think you can just set these in `test/group_vars/all/all`. We separate the vars for CI so that it's easier to add things to satisfy test prompts.

tests/group_vars/all/all
cpu: If you added the vars from `vars_ci.yml` here could you leave this as-is? Does that work or is there something about the priority adjustment that your PR requires?

tests/vars_ci.yml
cpu: > In Ansible, vars_prompt has never supported when: statement.

Ahh! That's super annoying/surprising!!! I understand now. Thanks

> We might put the LE interview question (or questions) in the streisand script. It sounds like it’s hard to do the user-friendly thing if it’s in a role.

@nopdotcom I'm somewhat wary of doing too much in the wrapper script if it can be avoided. I think we probably need to figure out some guidelines for what should go where. Without care we'll end up reinventing much of Ansible in very brittle shell script :-) This might be one case where it makes sense but for now to keep things simple I think we should try the Ansible prompt approach.

cpu:
> It looks like something wrong with the LXC container. It's weird because the "Install python using..." task is successful.

Hmm. That's possible but I still think it's somehow related to a diff from this PR. [master currently passes CI](https://travis-ci.org/jlund/streisand/builds/271198230). 

I'm afraid I have family visiting for the next couple of days and may not have time to help dig into the root cause that CI is failing.

tests/tests.sh
wzyboy: @cpu No problems. I have moved CI variables from extra- to group_vars.

wzyboy: @cpu Oh no. My memory tricked me. I remebered that why I move the CI variables from group_vars to extra vars: Defining them in group_vars does not provide enough priority so Ansible will still asks questions: https://travis-ci.org/jlund/streisand/builds/271395890

I'll delete the commit.

tests/group_vars/all/all
wzyboy: @cpu This could be left here as-is. It's just a habit to clean up duplicate definitions.

wzyboy:
@cpu Good news: the CI build is passing now. It turns out disabling `gather_facts` does the work. I do not know why though.

playbooks/lets-encrypt.yml
wzyboy: @cpu Ah I missed this comment. This is why the CI build fails.

playbooks/lets-encrypt.yml
```diff
@@ -0,0 +1,65 @@
...
+        Streisand server, the installation scripts can request Let's Encrypt
+        HTTPS certificates for you automatically.  If you do not provide one or
+        the request fails, self-signed certificates will be used instead.
+
+        If you have just created a new cloud server in previous steps, it is
+        good time to point your domain to your server's public address, as DNS
+        changes take time to propagate.
+
+        In order to use Let's Encrypt, you must agree to its agreements at
+        {{ le_agreements }}.
```
cpu: This templating won't work in a `prompt` and ends up displaying the literal `{{ le_agreement }}`.

playbooks/lets-encrypt.yml
```diff
@@ -0,0 +1,65 @@
...
+
+    - name: "streisand_admin_email_var"
+      prompt: |
+        Which email address do you want to use as contact of Streisand server?
+
+        This is an optioanl question. The email is only used as contact method
+        of your Let's Encrypt account.
+
+        Please type your email below. Press enter to use hostmaster email of Streisand domain.
+      default: "hostmaster@{{ streisand_domain_var }}"
```
cpu: Unfortunately this `default` templating doesn't work. It sets the default to the literal `"hostmaster@{{ streisand_domain_var}}"`.

playbooks/lets-encrypt.yml
wzyboy: @cpu Ah the fact that `vars_prompt` does not support templating is just as annoying as the fact that `vars_prompt` does not support `when` statement: https://github.com/ansible/ansible/issues/13263

I have removed this line.

playbooks/lets-encrypt.yml
wzyboy: @cpu Same as above. `le_ok` is now only set if **both** domain **and** email are given.

wzyboy:
@cpu 

> The streisand-gateway-url still uses the IPv4 address when there is a LE cert which makes you encounter a hostname mismatch error if you click the link from the instructions. I think this var should use the cert hostname if its available.

Good catch! Fixed.

> I think we can also skip all of the initial CA import instructions when the server is using a LE cert. I'm not 100% on this since some of the other instructions might assume you've already done this for cases where the self-signed CA is still needed (OpenVPN? Ipsec?)

Sorry I do not understand. In generated_docs, all instructions about how to import self-signed certificates are already hidden if `le_ok` is true.

cpu:
> Good catch! Fixed.

That was fast! :trophy: :racehorse: Thank you!

> Sorry I do not understand. In generated_docs, all instructions about how to import self-signed certificates are already hidden if le_ok is true.

Hmm! That's odd. I'm pretty sure the instructions were still included on the server I tested with. Unfortunately I just recycled it so I can't check. Maybe I'm crazy :laughing: 

I will run another installation shortly (using the LE staging server since I'll be creeping up on the duplicate certificate limit soon).

playbooks/roles/streisand-gateway/templates/instructions.md.j2
```diff
@@ -4,9 +4,16 @@
...
+Your Streisand Gateway contains step-by-step instructions for the services it provides, and mirrors of all necessary client software.
+
+{% if le_ok %}
+This document will show you how to connect to the Gateway to learn how to configure clients to use the Streisand services.
+{% else %}
+This document covers the SSL Certificate installation instructions and will show you how to connect to the Gateway to learn how to configure clients to use the Streisand services.
+{% endif %}
 
 ---
+{% if not le_ok %}
```
wzyboy: @cpu Hmm. As you could see, instructions about self-signed certificates are only rendered if `le_ok` is not true.

cpu: You're right! I think I was looking at the wrong generated-docs file. Apologies. A second run with d4ca89a produced a generated docs file with no CA import instructions and the correct HTTPS URL :+1:

playbooks/roles/lets-encrypt/tasks/main.yml
```diff
@@ -0,0 +1,42 @@
...
+      command: "acmetool want {{ streisand_domain }}"
+      args:
+        creates: "{{ le_certificate }}"
+
+    - set_fact:
+        le_ok: True
+
+  rescue:
+    - set_fact:
+        le_ok: False
```
wzyboy: @cpu How did you skip the LE setup? The `le_ok` variable should always be set, as you could see here...

wzyboy:
@cpu I see. My bad. :-( `le_ok` should be set false if `streisand_le_enabled` is false. Fixed that.

wzyboy:
@cpu Yet a another commit to fix that. :-(

playbooks/roles/lets-encrypt/templates/response.yaml.j2
```diff
@@ -0,0 +1,10 @@
+"acme-enter-email": "{{ streisand_admin_email }}"
+"acme-agreement:{{ le_agreements }}": true
+"acmetool-quickstart-choose-server": https://acme-v01.api.letsencrypt.org/directory
```
cpu: Can you make the value of the `acmetool-quickstart-choose-server` parameter a variable set in `playbooks/roles/lets-encrypt/vars/main.yml`? I find myself switching it to `https://acme-staging.api.letsencrypt.org/directory` before running the playbook for testing purposes and it would be nice to not have to edit the template file to do that.

cpu:
> Yet a another commit to fix that. :-(
@wzyboy Good news! With 6d32937 applied both my LE cert test run and my non-LE cert test run worked :tada:

alimakki:
Tested on fresh Linode instance with a default run, no regressions. LGTM 👍

playbooks/roles/lets-encrypt/templates/response.yaml.j2
wzyboy: @cpu No problem. It's a role variable now :-)

wzyboy:
@cpu Thanks for reviewing and testing work! I should have tested the situation when LE is skipped when writing the role. I'll be more careful nex timte.

cpu:
>  Thanks for reviewing and testing work! I should have tested the situation when LE is skipped when writing the role. I'll be more careful next time.

No problem at all! That's why we test :-)

nopdotcom:
I did a localhost deploy on Vultr, and it works fine that way.

jlund:
This is fantastic. I just successfully tested this on a few different providers. Thanks for making this happen, @wzyboy!

cpu:
Thanks again @wzyboy :tada: :tada: :tada:

playbooks/roles/lets-encrypt/templates/response.yaml.j2
wzyboy: The URL changed 🤦‍♂️  https://github.com/StreisandEffect/streisand/pull/1053

fghorow:
A question: how are the Let's Encrypt certs supposed to get renewed?

wzyboy:
@fghorow acmetool installs a cron job under `/etc/cron.d`, which checks the certificates every day. If it finds out the the certificate is about to expire, it will renew the ceritifcate. After the renewal, hook scripts (`/usr/lib/acme/hooks`) would be run, in order to reload common web servers.

PR #916, opened 2017-09-03
Validation refactor

### LINKED ISSUE #913: Perform validation before spinning up a new instance
Streisand currently implements a validation role to perform some simple sanity checks, however they are done post VM instance creation.

This poses a problem as if a user selects an invalid option (e.x. `streisand_openvpn_enabled = no` && `streisand_stunnel_enabled = yes`, the execution will halt leaving the user with a freshly spun server and nothing in it (assuming the user has selected one of supported VPS providers).

The validation role should be refactored to: 
 - execute prior to a VM instance creation and fail early if it must
 - execute in a localhost
 - continue to execute in CI

alimakki:
- added validation task to execute from Streisand script prior to installer running
- modified vagrant test run to include validation role
- modified CI task to include validation role
- ensure that at least one VPN is selected, otherwise we have a gateway that feels empty inside
- sorted modular variables in ascending order
- bumped yaml lint max characters to 280, reduces CI noise

Resolves #913

### DIFF
--- a/playbooks/roles/validation/tasks/main.yml
+++ b/playbooks/roles/validation/tasks/main.yml
@@ -13,3 +13,13 @@
   fail:
     msg: "Too many VPN clients specified. vpn_clients must be greater than 0 and less than 20"
   when: (vpn_clients > 20) or (vpn_clients < 1)
+
+- name: Validate that at least one VPN is specified
+  fail:
+    msg: "At least one VPN service must be installed :("
+  when: (not streisand_l2tp_enabled)
+    and (not streisand_openconnect_enabled)
+    and (not streisand_openvpn_enabled)
+    and (not streisand_ssh_forward_enabled)
+    and (not streisand_tor_enabled)
+    and (not streisand_wireguard_enabled)
--- a/playbooks/roles/validation/tasks/main.yml
+++ b/playbooks/roles/validation/tasks/main.yml
@@ -16,7 +16,7 @@
 
 - name: Validate that at least one VPN is specified
   fail:
-    msg: "At least one VPN service must be installed :("
+    msg: "At least one Streisand VPN service must be enabled."
   when: (not streisand_l2tp_enabled)
     and (not streisand_openconnect_enabled)
     and (not streisand_openvpn_enabled)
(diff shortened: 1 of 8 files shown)


playbooks/roles/validation/tasks/main.yml
```diff
@@ -13,3 +13,13 @@
   fail:
     msg: "Too many VPN clients specified. vpn_clients must be greater than 0 and less than 20"
   when: (vpn_clients > 20) or (vpn_clients < 1)
+
+- name: Validate that at least one VPN is specified
+  fail:
+    msg: "At least one VPN service must be installed :("
```
cpu: I think this `msg` might cause some confusion. I could imagine a confused user thinking they needed to install one of these VPN services on the Streisand builder machine.

How about: "At least one Streisand VPN service must be enabled".

playbooks/validate.yml
```diff
@@ -0,0 +1,14 @@
+---
+- name: Perform global variables sanity check
```
cpu: Where possible I think it would be nice to avoid terms with too much mental health stigma. Can we say "Perform global variables validation"? I think that's a better fit since this is `validate.yml`. Thoughts?

streisand
```diff
@@ -119,9 +119,25 @@ EOF
   echo; echo; ansible-playbook -i inventories/inventory-existing --extra-vars=@global_vars/vars.yml playbooks/existing-server.yml
 }
 
+# be_careful_friend asks you to pay attention because you're about to do
+# something that might be impossible to undo!
+function be_careful_friend() {
```
cpu: I think you accidentally duplicated `be_careful_friend`. This seems like the same definition as on L61..67 above

streisand
```diff
@@ -119,9 +119,25 @@ EOF
...
+  read -r -p "$1" confirm
+  case "$confirm" in
+    streisand) echo; echo "Confirmed. Continuing";;
+    *) echo; echo "Cancelling & exiting."; exit 1;;
+  esac
+}
+
+# runs the validation role to sanity check global_vars/vars.yml
```
cpu: same nit as before RE: sanity check. how about: "runs the validation role to validate global_vars/vars.yml"

alimakki:
@cpu, thanks for the feedback, I'll have the changes pushed later tonight!

alimakki:
@cpu, ready for another go.

PR #918, opened 2017-09-03
sysctl refactor

### LINKED ISSUE #908: Refactor `sysctl` tasks to common role
the l2tp role and the ip-forwarding role both set sysctl values. I think there are some other places that do as well. We should make an effort to centralize this where possible.

alimakki:
- refactor l2tp_ipsec sysctl settings into a separate role, is now applied when the common role executes
- added `fs.suid_dumpable = 0` to restrict core dumps

Resolves #908


playbooks/roles/sysctl/tasks/main.yml
```diff
@@ -0,0 +1,15 @@
+---
+- block:
+    - name: Harden the kernel
```
nopdotcom: Maybe this should say “Get the combined sysctl values”, if I’m reading this right?

alimakki: That sounds reasonable, I can push a change.

alimakki: Even better, removed entirely.

playbooks/roles/common/tasks/main.yml
```diff
@@ -62,3 +62,7 @@
   with_items:
     - { src: "20auto-upgrades.j2",       dest: "/etc/apt/apt.conf.d/20autoupgrades" }
     - { src: "50unattended-upgrades.j2", dest: "/etc/apt/apt.conf.d/50unattended-upgrades" }
+
+- name: Apply the sysctl values
```
cpu: Nit: Maybe the `name` could be a little clearer as "Apply custom sysctl values"?

playbooks/roles/sysctl/tasks/main.yml
```diff
@@ -0,0 +1,12 @@
+---
+- block:
+    - name: Apply the sysctl values
```
cpu: Similar nit: "Apply custom sysctl values"?

playbooks/roles/common/tasks/main.yml
alimakki: Ended up removing this since it happens in the role anyway

playbooks/roles/sysctl/tasks/main.yml
alimakki: Done!

alimakki:
@cpu Thanks fam :)

PR #921, opened 2017-09-05
Correctly limit OpenConnect clients

alimakki:
- openconnect client certificates uses the generated VPN client names as CN rather than an IP address, allowing ocserv to distinguish between users
 - set the max-same-clients to 1

Tested on a DO droplet with multiple simultaneous users, no issues.

Edit: typo


cpu:
Thanks @alimakki !

PR #926, opened 2017-09-07
Collect diagnostic info locally & suggest sharing in new issues

cpu:
This commit adds a diagnostics role that is used to collect information
about the Ansible installation & the Streisand clone being run. This
information is written to `streisand-diagnostics.md` in the project
directory but **not shared anywhere**. Care was taken to not include
anything that is especially sensitive (e.g. IP addresses, BIOS info).

The information written to the `streisand-diagnostics.md` file includes:
* Ansible version
* Host OS/version
* Python version
* Streisand git revision
* Whether there are untracked changes in the git clone (e.g. you've
  mucked about!)
* The genesis role that was being used (Linode, EC2, Localhost, etc)
* Which optional roles were enabled
* How many VPN clients were being configured

This commit updates the Github issue template to indicate that users
opening new issues about Streisand should share the
`streisand-diagnostics.md` file contents. This should make
troubleshooting much easier for everyone!

To be 1000% clear: no information is collected and sent to anyone
without the user doing it themselves as part of a Github issue.

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -3,9 +3,6 @@ Do you have a broad question? Are you suggesting a feature, e.g. a new VPN
 service or cloud provider? If so, please create a new issue in the Streisand
 Discussions repository instead:
 https://github.com/jlund/streisand-discussions/issues/ 
-
-If you believe you have encountered a bug, please fill out the following
-template and create an issue on this repo
 -->
 
 ### Expected behavior:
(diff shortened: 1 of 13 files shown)


alimakki:
@cpu, no problems at all running on macOS (sierra).

Gave it a trial run on Linode, attached is a generated diagnostics file.

Minor nitpick, would it be better to have the streisand-diagnostics.md put into the generated-docs folder? This would avoid accidental commits, if not, perhaps a gitignore rule to exclude it.

Otherwise it all looks kosher to me.

[streisand-diagnostics.md.zip](https://github.com/jlund/streisand/files/1284892/streisand-diagnostics.md.zip)

cpu:
@alimakki Great! I feel pretty safe merging this based on our combined tests. I originally was going to put it into the `generated-docs` folder but the var defining that location isn't in scope as early as the diagnostics run is being performed. I think having it in the `.gitignore` and putting it in the project root will make it more visible and hopefully result in more people providing it when they open issues (that's my theory anyway!).

nickolasclarke:
we should update the issue template to highlight it's existence. I'll try
to remember to do that tomorrow.

cpu:
@nickolasclarke I did that as part of this PR :-)

nickolasclarke:
i guess if I read the tile fully, I'd know that. :P

PR #929, opened 2017-09-07
Increase OpenVPN RSA key size

alimakki:
- 2048 -> 4096


cpu:
@alimakki Is there a motivation for this you had in mind? I worry that this change will likely cause additional server overhead without really meaningfully increasing the security.

alimakki:
Paranoia mostly, but for other roles that do involve SSL/TLS certs, we're already doing > 2048 bits.

cpu:
> Paranoia mostly, but for other roles that do involve SSL/TLS certs, we're already doing > 2048 bits.

Ah! That's a good argument. I'm fine with that for consistency's sake. :+1:

Issue #931, opened 2017-09-09
Amazon Lightsail existing server install fails diagnostics role "AnsibleUndefinedVariable"


mstarr68:
### Expected behavior:
To install without error. Found ubuntu user parameter being ignored. Edited ./streisand script to default to ubuntu instead of root for installer to work
Once triggered for exisiting server, enter IP, connects successfully and installs
(Would rather have native support for Amazon Lightsail instance instead of normal EC2 instance)

### Actual Behavior:
After entering IP, and after a few checks, it gets to "produce the diagnostics markdown file to share if there is an error" and errors out with an error in ansible: "AnsibleUndefinedVariable: 'dict object' has no attribute 'ansible_architecture'
Can not resolve or get passed this point.
I tried several times, in several instances both Ubuntu and Debian. Also inside a linux machine to connect remotely, as well as inside server doing localhost or existing pointing back to itself. Nothing will install on this lightsail instance.

### Steps to Reproduce:
1.  start Amazon Lightsail account, create instance, Ubuntu 16,04. Apt-get update and upgrade. create SSH keygen and copy to server, add to authorized_keys. Make sure you can connect to SSH using said new keys (i find an initial ssh session required and prompted to accept needed before a successful test can connect using troubleshooting steps). Once ssh can connect without error, trigger modified ./streisand script so it connects and runs as ubuntu user to your IP. Enter in IP and wait for error.

[ contents of `streisand-diagnostics.md` here ]

nopdotcom:
Amazon Lightsail is pretty weird right now. The easiest way I’ve found to provision it:
1. Create an Ubuntu instance
1. Upgrade it and install dependencies
1. Create a static IP; associate it with the machine
1. Reboot, just in case
1. Run `streisand`, and deploy to localhost
1. Go edit a bunch of tedious firewall rules :‐)

You may find it useful to install all the dependencies at once. I have a [cloud-init script](https://gist.github.com/nopdotcom/f7f148ce57f2728b9747e8b5ad9066c5) that gets all that right. Add it to the instance when you’re creating it:

![capto_capture 2017-09-08_09-06-38_pm](https://user-images.githubusercontent.com/1357866/30235777-ccb43bcc-94da-11e7-9af0-471bd8ccfcfa.png)

cpu:
@mstarr68 Can you share your Ansible version? `ansible --version`

mstarr68:
[...]

PR #933, opened 2017-09-10
Use defaults when accessing hostvars for diagnostics.

cpu:
This commit updates the `hostvar`'s referenced in the
`streisand-diagnostics md.j2` template to assume nothing about which
vars are/aren't present on the system. For mysterious reasons it seems
like sometimes the same version of Ansible will have different hostvars
populated. Since we intend for this diagnostics information to be useful
in cases when things are already going wrong we should take the most
conservative approach possible in collecting the information.

This commit updates the template to use the `.get()` style of accessing
Python dictionary keys, providing a default value in the event the key
isn't present. This should allow for the template to render even when
none of the referenced Ansible hostvars are available.

Updates #931

### DIFF
--- a/playbooks/roles/diagnostics/templates/streisand-diagnostics.md.j2
+++ b/playbooks/roles/diagnostics/templates/streisand-diagnostics.md.j2
@@ -8,12 +8,12 @@ It will help the developers reproduce your problem and provide a fix.
 
 ### Ansible Information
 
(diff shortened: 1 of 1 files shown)


cpu:
I'm not sure this actually works the way I intended. _sigh_. Going to close and experiment more.

PR #936, opened 2017-09-10
Allow customizing Streisand services at install time.

cpu:
This commit updates Streisand to allow enabling/disabling the core
services at install time, or by customizing a site specific config file
before running Streisand.

Ansible's prompt functionality is unfortunately crippled with respect to
templating and `when` conditions. This makes it tricky to accomplish the
customization we want where you are only prompted for each service if
you want to customize the overall installation.

To allowing changing services on a per-site basis & remembering the
decision the wrapper script now creates a `$HOME/.streisand` directory
with a `$HOME/.streisand/site.yml` site specific config file. By
default it is populated with the `global_vars/default-site.yml` file
from the Streisand repository.

The `streisand` wrapper conditionally invokes a separate playbook
(`playbooks/customize.yml`) for customization that rewrites the
`$HOME/.streisand/site.yml` vars file based on what the user specifies.
It's a little bit ugly but It Works(!). Further refinement welcome!

This methodology also supports customizing the installed services
non-interactively by editing `$HOME/.streisand/site.yml` ahead of
running `./streisand` and skipping the customization step. This is
useful if (for example) you only ever want to install Wireguard on your
Streisand instances. You can create a `$HOME/.streisand/site.yml` that
only enables Wireguard and all of your Streisand instances will be
provisioned accordingly.

The validation role is run after customization to ensure that the
choices are valid and don't result in (for e.g.) no services enabled.

Presently both Travis and the Vagrant `streisand-host` ignore the
`$HOME/.streisand.yml` and use the `global_vars/default-site.yml` vars
resulting in a provision with all services enabled.

Updates https://github.com/jlund/streisand/issues/746


alimakki:
👍

playbooks/customize.yml
```diff
@@ -0,0 +1,88 @@
...
+    - name: streisand_tinyproxy_enabled
+      prompt: "Enable tinyproxy? Press enter for default "
+      default: "yes"
+      private: no
+    - name: streisand_tor_enabled
+      prompt: "Enable Tor? Press enter for default "
+      default: "yes"
+      private: no
+    - name: streisand_wireguard_enabled
+      prompt: "Enable Wireguard? Press enter for default "
```
jlund: s/Wireguard/WireGuard/

streisand
```diff
@@ -116,18 +177,13 @@ EOF
   # Create the inventory file
   echo "$TEMPL" > inventories/inventory-existing
   # Invoke the Streisand playbook on the existing server inventory
-  echo; echo; ansible-playbook -i inventories/inventory-existing --extra-vars=@global_vars/vars.yml playbooks/existing-server.yml
-}
-
-# runs the validation role to validate global_vars/vars.yml
-function validate() {
-  echo; echo; ansible-playbook playbooks/validate.yml
+  run_genesis existing-server.yml inventories/intenvory-existing
```
jlund: intenvory -> inventory

jlund: After fixing the inventory filename, I'm also seeing the following error when choosing Option 8 and running against an existing server:

```
fatal: [localhost]: FAILED! => {"changed": false, "failed": true, "msg": "AnsibleUndefinedVariable: 'streisand_genesis_role' is undefined"}
```

streisand
```diff
@@ -45,15 +61,60 @@ Please create this keypair by running 'ssh-keygen'"
...
+function validate() {
+  echo; echo; ansible-playbook \
+    --extra-vars="@$GLOBAL_VARS" \
+    --extra-vars="@$DEFAULT_SITE_VARS" \
+    --extra-vars="@$SITE_VARS" \
+    playbooks/validate.yml
+}
+
+# customize prompts the user to decide if they want to customize the Streisand
+# installation. If the user chooses to the playbooks/customize.yml role is used to
```
jlund: I think there should be a comma after "If the user chooses to".

streisand
jlund: That second error was unrelated to this commit and is also happening on master. I just submitted PR #955 as a proposed fix.

cpu:
@jlund Feedback addressed. Thanks!

describe19:
How does one customize the streisand services? For example I only want to install OpenVPN + stunnel... does this commit make it possible?

alimakki:
@describe19, you can do this now; before running the streisand script you can edit this [file](https://github.com/jlund/streisand/blob/master/global_vars/vars.yml), and set the services you don’t want to ‘no’

nopdotcom:
A feature for different PR: `./streisand-config` to drop you into an editor.

PR #937, opened 2017-09-11
Specify using iOS apps.

sandinmyjoints:
My experience was that auto-install did not happen when using Chrome for iOS or Google Inbox, only when using the Apple apps.


cpu:
@sandinmyjoints Thanks!

@nopdotcom @alimakki Thoughts on this PR? I think you two have the best handle on iOS client setup :-)

alimakki:
I'm always in favor of more clarity. Don't see any reason to say no to this PR :)

PR #945, opened 2017-09-15
Update shadowsocks-libev to 3.1.0, simple-obfs to tip of master.

cpu:
This commit updates the git source of `shadowsocks-libev` that Streisand
clones to version 3.1.0. This release removes udns support so we can
drop `libundns-dev` from `shadowsocks_dependencies`. Shadowsocks-libev
now uses `libc-ares-dev` so we add that dep in its place.

Unfortunately the latest tagged release of simple-obfs is 0.0.3 and
still uses `libudns-dev`. This commit switches to cloning the tip of
master while waiting for a release.

I tested a Shadowsocks Android client against a fresh server both with
and without simple-obfs and both worked correctly using a 3.1.0 server.


cpu:
Thanks @jlund

PR #946, opened 2017-09-16
Launch L2TP/IPsec services via systemd

z0rc:
This is first step on moving all service management and monitoring under systemd.

`libreswan` install proper `ipsec.service` which includes safety checks and automatic restart of failure. `xl2tpd` service file which comes from ubuntu repo, have explicit `Restart=no`. I don't know history of this yet, but will investigate it, but this behavior can be easily overridden with drop in config.


z0rc:
Oh, right, I forgot to sync my fork with upstream. Closing this one for now.

PR #947, opened 2017-09-16
Refactor l2tp-ipsec services launch

z0rc:
This is first step on sanitizing streisand's service management.

This PR enables separate ipsec and xl2tpd services, iptables rules are applied via oneshot service too.

I deliberately switched from `service` module to `systemd` module, as latter allows to do `systemctl daemon-reload` with single attribute. This command is required to make systemd know about newly installed service units.


z0rc:
I'm aware of test failure and investigating.

cpu:
Thanks for the PR @z0rc - let us know if we can help! I will endeavour to review it this weekend.

z0rc:
Test is green, ready for review.

playbooks/roles/l2tp-ipsec/templates/streisand-l2tp-iptables.service.j2
```diff
@@ -0,0 +1,14 @@
+[Unit]
+Description=Set iptables rules required to L2TP/IPsec gateway
```
cpu: Typo: "Set iptables rules required _to_ L2TP/IPSec gateway" -> "Set iptables rules required **for** L2TP/IPSec gateway"

playbooks/roles/l2tp-ipsec/templates/streisand-l2tp-iptables.service.j2
```diff
@@ -0,0 +1,14 @@
...
+After=network.target
+Before=ipsec.service
+
+[Service]
+Type=oneshot
+RemainAfterExit=true
+{% for rule in l2tp_ipsec_firewall_rules %}
```
cpu: I think it might be better to place a `l2tp_ipsec_firewall_up.sh` script on the system and have it run by `ExecStart` instead of having multiple `ExecStarts` in the one service directly. What do you think?

playbooks/roles/l2tp-ipsec/templates/streisand-l2tp-iptables.service.j2
```diff
@@ -0,0 +1,14 @@
...
+ExecStart=/sbin/{{ rule }}
```
cpu: A nice follow-up might be trying to make an `ExecStop` that would tear the firewall rules back down. Just mentioning in case you want another task afterwards in a similar area of the codebase :-)

playbooks/roles/l2tp-ipsec/tasks/main.yml
```diff
@@ -147,5 +145,15 @@
...
 - include: firewall.yml
 
+- name: Enable and start ipsec and l2tp services
+  systemd:
+    daemon_reload: yes
+    name: "{{ item }}"
+    enabled: yes
+    state: restarted
+  with_items:
+    - ipsec
```
cpu: Is there an existing `ipsec` service installed from upstream that this is using? 

I'm largely 100% ignorant about the IPsec/l2tp configuration Streisand uses but it seems like the former firewall init script called `ipsec setup start` and I'm wondering if that did something more than reloading the `ipsec` service will do.

@alimakki @jlund Do you know anything about this aspect?

playbooks/roles/l2tp-ipsec/templates/streisand-l2tp-iptables.service.j2
z0rc: Fixed.

playbooks/roles/l2tp-ipsec/templates/streisand-l2tp-iptables.service.j2
z0rc: I don't see a value in putting iptables execution in separate script. With separate script we have potential of masking issues with some iptables call. Also the separate script doesn't help if you need to reapply iptables rules, just run `systemctl restart streisand-l2tp-iptables`.

playbooks/roles/l2tp-ipsec/templates/streisand-l2tp-iptables.service.j2
z0rc: Yep, it's definitely a good task to look into afterwards.

playbooks/roles/l2tp-ipsec/tasks/main.yml
z0rc: Right, `ipsec.service` comes from upstream. If you trace the execution of `ipsec setup start`, you'll see that all it does is couple of safe-checks and calling `systemctl start ipsec`. Moreover `ipsec.service` includes the same safe-checks too.

z0rc:
Well, test is failing on something unrelated to this change. What's the procedure here?

playbooks/roles/l2tp-ipsec/templates/streisand-l2tp-iptables.service.j2
cpu: > With separate script we have potential of masking issues with some iptables calls

What kind of issues are you thinking? Like a non-zero exit? That's pretty easy to catch with `set -e`. It seems like it will cause duplication for the case where you wanted to also drop the rules no? With a script it might be easier to refactor the commonality between adding/dropping.

Why do you think it's better to have many `ExecStart`'s here? 

>  Also the separate script doesn't help if you need to reapply iptables rules, just run systemctl restart streisand-l2tp-iptables.

Wouldn't the same invocation also restart the script to reapply the firewall rules if it were the target of the `ExecStart`? If not then I agree that's a good advantage to this approach.

playbooks/roles/l2tp-ipsec/tasks/main.yml
cpu: @z0rc Aha, that makes sense. Thanks for clearing that up!

cpu:
> Well, test is failing on something unrelated to this change. What's the procedure here?

I think that must have been a transient problem. Sometimes a GPG key fetch will fail or another spooky Travis thing happens that breaks the build for a couple hours and then fixes itself. It seems OK to me right now, I don't think any action is required on your part for CI.

playbooks/roles/l2tp-ipsec/templates/streisand-l2tp-iptables.service.j2
z0rc: > It seems like it will cause duplication for the case where you wanted to also drop the rules no With a script it might be easier to refactor the commonality between adding/dropping.

I'm not sure what kind of duplication might happen here. It's still sequential execution of iptables one way or another. Multiple ExecStop is allowed here too.
> Why do you think it's better to have many ExecStart's here?

Because this is specifically supported by `Type=oneshot`. I still don't see a value of shell script which wraps sequential execution of iptables rules while this can be done directly in systemd unit. I prefer to keep things clean without unnecessary cruft.

z0rc: For example doing `systemctl cat streisand-l2tp-iptables` allows to view what exactly service does instantly, without need to dwell deeper into wrapper shell script.

cpu: Ok! I'm not entirely convinced but I don't want to bikeshed :-) I'm happy to leave this as-is in your PR. Thanks for engaging on the topic.

cpu:
@jlund @alimakki Any thoughts on this PR? Should I merge?

cpu:
I've tried a few times now to test this PR with an Android L2TP client and both times it has failed. I originally suspected this PR but I've tried twice more with master and have the same result. I'm not a L2TP user and can't easily tell if the issue is my particular Android client or a general regression. I've asked @nopdotcom to take a look at whether master has a regression and to test this PR with some clients.

Apologies about the delays getting this merged. I wanted to leave this note so you know it hasn't been forgotten.

nopdotcom:
In this PR, L2TP works with mobileconfigs on OS X and iOS. It doesn't work on the Android devices I grabbed. I'll try a deploy from master next.

nopdotcom:
I've filed #990, which is an L2TP regression for Android devices. It's present on the `master` branch too.

cpu:
@nopdotcom Thanks for confirming this PR works with OSX & iOS (And for opening the Android issue). I feel comfortable merging the PR now. 

@z0rc Thanks again for sticking with the review & being patient with us. I'll strive to merge future contributions with less lag. Would you be willing to look at some of the other places where Streisand could benefit from replacing init scripts with systemd units?

Issue #626, opened 2017-04-18
Linode: Alert_cpu_threshold error during provisioning.


depluser:
I am trying to deploy streisand on linode from an ubuntu vps server. I have seen it documented that local installs are not yet supported, but shouldn't it work if I'm coming from a completely different machine, even though it's not a workstation? But I am getting the localhost error:

===================================
Streisand will now set up your server. This process usually takes around ten minutes. Press Enter to begin setup...
:

PLAY [Provision the Linode Server] ***********************************************************************************************

TASK [Gathering Facts] ***********************************************************************************************************
ok: [localhost]

TASK [genesis-linode : Get the default SSH key] **********************************************************************************
changed: [localhost]

TASK [genesis-linode : Create the server] ****************************************************************************************
fatal: [localhost]: FAILED! => {"changed": false, "failed": true, "msg": "Alert_cpu_threshold must be between 0 and 2000."}
        to retry, use: --limit @/streisand/playbooks/linode.retry

PLAY RECAP ***********************************************************************************************************************
localhost                  : ok=2    changed=1    unreachable=0    failed=1

cpu:
@depluser What version of Ansible is installed on the Ubuntu VPS you are running Streisand from? What version of `linode-python` is installed?

depluser:
root@depl:~# ansible --version
ansible 2.3.0.0
  config file =
  configured module search path = Default w/o overrides
  python version = 2.7.12 (default, Nov 19 2016, 06:48:10) [GCC 5.4.0 20160609]

How can I find out what version of linode-python is installed?

Thanks!

cpu:
> How can I find out what version of linode-python is installed?

You should be able to use `dpkg -l linode-python`.

cpu:
Oops - I gave you bad advice. It's a pip package not a deb. Try:

`pip show linode-python`

depluser:
root@depl:~# pip show linode-python
Name: linode-python
Version: 1.1.1
Summary: Python bindings for Linode API
Home-page: https://github.com/tjfontaine/linode-python
Author: TJ Fontaine
Author-email: tjfontaine@gmail.com
License: UNKNOWN
Location: /usr/local/lib/python2.7/dist-packages
Requires:

depluser:
I was getting a warning about a pip upgrade being available, so I just went ahead and did the upgrade. Didn't change things tho - same error.

cpu:
> root@depl:~# ansible --version
> ansible 2.3.0.0

[...]

PR #949, opened 2017-09-16
Provide alert threshold defaults to `linode` task.

cpu:
Based on [a comment](https://github.com/jlund/streisand/issues/626#issuecomment-319812261) from Issue #626 this commit adds
`alert_x_threshold` values to the `linode` task in the `genesis-linode`
playbook. On the systems I've tested this is not necessary but also
harmless. For those affected by issue #626 it is necessary to avoid
a failure to create Linode servers based on the threshold values being
outside of validation range or undefined.

Thanks to @cmeid for suggesting the workaround and @i-s-o-g-r-a-m c for
testing the fix.

Resolves https://github.com/jlund/streisand/issues/626


playbooks/roles/genesis-linode/defaults/main.yml
```diff
@@ -1,3 +1,15 @@
...
 linode_plan_id: 1
 linode_distribution_id: 146
+
+# Threshold for receiving CPU usage alerts. Each CPU core adds 100% to total.
+# Since by default Streisand provisions a Linode 1024 with one core a value of
+# 90% seems ~reasonable
+linode_alert_cpu_threshold: 90
+
+# Other values left as the defaults from the Linode module. See
+# https://github.com/jlund/streisand/issues/626 for more detial.
```
jlund: detial -> detail

cpu: Good catch :eagle: :eyes:  Thanks!

cpu:
Merging with @alimakki's review since @jlund's feedback was small & addressed :-)

PR #958, opened 2017-09-21
Add new GCP zones in South America

pguizeline:


alimakki:
Thank you @pguizeline.

PR #968, opened 2017-09-28
GCE: document that id_rsa.pub must be registered with IAM.

nopdotcom:
Until we get a PR for registration of GCE keys, at least tell people they need to talk to IAM.


README.md
```diff
@@ -145,6 +145,8 @@ Complete all of these tasks on your local home machine.
 
         sudo pip install "apache-libcloud>=1.5.0"
 
+>For Google, your SSH `id_rsa.pub` must be registered with Google IAM; see [issue #919](https://github.com/jlund/streisand/issues/919).
```
cpu: I think you don't want the leading `>` here. It makes this a quote and gives the text a ghosted appearance that I think makes it harder to spot.

How about as an alternative you add this text immediately adjacent to the `* Google` bullet point on L144?

Can you also add the text in-English to the other READMEs? I don't know if it merits pinging the translators yet because I believe we have some other README updates coming down the line from @jlund.

README.md
```diff
@@ -142,9 +142,14 @@ Complete all of these tasks on your local home machine.
 
         sudo pip install dopy==0.3.5
   * Google
+<p>
```
cpu: This seems to cause an issue where the alignment of subsequent bullets is now off:
offset screenshot

nopdotcom: Gahhh. Lemme look at that again.

cpu: Thanks - sorry about all the markdown hassle!

cpu:
I'm going to close this in favour of https://github.com/StreisandEffect/streisand/pull/1048 - Thanks!

Issue #300, opened 2016-05-26
Specify your own SSH key


obilodeau:
I use several SSH keys side by side and I would like to be able to tell streisand to use a specific one. I can make my own `.ssh/config` file to access my host properly afterwards. An optional command line flag would be nice.

jlund:
Can you provide an example of what your desired workflow for this would look like? Ansible needs to be able to connect in the absence of an `.ssh/config` file in order to do the initial server configuration in a seamless way.

Right now the way that you would do this is by running the playbook directly on a new server that was created outside of Streisand instead of using the `./streisand` command.

jmusarra:
You can tell Ansible to use a specific key - in the inventory file, add something like:

`[streisand-host:vars]`
`ansible_ssh_private_key = /path/to/keyfile`

This presumes that you already have a host instead of letting streisand provision one for you, and that the host details are also in the inventory file.

nopdotcom:
Would it be satisfactory for Streisand to generate a short-lived SSH keypair just for Ansible deployment? 

At the end of installation it could install `your_favorite_key.pub` into `authorized_keys`, then delete its own public key.

obilodeau:
> Would it be satisfactory for Streisand to generate a short-lived SSH keypair just for Ansible deployment?

Yes, that would be better than what is happening right now for my use case.

### About the work-around

@jmusarra's workaround doesn't work for me with current master using digital ocean.

The server's IP address is unknown so `streisand-host` can't be defined in advance and it needs to be defined in order for the YAML inventory file to validate:

    Attempted to read "inventories/inventory" as ini file: inventories/inventory:4: Section [streisand-host:vars] not valid for undefined group: streisand-host


I also tried re-starting the installation using the `existing server` option but the private key file is not honored:

```
What is the IP of the existing server: 138.xx.xx.58

THIS WILL OVERWRITE CONFIGURATION ON THE EXISTING SERVER.
STREISAND ASSUMES 138.xx.xx.58 IS A BRAND NEW UBUNTU INSTANCE AND WILL
NOT PRESERVE EXISTING CONFIGURATION OR DATA.

ARE YOU 100% SURE THAT YOU WISH TO CONTINUE?

Please enter the word 'streisand' to continue: streisand

Confirmed. Continuing
Warning: Identity file /home/xxxx/.ssh/id_rsa not accessible: No such file or directory.
Permission denied (publickey).
```

[...]

PR #970, opened 2017-09-28
Support specifying custom SSH public key.

### LINKED ISSUE #739: hostvars['127.0.0.1']['ansible_system'] may not exist
### Expected behavior:

Installation completed without errors.

### Actual Behavior:

Installation completed with minor errors:
```
TASK [streisand-gateway : Attempt to open the instructions on Linux (if applicable). Errors in this task are ignored because the `xdg-open` command is not always available.] ***
fatal: [x.x.x.x]: FAILED! => {"failed": true, "msg": "The conditional check 'hostvars['127.0.0.1']['ansible_system'] == \"Linux\"' failed. The error was: error while evaluating conditional (hostvars['127.0.0.1']['ansible_system'] == \"Linux\"): 'dict object' has no attribute 'ansible_system'\n\nThe error appears to have been in '/media/veracrypt1/streisand/playbooks/roles/streisand-gateway/tasks/main.yml': line 199, column 9, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n      - name: Attempt to open the instructions on Linux (if applicable). Errors in this task are ignored because the `xdg-open` command is not always available.\n        ^ here\n"}
```

### Steps to Reproduce:

Run `./streisand`.

### Additional Details:

Really minor issue. I guess it should be fixed by replacing of
```
playbooks/roles/streisand-gateway/tasks/main.yml:202
- when: hostvars['127.0.0.1']['ansible_system'] == "Linux"
+ when: hostvars['127.0.0.1']['ansible_system'] is defined and hostvars['127.0.0.1']['ansible_system'] == "Linux"
```
and
```
playbooks/roles/streisand-gateway/tasks/main.yml:207
- when: hostvars['127.0.0.1']['ansible_system'] == "Darwin"
+ when: hostvars['127.0.0.1']['ansible_system'] is defined and hostvars['127.0.0.1']['ansible_system'] == "Darwin"
```

cpu:
Prior to this PR Streisand universally assumed the user's SSH private key was `~/.ssh/id_rsa`. The public component was then used to provision cloud instances.

This PR adds a new `streisand_ssh_private_key` var to the `global_vars/default-site.yml` file. The site-specific override for these defaults, `~/.streisand/site.yml` is now customized by `playbooks/customize.yml` when the user specifies that they want to use a different SSH key. The public component is expected to exist at the same file path with a `.pub` suffix to keep things simple.

This PR proactively sets the `ansible_ssh_private_key_file` for the Streisand  SSH key specified. This ensures Ansible will load & use this private key to access the streisand-host bypassing the need to use a ssh-agent (though it will play nicely if you do).

The validation playbook is updated to check that the  `streisand_ssh_private_key` exists. The corresponding shell based checks for the SSH key were removed from the `./streisand` wrapper script since they are now replaced by these Ansible tasks. Similarly the existing server playbook
now performs its own quick test of a raw_cmd on the server instead of doing a SSH check in the `./streisand` wrapper. This means things fail a little bit later in the process if you can't actually SSH to the
existing server but it isn't *much* later in the process and its nice to `rm` some shell script code.

Two unrelated bugs were fixed with the existing server provisioning:  the genesis role was not being saved as a fact on localhost causing the diagnostics to print "unknown" for the genesis role used. Secondly, because no `gather_facts: yes` for the localhost was run for the existing server provision other diagnostic information was missing. This was also the root cause of Issue #739. 

Resolves #739, #300 

I have tested this PR with a  few providers (Linode, DigitalOcean, Existing Server, Localhost) using my own SSH key loaded in an ssh-agent ahead of time as well as with a brand new ed25519 keypair that used a passphrase. Both cases worked as expected.


playbooks/customize.yml
```diff
@@ -4,6 +4,10 @@
   gather_facts: no
 
   vars_prompt:
+    - name: streisand_ssh_private_key
+      prompt: "Enter the path to your SSH private key? Press enter for default "
```
alimakki: "Enter the path to your SSH private key, or press enter for default" reads a bit better in my opinion.  Either way the question mark can be dropped.

playbooks/existing-server.yml
```diff
@@ -3,6 +3,30 @@
...
+  tasks:
+    - block:
+        - raw: whoami
+          args:
+            executable: /bin/bash
+          changed_when: False
+      rescue:
+        - fail:
+            msg: "Unable to SSH to existing streisand-host.\nEnsure private key corresponding to \"{{ streisand_ssh_key }}\" is loaded in your SSH key agent.\nTry using `ssh-keygen -i {{ streisand_ssh_key }} to generate your key if it does not exist\n"
+
```
alimakki: Good hint 👍

playbooks/customize.yml
cpu: Thanks! fixed in https://github.com/jlund/streisand/pull/970/commits/ef529d198bcf6643a39f5b6bc445c21ada698f48

cpu:
> Would this be simpler if we could guarantee Ansible runs under an ssh-agent, either an existing one or a freshly launched one? I ask because my ssh-agent talks to my Yubikey, so in some sense I don't have an id_rsa, only an id_rsa.pub.

It doesn't sound simpler to me :-) I'll give it some thought but I likely won't work on it myself if you're interested in trying as a follow-up. If so creating a second issue with some prelim details and assigning it to yourself would be great. Thanks!

PR #980, opened 2017-10-07
Proof of concept: docker builder

nopdotcom:
Probably you don't want to merge this.

In my continuing quest to make streisand setup easier, I decided to try Docker as an environment for a builder. This is the kind of thing that might go in a contrib directory, though.


Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+RUN mkdir /work
+RUN mkdir /work/streisand
+WORKDIR /work
+ADD . .git .travis.yml /work/streisand/
+WORKDIR /work/streisand
+
+CMD ["/work/streisand/streisand"]
+
+# ENTRYPOINT ["bash"]
+# ENTRYPOINT ["sleep 86400"]
```
cpu: Some comment cruft to nix here :point_up:

docker-build
```diff
@@ -0,0 +1,2 @@
+#!/bin/bash
```
cpu: Please use `-e`

docker-run
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
```
cpu: Dito `-e`

Dockerfile
```diff
@@ -0,0 +1,29 @@
+FROM ubuntu:16.04
+MAINTAINER nop@nop.com
```
cpu: `MAINTAINER` is [deprecated](https://docs.docker.com/engine/reference/builder/#maintainer-deprecated)

Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+
+RUN apt-get update -q
+RUN apt-get install --yes build-essential
```
cpu: You use `-y` elsewhere and `--yes` here. Could you update for consistency?

Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+RUN apt-get install -y python-pip git python-dev python-setuptools python-cffi libffi-dev libssl-dev python-nacl
+
+# Ansible
+RUN pip install ansible
+
+# pipping
+RUN pip install boto boto3 msrest msrestazure azure==2.0.0rc5 packaging dopy==0.3.5 "apache-libcloud>=1.5.0" linode-python pyrax
+
+RUN mkdir /work
+RUN mkdir /work/streisand
```
cpu: These two `RUN mkdir`'s can be collapsed into one `RUN mkdir -p /work/streisand`

Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+WORKDIR /work
+ADD . .git .travis.yml /work/streisand/
+WORKDIR /work/streisand
+
+CMD ["/work/streisand/streisand"]
```
cpu: It's probably redundant for folks familiar with Docker but could you add a comment here that specifies this is the default argument used by the entrypoint? I had to look up `CMD`

docker-run
```diff
@@ -0,0 +1,16 @@
...
+    docker run -it --rm -v $(pwd)/generated_docs:/work/streisand/generated_docs \
+	   -v $HOME/.ssh:/work/streisand/sshkeys \
+	   -e STREISAND_SSH_KEY_FROM_HOST="/work/streisand/sshkeys/id_rsa.pub" \
+	   -e STREISAND_SSH_ADD_KEY=1 \
+	   streisand-builder ssh-agent "$@"
+}
+
+if [ -z "$1" ]
+then
+    run_docker /work/streisand/streisand
```
cpu: Is this redundant with the `CMD` in the Dockerfile? Could you do `run_docker` here instead of `run_docker /work/streisand/streisand` and let the `CMD` specify the default path?

docker-run
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
+run_docker () {
+    docker run -it --rm -v $(pwd)/generated_docs:/work/streisand/generated_docs \
```
cpu: Can you have this script also check that `docker` is a command in the `$PATH` and if not indicate that you need to install docker first?

streisand
```diff
@@ -52,6 +52,24 @@ function check_python() {
 function check_ssh_key() {
   local STREISAND_SSH_KEY="$HOME/.ssh/id_rsa.pub"
 
+  # In case we're sharing SSH keys inside Docker
```
cpu: Unfortunately your diffs to `streisand` to support this SSH key stuff is going to have to be reworked on top of https://github.com/StreisandEffect/streisand/pull/970. Apologies.

Dockerfile
```diff
@@ -0,0 +1,29 @@
+FROM ubuntu:16.04
```
cpu: I think this is likely to be a better approach than the `FROM python:2.7.12` base image used in https://github.com/StreisandEffect/streisand/pull/442 :+1:

Dockerfile
nopdotcom: Done.

Dockerfile
nopdotcom: Done.

Dockerfile
nopdotcom: Docker creates directories when not present. I've eliminated this.

Dockerfile
nopdotcom: Done.

Dockerfile
nopdotcom: Done.

docker-build
nopdotcom: OK; to be clear, house style is:

```
#!/bin/bash -e
```

instead of

```
#!/bin/bash
set -e
```
? Happy to do whichever.

docker-run
nopdotcom: I started doing this with `-z "$(type -f docker 2>/dev/null)"` but turned it into an `if ! docker 2>/dev/null` instead; any preference?

docker-run
nopdotcom: Yes, and I've eliminated that cruft.

streisand
nopdotcom: It was based on my broken ideas of how we were handling SSH keys. My current plan is to make a separate wrapper around `./streisand` to handle copying host `$HOME/.ssh` into `/root/.ssh`.

docker-build
cpu: I'm ok with either. `set -e` is a little better since you can add a comment at the same time that spells out that errors are expected to terminate immediately.

docker-run
cpu: I think the first might be more idiomatic for a "command is installed" check and is what we do in [the util/ansible_check.sh script](https://github.com/StreisandEffect/streisand/blob/master/util/ansible_check.sh#L11).

Dockerfile
nopdotcom: TBH, I'm just churning these out looking for the easiest path from A to B. Use of Docker for this is pretty heavy-duty, but in return, we get an entire OS we can scribble on; this is similar to the assumptions made in the `cloud-init` stuff.

Python virtualenv is sorta the next step down.

Yes, this stuff needs to migrate into a subdirectory.

docker-run
nopdotcom: OK, I'm planning on making a util/docker_check.sh, because besides existence and version, there are probably some other things I can do for sanity checking.

docker-build
```diff
@@ -1,2 +1,16 @@
 #!/bin/bash
+set -e
+if [ -z "$(type -f docker 2>/dev/null)" ]; then
```
cpu: Instead of duplicating the docker command check here and in `docker-run` can you put it in a separate `util` script like the Ansible command check?

docker-run
```diff
@@ -18,8 +18,8 @@ run_docker () {
     #
     # Currently, ssh keys need to exist in $HOME/.ssh/ to be available.
     docker run -it --rm --init \
-	   -v "$(pwd)/generated_docs:/work/streisand/generated_docs" \
-	   -v "$HOME/.ssh:/root/.ssh" \
+	   -v "$(pwd)/generated-docs:/work/streisand/generated-docs" \
+	   -v "$HOME/.ssh:/work/sshkeys" \
```
cpu: I'm a little bit nervous about mounting the whole `.ssh` directory into the container and I suspect it might give some people heeby-jeebies too. Is it possible to prompt in Bash for a specific keyfile to mount into the container? That might reduce the surprise and minimize potential damage if something in the container goes rogue.

streisand
cpu: Seeing it now I think it would be better to try and only mount the single key required into the container vs the entire `.ssh` directory (https://github.com/StreisandEffect/streisand/pull/980#discussion_r145297893)

docker-run
alimakki: Another approach would be to generate an ssh key on the fly output it somewhere at the end of the run.

nopdotcom:
This has gotten very stale, and I have gotten smarter.

Dockerfile
hammady: I don't agree on basing on Ubuntu. Python base is the shortest path where no need to rework a lot of stuff. virtualenv is nothing to look at when using docker, the whole container is a virtual env by itself.

PR #574, opened 2017-04-04
Specify GRUB2 kernel by default when deploying via Linode

rmcintosh:
This ties to https://github.com/ansible/ansible/pull/23100 which was submitted to help ease the issues described in https://github.com/jlund/streisand/issues/500

The change makes the deployed instance use the GRUB2 Linode kernel option which now allows use of the distribution's supplied kernel, rather than the hypervisor's kernel.

I'm not intimately familiar with streisand so If I've missed the mark somewhere, please let me know.

PR #982, opened 2017-10-11
Use GRUB2 kernel option when deploying to Linode

rmcintosh:
See https://github.com/StreisandEffect/streisand/pull/574

I'm removing the monkey patch stuff because it seems like it only ever got used for this... will re-add it if that isn't kosher


cpu:
Thanks @rmcintosh! 

It looks like this is dying in CI, likely as a side effect of the Ansible version bump and not from your diffs. 

If you're not feeling up to playing whack-a-mole with our CI feel free to say so and I can try and figure out the problem in the next few days.
If you'd like to try I'm happy to let you! :-)

rmcintosh:
I'm up for taking a wack at it 🙂

wzyboy:
Hi @rmcintosh ,

as @cpu mentioned in https://github.com/StreisandEffect/streisand/issues/994#issuecomment-336456257 , the CI fails if you bump Ansible version from 2.3 to 2.4. This is because in 2.4 they changed the way of handling `roles_path` in `ansible.cfg` -- it's always relative to CWD.

I encountered the same CI problem as you do when porting to Ansible 2.4 in https://github.com/StreisandEffect/streisand/pull/996 . The CI is now passing after updating the `roles_path` setting in `tests/ansible.cfg`.

If this PR is to fix the issue of Linode, would you please remove the version-bump-related commits? :-)

cpu:
@wzyboy I'm glad you were able to figure out the CI problem :-)

> EDIT: Oh I'm sorry, I did not read #574. So it's Ansible 2.4 that fixes this issue. Maybe we could merge #996 before #982 (without the version bump) to avoid merging conflicts.

I think the merge conflicts will be minimal either way. I will review both ASAP and we can decide which make sense to merge first and which will need to resolve with master.

cpu:
I merged #996 - @rmcintosh Can you rebase this PR on top of master when you get a chance? Thanks! You shouldn't have to do any CI battle now :no_entry_sign: :crossed_swords: :tada:

PR #988, opened 2017-10-12
Modular Streisand: improved test coverage of enabled/disabled services

cpu:
This PR updates the `tests/tests.sh` driver script so that different build configurations can be run as subtasks depending on a provided env var. The `.travis.yml` configuration is updated to use a matrix of env vars such that we do several subtasks for every CI run. Presently there are 6 build subtasks:

1. A build with everything enabled & all of the syntax checking (the same as was done prior to this commit for all CI runs)
2. A build with just l2tp enabled
3. A build with just openvpn enabled
4. A build with just shadowsocks enabled
5. A build with just the SSH forward user enabled
6. A build with the enabled/disabled services randomized

Care was taken with the randomized build to ensure that there is always 1 service enabled and that none of the dependent services end up in a state that the validation role will reject (e.g. stunnel without openvpn). It is my hope that with the random build we will get some good coverage over time, especially with the daily cron build.

The `development-setup.yml` playbook was updated to only install/build Libreswan on the container host if the container will use l2tp. This should help speed up non-l2tp builds.

Updates #746


cpu:
I'll have to catch this branch up with master in the next few days. Out of time for tonight and wanted to get my diffs so far out for review.

cpu:
Broke the build with a last minute change too :blush: - I put a revisions needed label on and will get this fixed shortly.

tests/site_vars/l2tp.yml
```diff
@@ -0,0 +1,13 @@
+---
+# This site config only enables L2TP
+
+vpn_clients: 5
```
alimakki: Nit pick, but as it is now l2tp doesn't do much with vpn clients, can be set to 1 for this case

cpu:
FWIW I did 18 separate builds of just the randomized build task (branches `cpu-optional-test-matrix-run-1` through `cpu-optional-test-matrix-run-18` on my fork). 100% of them completed successfully. Builds with as few as 3 services were tested as well as many of 8. Pretty good differences in which services were tested between individual builds. It's not perfect but I'm happy with it!

cpu:
Squashed the last bugs and caught this branch up with master.

@jlund @alimakki @nopdotcom @nickolasclarke This is ready for a review pass now.

.travis.yml
```diff
@@ -9,6 +9,13 @@ sudo: required
 env:
   global:
     - ANSIBLE_CONFIG=tests/ansible.cfg
+  matrix:
+    - RUN="shellcheck yamlcheck syntax ci" SITE="global_vars/default-site.yml"
+    - RUN="ci" SITE="tests/site_vars/l2tp.yml"
+    - RUN="ci" SITE="tests/site_vars/openvpn.yml"
+    - RUN="ci" SITE="tests/site_vars/shadowsocks.yml"
+    - RUN="ci" SITE="tests/site_vars/ssh.yml"
+    - RUN="ci" SITE="random"
```
alimakki: Another entry for OpenConnect?

tests/libreswan-setup.yml
```diff
@@ -0,0 +1,58 @@
+---
+
+# Since LXC loads kernal modules from the host, we need to ensure libreswan is compiled and installed
```
alimakki: kernal -> kernel 
libreswan -> LibreSwan (nitpick)

alimakki:
I think maybe somewhere down the line, we could also randomize the number of VPN clients.
But that's for another PR :)

.travis.yml
cpu: I was initially thinking that a build task for each service in isolation was too many and so I'd pick a few services to do that with and then let the random build catch the rest.

Since it ended up being _only_ OpenConnect that isn't represented here I might as well add it :-)

cpu:
> I think maybe somewhere down the line, we could also randomize the number of VPN clients.
But that's for another PR :)

@alimakki - Agreed! That's a very natural follow-up addition to the random build task. I thought about putting it into this branch but it was already a pretty big hunk of work.

Thanks for the review! Feedback addressed & conflicts resolved.

PR #998, opened 2017-10-14
POC: Replace id_rsa with just id_rsa.pub and a working ssh-agent

nopdotcom:
(Another one of those "don't merge this" specials, to document where I am.)

We don't need the private key for anything; what we need is the
capability of logging in. Depend on a working `ssh-agent` that can
unlock `~/id_rsa.pub`.

This PR is not user-friendly now. It requires people who don't
understand ssh to understand `ssh-agent` too. It could be made more
friendly for the old case of an existing `id_rsa`/`id_rsa.pub` by:

1. forcing an `ssh-agent` inside of streisand

2. running `ssh-add` early on, so we can bomb if the user can't unlock
the key.

This is what I do in the Docker builder.

This commit breaks CI, and I don't know how to fix it yet.

### DIFF
--- a/global_vars/default-site.yml
+++ b/global_vars/default-site.yml
@@ -7,7 +7,7 @@
 # The SSH private key that Ansible will use to connect to the Streisand node.
 # The associated public key will be used if required when provisioning cloud
 # nodes for the authorized_keys file.
-streisand_ssh_private_key: "~/.ssh/id-rsa"
+streisand_ssh_public_key: "~/.ssh/id_rsa.pub"
 
 vpn_clients: 5
 
--- a/playbooks/customize.yml
+++ b/playbooks/customize.yml
@@ -4,9 +4,9 @@
   gather_facts: no
 
   vars_prompt:
-    - name: streisand_ssh_private_key
-      prompt: "Enter the path to your SSH private key, or press enter for default "
-      default: "~/.ssh/id_rsa"
+    - name: streisand_ssh_public_key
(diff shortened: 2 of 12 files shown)


cpu:
Ah, I misunderstood your earlier message about this. 

> We don't need the private key for anything;

It's required for `ansible_ssh_private_key` to work.

If you look back through the comment history on #300 and match it up to my WIP branches you'll see I actually implemented the custom SSH key support based on only knowing the public key initially like you propose here. It didn't work well for @obilodeau's use-case and I suspect there are other users in the same position.

I switched to asking for the private key location because that's the only way to make the `ansible_ssh_private_key_file` setting work. As you mention without being able to use `ansible_ssh_private_key` you're stuck having to understand ssh-agent.

It might be possible to find middle ground but (as you point out) this PR isn't it (yet).

nopdotcom:
My mental framework is that filenames of pairs are not that relevant, except as a transient way of distinguishing keys. I've had my (GPG) ssh-agent hold keys named like `./id_rsa` and who knows if that's the same as `~/.ssh/id_rsa` which it also held. (It wasn't.)

ssh doesn't appear to think about keys as content-addressable :‐( 

I think any solution that involves an ssh-agent is also going to involve the user picking which line from `ssh-add -l` they want to use. Then the problem is convincing `/usr/bin/ssh` to only use that particular identity.

Handling the "use *this* key" stuff sounds like it's big enough and reusable enough that it should go in a separate project.

Issue #906, opened 2017-08-27
OS X cert trust instructions don’t trust cert


nopdotcom:
### Expected behavior:

From `streisand.html`’s description of how to install the cert on OS X:

> * Go to Applications > Utilities and launch Keychain Access.
> * Download the SSL certificate embedded in this document.
> * Drag the downloaded certificate into the login keychain section.
> * A window will appear asking if you want to trust the certificate. Click Always Trust.

### Actual Behavior:

On OS X 10.11.6 and 10.12.6, no window appears asking if you want to trust the certificate.

### Steps to Reproduce:
1. Follow the OS X cert installation instructions until the trust window doesn’t show up.

### Additional Details:

Here’s what you have to do to make it work, I think.

![capto_capture 2017-08-27_03-51-28_pm copy](https://user-images.githubusercontent.com/1357866/29753607-e128e220-8b42-11e7-9ff0-948333027bc6.png)

![capto_capture 2017-08-27_03-51-49_pm](https://user-images.githubusercontent.com/1357866/29753605-c5304fcc-8b42-11e7-9b65-9079d26acf67.png)

However, this only gets us as far as the SSL prompt, which is still broken because of SAN issues. From Safari:

![capto_capture 2017-08-27_04-19-31_pm](https://user-images.githubusercontent.com/1357866/29753635-9d210e9e-8b43-11e7-9adf-e08e036071f2.png)

##### *Target Cloud Provider:*

EC2

##### *Operating System of client:*

Builder is running 16.04.3 on EC2. Clients are OS X 10.11.6 and 10.12.6.

##### *Version of Ansible, using `ansible --version` :*

```
ansible 2.3.1.0
  config file = /home/ubuntu/jlund-streisand/ansible.cfg
  configured module search path = Default w/o overrides
```

##### *Output from `git rev-parse HEAD` in your Streisand directory :*

acea35ad45533beb17235667bf78fd9a91794caf

nopdotcom:
I don’t see much point in fixing this until the SAN stuff is figured out.

Is the name of the server secret? It might be useful to have some kind of server ID (like IP address) in the server’s CA, to make it easier to find which certificate to add or delete.

nopdotcom:
You don’t have to trust the root CA fully--not that this fixes the SAN problem. You just need to trust the basic constraints:

[...]

PR #999, opened 2017-10-14
Refactor OpenSSL CA/Server/Client certificate generation into its own role

alimakki:
This PR aims to refactor CA/Server/Client certificate generation into its own role (in a similar vein as the download-and-veryify role) with the goals of:

 - Reducing code duplication by way of centralizing certificate generation
 - Separation of CA/server and client certificate generation into separate tasks (separation of concerns)
 - Simplify any future endeavor if OpenSSL need be replaced with alternative cryptography toolkits (e.x. LibreSSL)

Additionally, the changes in this PR will now assign subject alternative names to (the self-signed) server certificate, which may help in addressing the issue presented in #906 (to be issued in a separate PR).

The OpenVPN and OpenConnect roles have been refactored to utilize the role, without any observable regressions thus far.

### DIFF
--- a/documentation/certificates.md
+++ b/documentation/certificates.md
@@ -1,0 +1,91 @@
+Streisand PKI
+=============
+
+Streisand includes a role responsible for PKI certificate generation,
+separated into two distinct tasks:
+
+ - CA/Server
+ - Client(s)
+
+The first of which will generate a certificate authority, and a server 
+certificate signed by the newly minted certificate authority.
+
+The latter task is responsible for generating client side certificate
+(using the same CA) used for authenticting connecting VPN clients.
+
+Each VPN service that leverages PKI generates its own CA/Server and 
+client certificates.
+
+Infrastructure
+--------------
+
+The overall PKI infrasture assumes the following layout:
+
+```
+Web Root CA: 
+   \__ gateway HTTPS cert (unless LE; see below)
+
+OpenConnect CA: 
+   \__ OpenConnect server cert
+   \__ OpenConnect client 1
+   \__ OpenConnect client ...
+   \__ OpenConnect client n
+
+OpenVPN CA: 
+   \__ OpenVPN server cert
+   \__ OpenVPN client 1
+   \__ OpenVPN client ...
+   \__ OpenVPN client n
+
+IPsec-IKEv2 CA (upcoming):
+   \__ IPsec-IKEv2 server cert
+   \__ foobar client 1
+   \__ foobar client ...
+   \__ foobar client n
+
+ISRG X1:
+   \__ Let's Encrypt Authority X3
+        \__ poem-walk.example.com HTTPS cert
+``` 
+
+Usage
+-----
+
+Utilizing the `certificates` role can be invoked as follows (using OpenVPN as an example):
+(Should client certificates need to be generated, they should be done so priort to invocation)
+
+```
+- name: Create directories for clients
+  file:
+    path: "/etc/openvpn/{{ client_name.stdout }}"
+    state: directory
+  with_items: "{{ vpn_client_names.results }}"
+  loop_control:
+    loop_var: "client_name"
+    label: "{{ client_name.item }}"
+
+- include_role:
+    name: certificates
+  vars:
+    ca_path:            "/etc/openvpn"
+    tls_ca:             "/etc/openvpn/ca"
+    tls_client_path:    "/etc/openvpn"
+    generate_ca_server: yes
+    generate_client:    yes
+    tls_request_subject:         "/C=US/ST=California/L=Beverly Hills/O=ACME CORPORATION/OU=Anvil Department"
+    tls_server_common_name_file: "/etc/openvpn/openvpn-common-name.txt"
+    tls_sans:                    # sets the Subject Alternative Name(s) attribute(s)
+      - "1.2.3.4"
+      - "5.6.7.8"
+```
+
+Note the following flags:
+ - `generate_ca_server: yes`
+ - `generate_client: yes`
+
+ These must be made explicit, as the default value is `no`; certificates should not be generated unless
+ they have to be.
+
+The certificates role makes certain assumptions with regards to default values,
+all of which can be inspected [here](https://github.com/StreisandEffect/streisand/blob/master/playbooks/roles/certificates/defaults/main.yml).
+
--- a/documentation/certificates.md
+++ b/documentation/certificates.md
@@ -7,23 +7,23 @@ separated into two distinct tasks:
  - CA/Server
  - Client(s)
 
-The first of which will generate a certificate authority, and a server 
+The first will generate a certificate authority, and a server 
 certificate signed by the newly minted certificate authority.
 
-The latter task is responsible for generating client side certificate
-(using the same CA) used for authenticting connecting VPN clients.
+The latter task is responsible for generating client side certificates
+(using the same CA) used for authenticating connecting VPN clients.
 
-Each VPN service that leverages PKI generates its own CA/Server and 
-client certificates.
+Each VPN service that leverages the Streisand PKI generates its
+own CA/Server certificate and client certificates.
 
 Infrastructure
 --------------
 
-The overall PKI infrasture assumes the following layout:
+The overall PKI infrastructure assumes the following structure:
 
 ```
 Web Root CA: 
-   \__ gateway HTTPS cert (unless LE; see below)
+   \__ gateway HTTPS cert (unless Let's Encrypt; see below)
 
 OpenConnect CA: 
    \__ OpenConnect server cert
@@ -37,25 +37,20 @@ OpenVPN CA:
    \__ OpenVPN client ...
    \__ OpenVPN client n
 
-IPsec-IKEv2 CA (upcoming):
-   \__ IPsec-IKEv2 server cert
-   \__ foobar client 1
-   \__ foobar client ...
-   \__ foobar client n
-
 ISRG X1:
    \__ Let's Encrypt Authority X3
-        \__ poem-walk.example.com HTTPS cert
+        \__ streisand.example.com HTTPS cert
 ``` 
 
 Usage
 -----
 
 Utilizing the `certificates` role can be invoked as follows (using OpenVPN as an example):
-(Should client certificates need to be generated, they should be done so priort to invocation)
+(If client certificates need to be generated, a directory for each client must be created
+first)
 
 ```
-- name: Create directories for clients
+- name: Create directories for clients # needed if client certificates are required
   file:
     path: "/etc/openvpn/{{ client_name.stdout }}"
     state: directory
--- a/playbooks/roles/certificates/defaults/main.yml
+++ b/playbooks/roles/certificates/defaults/main.yml
@@ -1,0 +1,9 @@
+---
+tls_key_country:  "US"
+tls_key_province: "California"
+tls_key_city:     "Beverly Hills"
+tls_key_org:      "ACME CORPORATION"
+tls_key_ou:       "Anvil Department"
+tls_days_valid:   "1825"
+tls_default_md:   "sha256"
+tls_key_size:     "4096"
--- a/playbooks/roles/certificates/defaults/main.yml
+++ b/playbooks/roles/certificates/defaults/main.yml
@@ -7,3 +7,8 @@ tls_key_ou:       "Anvil Department"
 tls_days_valid:   "1825"
 tls_default_md:   "sha256"
 tls_key_size:     "4096"
+
+# What type of certificates to be generated
+# must be explicitly set
+generate_ca_server: no
+generate_client:    no
--- a/playbooks/roles/certificates/defaults/main.yml
+++ b/playbooks/roles/certificates/defaults/main.yml
@@ -9,6 +9,7 @@ tls_default_md:   "sha256"
 tls_key_size:     "4096"
 
 # What type of certificates to be generated
-# must be explicitly set
+# must be explicitly set by playbooks that
+# include the certificates role
 generate_ca_server: no
 generate_client:    no
--- a/playbooks/roles/certificates/tasks/ca-server.yml
+++ b/playbooks/roles/certificates/tasks/ca-server.yml
@@ -1,5 +1,5 @@
 ---
-- name: "Generate RSA keys for the CA and Server"
+- name: "Generate the private keys for the CA and Server certificates"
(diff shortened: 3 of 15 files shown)


playbooks/roles/certificates/tasks/ca-server.yml
```diff
@@ -0,0 +1,67 @@
...
+    owner: root
+    group: root
+    mode: 0600
+
+- name: Register the TLS server common name
+  command: cat "{{ tls_server_common_name_file }}"
+  register: tls_server_common_name
+  changed_when: False
+
+- name: Generate the OpenSSL configuration that will be used for the Server certificate's req and ca commands
```
nopdotcom: I'm confused by what the `Server` is.

playbooks/roles/certificates/tasks/client.yml
```diff
@@ -0,0 +1,39 @@
+---
+- name: Generate RSA keys for the clients
```
nopdotcom: I think this role could use a little documentation, since it's designed to be reused. When I got to this line, I said "which clients?" too. By now I've figured out what you're doing, but it wasn't immediately obvious to me at the time.

playbooks/roles/certificates/tasks/main.yml
```diff
@@ -0,0 +1,6 @@
+---
+- include: ca-server.yml
```
cpu: I think these should be `import_tasks` for Ansible 2.4.0+

playbooks/roles/certificates/templates/openssl.cnf.j2
```diff
@@ -42,25 +43,37 @@ emailAddress = optional
...
 [ req ]
 distinguished_name = req_distinguished_name
+req_extensions     = req_ext
+
+[ req_ext ]
+subjectAltName = @alt_names
+
+[ alt_names ]
+{% for item in tls_san %}
+IP.{{ loop.index }} = {{ item }}
```
cpu: Can you also add a `DNS.` SAN here? Even if the value is an IP address I think it will be better for compatibility and these self-signed certs are pretty YOLO already :-)

playbooks/roles/openconnect/tasks/main.yml
```diff
@@ -23,71 +23,26 @@
...
 
-- name: Generate the client certificates
-  command: certtool --generate-certificate --load-privkey {{ ocserv_path }}/{{ client_name.stdout }}-key.pem --load-ca-certificate {{ ocserv_ca_certificate_file }} --load-ca-privkey {{ ocserv_ca_key_file }} --template {{ ocserv_path }}/client-{{ client_name.stdout }}.tmpl --outfile {{ ocserv_path }}/{{ client_name.stdout }}-cert.pem
-  args:
-    creates: "{{ ocserv_path }}/{{ client_name.stdout }}-cert.pem"
-  with_items: "{{ vpn_client_names.results }}"
-  loop_control:
-    loop_var: "client_name"
-    label: "{{ client_name.item }}"
+- include_role:
```
cpu: I think this will need to be `import_playbook` for Ansible 2.4.0+

playbooks/roles/openconnect/templates/instructions-fr.md.j2
```diff
@@ -62,11 +62,11 @@ Les certificats clients sont un mécanisme par lequel les clients peuvent s'auth
    `brew install openconnect`
 
    * Si l'installation de Homebrew n'est pas une option, vous pouvez également télécharger et compiler le [code source OpenConnect](/mirror/index-fr.html#openconnect).
-1. Téléchargez le fichier [ca-cert.pem](/openconnect/ ca-cert.pem) et un [fichier de certificat client](#clientcerts) dans la liste ci-dessus.
+1. Téléchargez le fichier [ca.crt](/openconnect/ ca.crt) et un [fichier de certificat client](#clientcerts) dans la liste ci-dessus.
```
cpu: I think a space slipped into this URL: `[ca.crt](/openconnect/ ca.crt)` vs `[ca.crt](/openconnect/ca.crt)`

playbooks/roles/openvpn/tasks/main.yml
```diff
@@ -26,99 +17,24 @@
...
-  args:
-    creates: "{{ openvpn_server_common_name_file }}"
-
-- name: Set permissions on the OpenVPN server common name file
-  file:
-    path: "{{ openvpn_server_common_name_file }}"
-    owner: root
-    group: root
-    mode: 0600
+- include_role:
```
cpu: Ditto: `import_playbook`

playbooks/roles/certificates/templates/openssl.cnf.j2
```diff
@@ -69,6 +82,7 @@ subjectKeyIdentifier=hash
 authorityKeyIdentifier=keyid,issuer:always
 extendedKeyUsage=serverAuth
 keyUsage = digitalSignature, keyEncipherment
+subjectAltName = @alt_names
 
 [ client ]
 basicConstraints=CA:FALSE
```
cpu: I think the client certs also need an `extendedKeyUsage` that includes `clientAuth`. I _think_ that's why `playbooks/roles/openconnect/templates/client.tmpl.j2` had a line `tls_www_client` that I don't see replicated here.

playbooks/roles/certificates/tasks/ca-server.yml
```diff
@@ -0,0 +1,67 @@
+---
+- name: "Generate RSA keys for the CA and Server"
```
cpu: How do you feel about removing the "RSA" bits from the `name`s in the certificates roles? One day I hope we can have `ECDSA` certificates as well :-) For the level of detail required for a `name` output maybe something like "Generate private keys for the CA and Server" would be sufficient and save updating in the future if we change away from RSA.

playbooks/roles/certificates/tasks/ca-server.yml
```diff
@@ -0,0 +1,67 @@
...
+  file:
+    path: "{{ ca_path }}"
+    recurse: yes
+    state: directory
+    owner: root
+    group: root
+    mode: 0600
+
+- name: Generate CA certificate
+  command: openssl req -nodes -batch -new -x509 -key {{ tls_ca }}.key -out {{ tls_ca }}.crt -days {{ tls_days_valid }} -subj "{{ tls_request_subject }}/CN=ca-certificate"
```
cpu: The CA config sets `default_days = {{ tls_days_valid }}` - is this redundant?

playbooks/roles/certificates/templates/openssl.cnf.j2
```diff
@@ -42,25 +43,37 @@ emailAddress = optional
 
 [ req ]
 distinguished_name = req_distinguished_name
+req_extensions     = req_ext
+
+[ req_ext ]
+subjectAltName = @alt_names
+
+[ alt_names ]
+{% for item in tls_san %}
```
cpu: Small nit: I think `tls_san` should be plural, `tls_sans`.

playbooks/roles/certificates/tasks/client.yml
cpu: @alimakki How about writing a tiny file in `docs/` on certificate generation to go with this? Or maybe a mention in `CONTRIBUTING.md` "How to generate certificates"?

nopdotcom: Explicitly showing it as a tree might help. TBH, I am confused by all of the certs wandering around.

```
Root CA: (dead)
   \__ gateway HTTPS cert (unless LE; see below)
   \__ Foobar server cert
   \__ Foobar CA
        \__ foobar client 1

Let's Encrypt Authority:
   \__ poem-walk.example.com HTTPS cert
```

playbooks/roles/certificates/templates/openssl.cnf.j2
nopdotcom: Could we add this later after a first round of testing? It's easy to add later, and it's one fewer dead chicken to swing around if not necessary.

playbooks/roles/certificates/tasks/client.yml
cpu: I think presently there are multiple Root CAs being generated:

```
Web Root CA: 
   \__ gateway HTTPS cert (unless LE; see below)

Foobar Root CA: 
   \__ Foobar server cert
   \__ foobar client 1

ISRG X1:
   \__ Let's Encrypt Authority X3
        \__ poem-walk.example.com HTTPS cert
```

We can improve on the structure of the internal PKI as a follow-up. The way you're imagining it is closer to how it probably should be (though I would make one root CA and then issue constrained intermediates per-service).

playbooks/roles/certificates/templates/openssl.cnf.j2
cpu: Sure. We can take this as a follow-up after a baseline level of compatibility is established. :+1:

playbooks/roles/certificates/tasks/client.yml
alimakki: @cpu, I'll try to write up something in a separate .md file, should be easier to hunt down if someone were to look for it.

playbooks/roles/openconnect/tasks/main.yml
alimakki: Fails in [CI](https://travis-ci.org/alimakki/streisand/builds/288296374) if changed

playbooks/roles/openvpn/tasks/main.yml
alimakki: ditto ^

alimakki:
@cpu, I think this is ready for another review, now with 100% more documentation :)

playbooks/roles/openconnect/tasks/main.yml
cpu: Does this `include_role` generate a warning in 2.4.0?

alimakki: Nope

cpu: Then I'm satisfied :-)

alimakki: FWIW even the download-and-verify stuff weren't touched as part of the Ansible 2.4 update

documentation/certificates.md
```diff
@@ -0,0 +1,91 @@
+Streisand PKI
+=============
+
+Streisand includes a role responsible for PKI certificate generation,
+separated into two distinct tasks:
+
+ - CA/Server
+ - Client(s)
+
+The first of which will generate a certificate authority, and a server
```
cpu: nit: I think you can remove "of which" and just say "The first will generate"...

documentation/certificates.md
```diff
@@ -0,0 +1,91 @@
...
+Streisand includes a role responsible for PKI certificate generation,
+separated into two distinct tasks:
+
+ - CA/Server
+ - Client(s)
+
+The first of which will generate a certificate authority, and a server 
+certificate signed by the newly minted certificate authority.
+
+The latter task is responsible for generating client side certificate
```
cpu: "client side certificate" -> "client side certificates"

documentation/certificates.md
```diff
@@ -0,0 +1,91 @@
...
+(using the same CA) used for authenticting connecting VPN clients.
```
cpu: typo: "authenticting"

documentation/certificates.md
```diff
@@ -0,0 +1,91 @@
...
+
+Each VPN service that leverages PKI generates its own CA/Server and
```
cpu: "that leverages PKI" -> "that leverages the Streisand PKI"

cpu: "CA/Server" -> "CA/Server certificate"

documentation/certificates.md
```diff
@@ -0,0 +1,91 @@
...
+The latter task is responsible for generating client side certificate
+(using the same CA) used for authenticting connecting VPN clients.
+
+Each VPN service that leverages PKI generates its own CA/Server and 
+client certificates.
+
+Infrastructure
+--------------
+
+The overall PKI infrasture assumes the following layout:
```
cpu: typo: "infrasture"

cpu: I think "layout" might be better described as "structure". What do you think?

documentation/certificates.md
```diff
@@ -0,0 +1,91 @@
...
+   \__ OpenConnect client ...
+   \__ OpenConnect client n
+
+OpenVPN CA: 
+   \__ OpenVPN server cert
+   \__ OpenVPN client 1
+   \__ OpenVPN client ...
+   \__ OpenVPN client n
+
+IPsec-IKEv2 CA (upcoming):
```
cpu: I think it's probably less confusing if you leave this out for now since it's upcoming

documentation/certificates.md
```diff
@@ -0,0 +1,91 @@
...
+   \__ IPsec-IKEv2 server cert
+   \__ foobar client 1
+   \__ foobar client ...
+   \__ foobar client n
+
+ISRG X1:
+   \__ Let's Encrypt Authority X3
+        \__ poem-walk.example.com HTTPS cert
```
cpu: What about `streisand` instead of `poem-walk` here?

documentation/certificates.md
```diff
@@ -0,0 +1,91 @@
...
+``` 
+
+Usage
+-----
+
+Utilizing the `certificates` role can be invoked as follows (using OpenVPN as an example):
+(Should client certificates need to be generated, they should be done so priort to invocation)
```
cpu: typo: "priort".

I'm also not sure I understand the line about needing to generate client certs first. Can you clarify?

playbooks/roles/certificates/defaults/main.yml
```diff
@@ -0,0 +1,14 @@
...
+tls_key_province: "California"
+tls_key_city:     "Beverly Hills"
+tls_key_org:      "ACME CORPORATION"
+tls_key_ou:       "Anvil Department"
+tls_days_valid:   "1825"
+tls_default_md:   "sha256"
+tls_key_size:     "4096"
+
+# What type of certificates to be generated
+# must be explicitly set
```
cpu: "must be explicitly set" -> "must be explicitly set by playbooks that include the certificates role"?

documentation/certificates.md
alimakki: yup sounds better

alimakki:
All done!

PR #1002, opened 2017-10-15
Fix the default-site ssh_private_key filepath.

cpu:
Previous to this commit the `default-site.yml` file that was used to
initialize the user's `~/.streisand/site.yml` used `~/.ssh/id-rsa` as
the default SSH private key file location if one were not specified.

`ssh-keygen` actually uses `~/.ssh/id_rsa` (note the underscore) and so
we should too. This commit updates the default
`streisand_ssh_private_key` value from `default-site.yml` accordingly.

Resolves https://github.com/StreisandEffect/streisand/issues/1001


cpu:
@alimakki @nopdotcom Thanks! I will merge after CI finishes.

PR #1003, opened 2017-10-15
Only generate Mirror page when there is mirrored software.

cpu:
If no roles are enabled that mirror client software then the index page
shouldn't reference a mirrored software page and we should not generate
one.

Resolves https://github.com/StreisandEffect/streisand/issues/961


cpu:
Looking at this diff, I should find a way to not duplicate the conditional in three places and instead have a `streisand_did_mirror` or something var that can be conditionally set once and referenced multiple times. 
I'll apply a rev-needed label for now.

cpu:
Much better after 8323ae1 - should be ready for review now.

PR #1005, opened 2017-10-16
Disable duplicate-cn for OpenVPN

alimakki:
Since Streisand now allows defining how many VPN clients are desired, this should be the preferred mechanism by which client profiles should be handed out.

Additionally, it also servers to make the end-user more conscious of who these profiles are handed out to.

This type of setting is more consistent with how OpenConenct is configured and how WireGuard behaves.

Should a user desire the classic behavior, they are always free to uncomment the setting in question.


cpu:
@alimakki I'm close to having an end-to-end integration test of the Streisand OpenVPN profiles finished. I'm selfishly waiting on reviewing this so I can try using it before and after. I'll try to get to this before the end of the week.

Issue #990, opened 2017-10-13
L2TP connection does not complete on Android


nopdotcom:
### Expected behavior:

By following the Android instructions in the gateway, I connect to my streisand server using L2TP.

### Actual Behavior:

On Android, the connection never completes; the VPN tab shows "Unsuccessful" after a minute or two.

Android, determined to make L2TP unfriendly, treats the PSK field as a password field with no "view" option. Great move.

### Ansible Information

* Ansible version: 2.3.1.0
* Ansible system: Linux
* Host OS: Ubuntu
* Host OS version:  16.04
* Python interpreter: python
* Python version: 2.7.12

### Streisand Information

* Streisand Git revision: 66f27ec3befe21ba9e24ab492e536317cfd54e58
* Streisand Git clone has untracked changes: 
* Genesis role: genesis-amazon

### Enabled Roles

* Shadowsocks enabled:  True
* Wireguard enabled: True
* OpenVPN enabled: True
* stunnel enabled: True
* Tor enabled: True
* Openconnect enabled: True
* TinyProxy enabled: True
* SSH forward user enabled: True
* L2TP enabled: True
* Configured number of VPN clients: 5

----

I have no idea why it says the streisand git clone has untracked changes:

```
$ git status
On branch master
Your branch is up-to-date with 'origin/master'.
nothing to commit, working directory clean

$ git remote -v
origin	https://github.com/StreisandEffect/streisand.git (fetch)
origin	https://github.com/StreisandEffect/streisand.git (push)
```

cpu:
> I have no idea why it says the streisand git clone has untracked changes:

It doesn't, `Streisand Git clone has untracked changes:` would be `Streisand Git clone has untracked changes: true` if so. I think I just didn't template the negative case well :-)

nopdotcom:
I have been having random issues with this and can't reproduce reliably. I'm putting this on the back burner until it bites someone else. I'm leaving it open, because we need to test against this, at least during the next testing spree.

If you have this bug, please comment.

genewitch:
I'm having this bug, android won't connect over cellular, and windows won't connect over DSL (win10 default VPN client) - i checked android to make sure it wasn't my VPN passthrough on my router.

I installed onto "existing server" using default settings; ubuntu 16.04.3. The existing server is in a datacenter and shouldn't have any restrictions (like NAT or whatever)

genewitch:
[...]

PR #1006, opened 2017-10-17
Clarify whether or not there are untracked changes

MartinLeedotOrg:
Refer to issue #990 for an instance where this has caused confusion.


MartinLeedotOrg:
I think this is because we're using git diff-index instead of git status.

```
$ git status
On branch master
Your branch is up-to-date with 'origin/master'.
nothing to commit, working tree clean
$ git diff-index  HEAD
$ touch foo
$ git status
On branch master
Your branch is up-to-date with 'origin/master'.
Untracked files:
  (use "git add <file>..." to include in what will be committed)

    foo

nothing added to commit but untracked files present (use "git add" to track)
$ git diff-index HEAD
$ git add foo
$ git diff-index HEAD
:000000 100644 0000000000000000000000000000000000000000 e69de29bb2d1d6434b8b29ae775ad8c2e48c5391 A  foo
```

cpu:
> I think this is because we're using git diff-index instead of git status.

@MartinLeedotOrg Ahhh! That makes sense! Thanks for explaining. I understand now why an untracked file doesn't flip the status (and agree that it shouldn't). :+1:

cpu:
Thanks again @MartinLeedotOrg :trophy:

PR #1008, opened 2017-10-21
Restore sshuttle as optional role with separate user.

### LINKED ISSUE #993: Modular Streisand: optional sshuttle role with dedicated user
### Expected behavior:
[sshuttle](https://github.com/sshuttle/sshuttle) should be available as a tunnelling method

### Actual Behavior:
sshuttle support has been removed by commit aa69d23

### Reason for request:
[sshuttle](https://github.com/sshuttle/sshuttle) is a very useful tunnelling method, with the most minimal requirements (no specialised VPN client software required, only SSH).

It is also most reliable in my experience. I've had issues with pretty much every VPN and debugging is usually a PITA, while sshuttle just works.

Could support be restored please?

cpu:
This is largely a revert of the portions of aa69d23 that removed
`sshuttle` in order to lock down the ssh forward user `forward`.

Since `sshuttle` doesn't require much from the server-side and to
provide a compromise for users still wanting `sshuttle` this commit
re-introduces `sshuttle` as an optional role that defaults to disabled.

If enabled a separate `sshuttle` user is added to the system and
provided full shell access (but no root). Client software is also mirrored
and added to the index.

The validation role is updated to ensure that if sshuttle support is
requested that the ssh_forward support is also enabled. It's possible to
break `sshuttle` out of the `ssh-forward` playbook into its own as
a follow-up since `sshuttle` doesn't rely on the `foward` user anymore.

Resolves #993


documentation/SOURCES.md
```diff
@@ -42,6 +42,7 @@
 - Windows
   - [OpenVPN (from build.openvpn.net)](https://github.com/StreisandEffect/streisand/blob/master/playbooks/roles/openvpn/vars/mirror.yml)
   - [Shadowsocks (from github.com/shadowsocks)](https://github.com/StreisandEffect/streisand/blob/master/playbooks/roles/shadowsocks/vars/mirror.yml)
+  - [Shuttle (from github.com/sshuttle)](https://github.com/jlund/streisand/blob/master/playbooks/roles/streisand-mirror/vars/ssh.yml)
```
gimoh: streisand URL not up to date here

playbooks/roles/ssh-forward/tasks/main.yml
```diff
@@ -14,11 +14,40 @@
...
+
+- name: Register the sshuttle user's public SSH key
+  command: cat {{ sshuttle_location }}/.ssh/id_rsa.pub
+  register: sshuttle_forward_key
+  changed_when: False
+  when: streisand_sshuttle_enabled
+
+- name: Authorize the sshuttle users's key for accessing the sshuttle user
+  authorized_key:
+    user: forward
```
gimoh: This should be `sshuttle` I believe :)

documentation/SOURCES.md
cpu: @gimoh Good catch, thank you!

playbooks/roles/ssh-forward/tasks/main.yml
cpu: You're 100% correct. Thanks @gimoh !

cpu:
Thanks for the review @gimoh - I've addressed all of your feedback as of dad7ea4.

PR #1009, opened 2017-10-22
Fix Vagrant launch for streisand-host.

exratione:
This fixes the failure of `vagrant up streisand-host` by creating a throwaway key pair in the virtual machine at the location specified in `{{ streisand_ssh_private_key }}`.

Since `vagrant ssh streisand-host` uses an entirely different key, it shouldn't be a problem to have this key resident in the VM only. It seems only really needed to run the provisioning.

Note that `vagrant up` still fails on another separate issue: streisand-host now completes successfully, but streisand-client fails. That is outside the scope of this PR.


cpu:
Thanks for the PR!

> but streisand-client fails. That is outside the scope of this PR.

I've been cleaning this up in https://github.com/cpu/streisand/tree/cpu-e2e-tests - the Wireguard test in master is broken.

PR #1013, opened 2017-10-24
Harden Nginx

pguizeline:
Generate a stronger DH

### DIFF
--- a/playbooks/roles/diffie-hellman-group/defaults/main.yml
+++ b/playbooks/roles/diffie-hellman-group/defaults/main.yml
@@ -1,2 +1,2 @@
 ---
-dhparams_size: "2048"
+dhparams_size: "4096"


nopdotcom:
This is blowing out on Travis because it takes more than ten minutes there. Perhaps we should make it conditional on not-CI.

alimakki:
We used to do 3072 but DH but had to bump it down to 2048 because of the time it takes on small VM instances and/or crashing them outright.

pguizeline:
It's possible to install haveged to speed it up a little bit, but you are correct 4096 bits takes some time, I think it's worth it, specially when you consider Streisand users with high treat models.

pguizeline:
Leaving to the user to manually generate the key afterwards could generate some trouble, broken deployments or even the user not running the commands afterwards. Maybe we could create a one time script that re-generates the key silently after deployment?

nopdotcom:
Another possibility is to ask the user if they'd like to use an existing dhparam when deploying.

pguizeline:
That's an option too! But it won't work for remote deployments ie. DO, EC2, Azure.

nopdotcom:
Doing a localhost deploy inside of a tmux/screen session is probably the safest place to generate that 4096-bit dhparam. For remote deploy, once `./streisand` starts provisioning, we're vulnerable to network failure. I had a 15-hour `dhparam 4096` run last night on an EC2 t2.micro instance. (Inside tmux of course. :‐)

If the user can pre-gen the key on their laptop, we don't have a risk of $failure_mode during that time. We could push the file during provisioning; it could be an additional customization option. (More likely, an environment variable.)

cpu:
Hi @pguizeline, thanks again for the PR but I don't think this is a good fit for Streisand.

2048 bit dhparams are already sufficiently strong. Favouring ECDHE ciphersuites that don't rely on the static dhparams file is a better alternative than cranking the dhparams bitsize up at the cost of performance & deploy time.

PR #1015, opened 2017-10-24
Harden Nginx

pguizeline:
Switch ECDSA for RSA for compatibility and generate a strong RSA key.


playbooks/roles/lets-encrypt/templates/response.yaml.j2
```diff
@@ -6,5 +6,5 @@
 "acmetool-quickstart-install-cronjob": true
 "acmetool-quickstart-install-haproxy-script": false
 "acmetool-quickstart-install-redirector-systemd": true
-"acmetool-quickstart-key-type": ecdsa
-"acmetool-quickstart-ecdsa-curve": nistp256
+"acmetool-quickstart-key-type": rsa
+"acmetool-quickstart-rsa-size": 4096
```
cpu: I agree that we should favour RSA certificates over ECDSA for compatibility (missed that in my own review!) but I do not think using 4096 bit RSA keys over 2048 bit RSA keys is sensible (the existing self-signed certificate using 4096 is from before my time).

If you update `acme-quickstart-rsa-size` to `2048` I'm happy to merge :+1:

PR #1016, opened 2017-10-24
Harden Nginx

pguizeline:
Use a strong RSA key


cpu:
Hi @pguizeline - Thanks for the PR but I don't think this is a good fit for Streisand.

3072 bits is already a strong RSA key. 4096 doesn't offer any compelling security advantages and will certainly decrease performance.

PR #1017, opened 2017-10-24
Harden Nginx ciphers and headers

pguizeline:
Use only modern ciphers and try to be compatible with old browsers, add headers for improved security, CSP should be added if possible. Used [Qualys SSL Labs](https://www.ssllabs.com/ssltest/index.html) and [SecurityHeaders.io](https://securityheaders.io/). Based on [Cipherli.st](https://cipherli.st/)


playbooks/roles/streisand-gateway/templates/vhost.j2
```diff
@@ -9,16 +9,32 @@ server {
...
+  ssl_session_cache shared:SSL:10m;
+  ssl_session_tickets off;
+  ssl_stapling on;
+  ssl_stapling_verify on;
 
-  ssl_protocols TLSv1 TLSv1.1 TLSv1.2;
-  ssl_ciphers 'ECDHE-ECDSA-CHACHA20-POLY1305:ECDHE-RSA-CHACHA20-POLY1305:ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-RSA-AES128-GCM-SHA256:ECDHE-ECDSA-AES256-GCM-SHA384:ECDHE-RSA-AES256-GCM-SHA384:DHE-RSA-AES128-GCM-SHA256:DHE-RSA-AES256-GCM-SHA384:ECDHE-ECDSA-AES128-SHA256:ECDHE-RSA-AES128-SHA256:ECDHE-ECDSA-AES128-SHA:ECDHE-RSA-AES256-SHA384:ECDHE-RSA-AES128-SHA:ECDHE-ECDSA-AES256-SHA384:ECDHE-ECDSA-AES256-SHA:ECDHE-RSA-AES256-SHA:DHE-RSA-AES128-SHA256:DHE-RSA-AES128-SHA:DHE-RSA-AES256-SHA256:DHE-RSA-AES256-SHA:ECDHE-ECDSA-DES-CBC3-SHA:ECDHE-RSA-DES-CBC3-SHA:EDH-RSA-DES-CBC3-SHA:AES128-GCM-SHA256:AES256-GCM-SHA384:AES128-SHA256:AES256-SHA256:AES128-SHA:AES256-SHA:DES-CBC3-SHA:!DSS';
-  ssl_prefer_server_ciphers on;
+  # resolver
+  resolver 8.8.8.8 8.8.4.4 valid=300s;
```
alimakki: We should set the resolver to the upstream dns servers set in the [global_vars](https://github.com/StreisandEffect/streisand/blob/master/global_vars/vars.yml) file

pguizeline: Ok! So it should resolv to local? The Nginx resolver is required for OCSP to work I think.

playbooks/roles/streisand-gateway/templates/vhost.j2
```diff
@@ -9,23 +9,43 @@ server {
...
 
+  ssl_protocols TLSv1.2;
+  ssl_prefer_server_ciphers on;
   ssl_dhparam {{ dhparams_location }};
+  ssl_ciphers ECDHE-RSA-AES256-GCM-SHA512:DHE-RSA-AES256-GCM-SHA512:ECDHE-RSA-AES256-GCM-SHA384:DHE-RSA-AES256-GCM-SHA384:ECDHE-RSA-AES256-SHA384:ECDHE-RSA-AES256-SHA:DHE-RSA-AES256-SHA;
+  ssl_ecdh_curve secp384r1;
+  ssl_session_timeout  10m;
+  ssl_session_cache shared:SSL:10m;
+  ssl_session_tickets off;
+  ssl_stapling on;
```
cpu: I don't think there's much value in setting `ssl_stapling` and `ssl_stapling_verify` for a self-signed PKI.  Can you scope this addition to the mode in which a Let's Encrypt certificate has been issued?

playbooks/roles/streisand-gateway/templates/vhost.j2
```diff
@@ -9,23 +9,43 @@ server {
...
+  ssl_stapling_verify on;
```
cpu: I believe this also requires setting `ssl_trusted_certificate`.

playbooks/roles/streisand-gateway/templates/vhost.j2
```diff
@@ -9,23 +9,43 @@ server {
   ssl_certificate      {{ nginx_self_signed_certificate }};
   ssl_certificate_key  {{ nginx_private_key }};
 {% endif %}
-  add_header Strict-Transport-Security max-age=631138519;
 
   auth_basic            "Authorization Required";
   auth_basic_user_file  {{ streisand_gateway_htpasswd_file }};
 
+  ssl_protocols TLSv1.2;
```
cpu: I would prefer to see this target Mozilla's "intermediate" profile for `ssl_protocols`, leaving TLS 1.0 and 1.1 enabled for the time being. I don't have a sense of how many HTTPS clients are likely to be affected by dropping TLS 1.0/1.1 and we should be conservative about breaking compatibility.

playbooks/roles/streisand-gateway/templates/vhost.j2
```diff
@@ -9,23 +9,43 @@ server {
...
+  ssl_prefer_server_ciphers on;
   ssl_dhparam {{ dhparams_location }};
+  ssl_ciphers ECDHE-RSA-AES256-GCM-SHA512:DHE-RSA-AES256-GCM-SHA512:ECDHE-RSA-AES256-GCM-SHA384:DHE-RSA-AES256-GCM-SHA384:ECDHE-RSA-AES256-SHA384:ECDHE-RSA-AES256-SHA:DHE-RSA-AES256-SHA;
```
cpu: Ditto RE: targetting the "intermediate" profile and slightly favouring compatibility. I also don't see any of the CHACHA-POLY1305 ciphersuites here.

playbooks/roles/streisand-gateway/templates/vhost.j2
```diff
@@ -9,23 +9,43 @@ server {
...
+  ssl_ecdh_curve secp384r1;
```
cpu: I don't think there's a convincing argument to use this over `secp256r1`.

playbooks/roles/streisand-gateway/templates/vhost.j2
```diff
@@ -9,23 +9,43 @@ server {
...
 
-  ssl_protocols TLSv1 TLSv1.1 TLSv1.2;
-  ssl_ciphers 'ECDHE-ECDSA-CHACHA20-POLY1305:ECDHE-RSA-CHACHA20-POLY1305:ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-RSA-AES128-GCM-SHA256:ECDHE-ECDSA-AES256-GCM-SHA384:ECDHE-RSA-AES256-GCM-SHA384:DHE-RSA-AES128-GCM-SHA256:DHE-RSA-AES256-GCM-SHA384:ECDHE-ECDSA-AES128-SHA256:ECDHE-RSA-AES128-SHA256:ECDHE-ECDSA-AES128-SHA:ECDHE-RSA-AES256-SHA384:ECDHE-RSA-AES128-SHA:ECDHE-ECDSA-AES256-SHA384:ECDHE-ECDSA-AES256-SHA:ECDHE-RSA-AES256-SHA:DHE-RSA-AES128-SHA256:DHE-RSA-AES128-SHA:DHE-RSA-AES256-SHA256:DHE-RSA-AES256-SHA:ECDHE-ECDSA-DES-CBC3-SHA:ECDHE-RSA-DES-CBC3-SHA:EDH-RSA-DES-CBC3-SHA:AES128-GCM-SHA256:AES256-GCM-SHA384:AES128-SHA256:AES256-SHA256:AES128-SHA:AES256-SHA:DES-CBC3-SHA:!DSS';
-  ssl_prefer_server_ciphers on;
+  # resolver
+  resolver 8.8.8.8 8.8.4.4 valid=300s;
+  resolver_timeout 5s;
+
+  # headers
+  add_header Strict-Transport-Security "max-age=63072000; includeSubDomains; preload";
```
cpu: Can you remove `includeSubDomains` and `preload`? They don't make sense for Streisand.

playbooks/roles/streisand-gateway/templates/vhost.j2
pguizeline: I think that when using a LE cert, the fullchain.pem already includes ca-certificate.pem, but if you wan't I can add it either way.

playbooks/roles/streisand-gateway/templates/vhost.j2
pguizeline: You're right!

playbooks/roles/streisand-gateway/templates/vhost.j2
cpu: The [nginx docs for `ssl_stapling_verify`](http://nginx.org/en/docs/http/ngx_http_ssl_module.html#ssl_stapling_verify) specifically say:
> For verification to work, the certificate of the server certificate issuer, the root certificate, and all intermediate certificates should be configured as trusted using the ssl_trusted_certificate directive.

I believe it's not sufficient to have a full chain in the `ssl_certificate` file directive.

playbooks/roles/streisand-gateway/templates/vhost.j2
```diff
@@ -9,23 +9,43 @@ server {
...
+  ssl_ciphers 'ECDHE-ECDSA-CHACHA20-POLY1305:ECDHE-RSA-CHACHA20-POLY1305:ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-RSA-AES128-GCM-SHA256:ECDHE-ECDSA-AES256-GCM-SHA384:ECDHE-RSA-AES256-GCM-SHA384:DHE-RSA-AES128-GCM-SHA256:DHE-RSA-AES256-GCM-SHA384:ECDHE-ECDSA-AES128-SHA256:ECDHE-RSA-AES128-SHA256:ECDHE-ECDSA-AES128-SHA:ECDHE-RSA-AES256-SHA384:ECDHE-RSA-AES128-SHA:ECDHE-ECDSA-AES256-SHA384:ECDHE-ECDSA-AES256-SHA:ECDHE-RSA-AES256-SHA:DHE-RSA-AES128-SHA256:DHE-RSA-AES128-SHA:DHE-RSA-AES256-SHA256:DHE-RSA-AES256-SHA:ECDHE-ECDSA-DES-CBC3-SHA:ECDHE-RSA-DES-CBC3-SHA:EDH-RSA-DES-CBC3-SHA:AES128-GCM-SHA256:AES256-GCM-SHA384:AES128-SHA256:AES256-SHA256:AES128-SHA:AES256-SHA:DES-CBC3-SHA:!DSS';
+  ssl_ecdh_curve secp256r1;
+  ssl_session_timeout  10m;
+  ssl_session_cache shared:SSL:10m;
+  ssl_session_tickets off;
+  ssl_stapling on;
+  ssl_stapling_verify on;
+
+  # resolver
+  resolver 127.0.0.1 valid=300s;
```
cpu: I think when [Ali left his comment about the resolver](https://github.com/StreisandEffect/streisand/pull/1017#discussion_r146736367) on a previous diff he wanted the `127.0.0.1` here to be populated using the `upstream_dns_servers` specified in the common vars for the project wide upstream resolver.

playbooks/roles/streisand-gateway/templates/vhost.j2
```diff
@@ -9,23 +9,43 @@ server {
...
   ssl_prefer_server_ciphers on;
+  ssl_dhparam {{ dhparams_location }};
+  ssl_ciphers 'ECDHE-ECDSA-CHACHA20-POLY1305:ECDHE-RSA-CHACHA20-POLY1305:ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-RSA-AES128-GCM-SHA256:ECDHE-ECDSA-AES256-GCM-SHA384:ECDHE-RSA-AES256-GCM-SHA384:DHE-RSA-AES128-GCM-SHA256:DHE-RSA-AES256-GCM-SHA384:ECDHE-ECDSA-AES128-SHA256:ECDHE-RSA-AES128-SHA256:ECDHE-ECDSA-AES128-SHA:ECDHE-RSA-AES256-SHA384:ECDHE-RSA-AES128-SHA:ECDHE-ECDSA-AES256-SHA384:ECDHE-ECDSA-AES256-SHA:ECDHE-RSA-AES256-SHA:DHE-RSA-AES128-SHA256:DHE-RSA-AES128-SHA:DHE-RSA-AES256-SHA256:DHE-RSA-AES256-SHA:ECDHE-ECDSA-DES-CBC3-SHA:ECDHE-RSA-DES-CBC3-SHA:EDH-RSA-DES-CBC3-SHA:AES128-GCM-SHA256:AES256-GCM-SHA384:AES128-SHA256:AES256-SHA256:AES128-SHA:AES256-SHA:DES-CBC3-SHA:!DSS';
+  ssl_ecdh_curve secp256r1;
+  ssl_session_timeout  10m;
+  ssl_session_cache shared:SSL:10m;
+  ssl_session_tickets off;
+  ssl_stapling on;
```
cpu: I still think that we should try to scope stapling configuration changes to the case where a Let's Encrypt certificate is in use. I'm not sure what nginx does w.r.t stapling for a self-signed PKI without an ocsp-responder.

pguizeline: I understand where you're going, it's probably best to create a rule to implement this part only if LE is selected. The only problem is that ansible is completely alien for me, I'm afraid I'll end up breaking something!

playbooks/roles/streisand-gateway/templates/vhost.j2
pguizeline: I'll see if I can point this to the common vars file

alimakki: I think something like `"{{ upstream_dns_servers | join(' ') }}"` might work.

playbooks/roles/streisand-gateway/templates/vhost.j2
cpu: > The only problem is that ansible is completely alien for me, I'm afraid I'll end up breaking something!

No problem! Happy to provide some more concrete guidance and some pointers to the docs if you're up for trying it :-)

There's a var `le_ok` that's defined when the gateway is using a Let's Encrypt certificate so in this case you can put the configuration options specific to OCSP stapling into a Jinja2 `if` block using the `le_ok` var like:

```
{% if le_ok %}
ssl_stapling on;
etc etc
{% endif %}
```

You might find reading through [the Ansible template documentation](http://docs.ansible.com/ansible/latest/playbooks_templating.html) and the [jinja2 docs](http://jinja.pocoo.org/docs/2.9/templates/) helpful.

playbooks/roles/streisand-gateway/templates/vhost.j2
```diff
@@ -1,31 +1,52 @@
...
   ssl_prefer_server_ciphers on;
+  ssl_dhparam {{ dhparams_location }};
+  ssl_ciphers 'ECDHE-ECDSA-CHACHA20-POLY1305:ECDHE-RSA-CHACHA20-POLY1305:ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-RSA-AES128-GCM-SHA256:ECDHE-ECDSA-AES256-GCM-SHA384:ECDHE-RSA-AES256-GCM-SHA384:DHE-RSA-AES128-GCM-SHA256:DHE-RSA-AES256-GCM-SHA384:ECDHE-ECDSA-AES128-SHA256:ECDHE-RSA-AES128-SHA256:ECDHE-ECDSA-AES128-SHA:ECDHE-RSA-AES256-SHA384:ECDHE-RSA-AES128-SHA:ECDHE-ECDSA-AES256-SHA384:ECDHE-ECDSA-AES256-SHA:ECDHE-RSA-AES256-SHA:DHE-RSA-AES128-SHA256:DHE-RSA-AES128-SHA:DHE-RSA-AES256-SHA256:DHE-RSA-AES256-SHA:ECDHE-ECDSA-DES-CBC3-SHA:ECDHE-RSA-DES-CBC3-SHA:EDH-RSA-DES-CBC3-SHA:AES128-GCM-SHA256:AES256-GCM-SHA384:AES128-SHA256:AES256-SHA256:AES128-SHA:AES256-SHA:DES-CBC3-SHA:!DSS';
+  ssl_ecdh_curve secp256r1;
+  ssl_session_timeout  10m;
+  ssl_session_cache shared:SSL:10m;
+  ssl_session_tickets off;
+
+  # resolver
+  resolver {{ upstream_dns_servers | join(' ') }} valid=300s;
```
cpu: I think `resolver` and `resolver_timeout` are only required when `ssl_stapling` is enabled. Can you move them into the `le_ok == true` clause as well?

playbooks/roles/streisand-gateway/templates/vhost.j2
```diff
@@ -1,31 +1,52 @@
...
 server {
-  listen 127.0.0.1:{{ nginx_port }} default_server ssl;
+  listen 127.0.0.1:{{ nginx_port }} default_server ssl http2;
 
 {% if le_ok %}
-  ssl_certificate      {{ le_certificate }};
-  ssl_certificate_key  {{ le_private_key }};
+  ssl_certificate         {{ le_certificate }};
+  ssl_certificate_key     {{ le_private_key }};
+  ssl_trusted_certificate {{ le_chain }};
```
cpu: `le_chain` isn't defined - I think you should close https://github.com/StreisandEffect/streisand/pull/1028 and add that commit to this PR so everything is self-contained.

cpu:
Hi @pguizeline,

Have you tried running this branch? It's failing in CI trying to restart nginx and I suspect there's something up with the configuration being written to disk:
> RUNNING HANDLER [streisand-gateway : Restart Nginx for the Gateway vhost] ******
fatal: [10.49.117.216]: FAILED! => {"changed": false, "failed": true, "msg": "Unable to start service nginx: Job for nginx.service failed because the control process exited with error code. See \"systemctl status nginx.service\" and \"journalctl -xe\" for details.\n"}

I can try and take a look over the weekend but if you have cycles in the meantime and could investigate that would be great. Thanks!

playbooks/roles/streisand-gateway/templates/vhost.j2
```diff
@@ -1,31 +1,51 @@
...
   ssl_prefer_server_ciphers on;
+  ssl_dhparam {{ dhparams_location }};
+  ssl_ciphers 'ECDHE-ECDSA-CHACHA20-POLY1305:ECDHE-RSA-CHACHA20-POLY1305:ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-RSA-AES128-GCM-SHA256:ECDHE-ECDSA-AES256-GCM-SHA384:ECDHE-RSA-AES256-GCM-SHA384:DHE-RSA-AES128-GCM-SHA256:DHE-RSA-AES256-GCM-SHA384:ECDHE-ECDSA-AES128-SHA256:ECDHE-RSA-AES128-SHA256:ECDHE-ECDSA-AES128-SHA:ECDHE-RSA-AES256-SHA384:ECDHE-RSA-AES128-SHA:ECDHE-ECDSA-AES256-SHA384:ECDHE-ECDSA-AES256-SHA:ECDHE-RSA-AES256-SHA:DHE-RSA-AES128-SHA256:DHE-RSA-AES128-SHA:DHE-RSA-AES256-SHA256:DHE-RSA-AES256-SHA:ECDHE-ECDSA-DES-CBC3-SHA:ECDHE-RSA-DES-CBC3-SHA:EDH-RSA-DES-CBC3-SHA:AES128-GCM-SHA256:AES256-GCM-SHA384:AES128-SHA256:AES256-SHA256:AES128-SHA:AES256-SHA:DES-CBC3-SHA:!DSS';
+  ssl_ecdh_curve secp256r1;
```
cpu: This curve name doesn't work with nginx 1.12.2 and OpenSSL 1.0.2g:
```
# nginx -V
nginx version: nginx/1.12.2
built by gcc 5.4.0 20160609 (Ubuntu 5.4.0-6ubuntu1~16.04.4) 
built with OpenSSL 1.0.2g  1 Mar 2016
TLS SNI support enabled
configure arguments: --prefix=/etc/nginx --sbin-path=/usr/sbin/nginx --modules-path=/usr/lib/nginx/modules --conf-path=/etc/nginx/nginx.conf --error-log-path=/var/log/nginx/error.log --http-log-path=/var/log/nginx/access.log --pid-path=/var/run/nginx.pid --lock-path=/var/run/nginx.lock --http-client-body-temp-path=/var/cache/nginx/client_temp --http-proxy-temp-path=/var/cache/nginx/proxy_temp --http-fastcgi-temp-path=/var/cache/nginx/fastcgi_temp --http-uwsgi-temp-path=/var/cache/nginx/uwsgi_temp --http-scgi-temp-path=/var/cache/nginx/scgi_temp --user=nginx --group=nginx --with-compat --with-file-aio --with-threads --with-http_addition_module --with-http_auth_request_module --with-http_dav_module --with-http_flv_module --with-http_gunzip_module --with-http_gzip_static_module --with-http_mp4_module --with-http_random_index_module --with-http_realip_module --with-http_secure_link_module --with-http_slice_module --with-http_ssl_module --with-http_stub_status_module --with-http_sub_module --with-http_v2_module --with-mail --with-mail_ssl_module --with-stream --with-stream_realip_module --with-stream_ssl_module --with-stream_ssl_preread_module --with-cc-opt='-g -O2 -fstack-protector-strong -Wformat -Werror=format-security -Wp,-D_FORTIFY_SOURCE=2 -fPIC' --with-ld-opt='-Wl,-Bsymbolic-functions -Wl,-z,relro -Wl,-z,now -Wl,--as-needed -pie'

# nginx -t -c /etc/nginx/nginx.conf 
# SSL Gateway
nginx: [emerg] SSL_CTX_set1_curves_list("secp256r1") failed (SSL:)
nginx: configuration file /etc/nginx/nginx.conf test failed
```

I suspect the correct name is "prime256v1".

After [reading the docs for `ssl_ecdh_curve`](http://nginx.org/en/docs/http/ngx_http_ssl_module.html#ssl_ecdh_curve) I think we should actually use the value `"auto"` instead of specifying the curve name ourselves:
> The special value auto (1.11.0) instructs nginx to use a list built into the OpenSSL library when using OpenSSL 1.0.2 or higher, or prime256v1 with older versions.

Changing the value to `auto` fixes the syntax error:
```
# nginx -t -c /etc/nginx/nginx.conf 
nginx: the configuration file /etc/nginx/nginx.conf syntax is ok
nginx: configuration file /etc/nginx/nginx.conf test is successful
```

cpu:
> I can try and take a look over the weekend but if you have cycles in the meantime and could investigate that would be great. Thanks!

I had some time to circle back to this PR and figured out the problem. See [this review comment](https://github.com/StreisandEffect/streisand/pull/1017#discussion_r151841005).

@pguizeline Do you want to make the required change? If I don't hear back in the next few days I will push a fix to the branch myself & merge it to help keep the backlog moving. Thanks!

cpu:
Hi again @pguizeline - it appears I don't have permission to push commits onto this branch of your fork (Did you [allow edits from maintainers](https://help.github.com/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork/) when opening the PR?)

Here's a patch you can apply to your branch to fix the issue I flagged: https://gist.github.com/cpu/935f2d79aee9a09cd231dc5fcd464f77 
You can download that to your Streisand directory with this branch cloned & run `git apply`.

PR #1019, opened 2017-10-24
README updates

alimakki:
Starting with Ansible 2.4, the Azure pip [requirements](http://docs.ansible.com/ansible/latest/guide_azure.html) have been updated. This PR documents that change as part of the README instructions to set up cloud dependencies.


cpu:
Good catch @alimakki !

PR #1022, opened 2017-10-26
Set up a simple noninteractive deployment mode.

exratione:
This PR is a proposed simple noninteractive deployment.

- Make a few changes to Ansible where needed to allow some prompts to be bypassed.
- Provide example vars files.
- Create some additional bash scripts to run non-interactive deployments.
- Remove duplicate function definitions from the `streisand` script.
- Add English documentation on running the new scripts.

It is possible, somewhat, to shoehorn all of the new bash scripting into the main `streisand` script, but it gets pretty ugly with if-blocks and checks if you do that. I'm not in favor of kitchen sink scripts - the code is much simpler and more maintainable when split out in this way.

Example configuration files are found under `global_vars/noninteractive`. Copy and edit the desired parameters, such as providing API tokens and other choices, and then run the appropriate script.

To deploy a new Streisand server:

```
 deploy/streisand-new-cloud-server \
  --provider digitalocean \
  --site-config global_vars/noninteractive/digitalocean-site.yml
```

To run the Streisand provisioning on the local machine:

```
deploy/streisand-local \
   --site-config global_vars/noninteractive/local-site.yml
```

To rerun the Streisand provisioning an existing server:

```
deploy/streisand-existing-cloud-server \
  --ip-address 10.10.10.10 \
  --ssh-user root \
  --site-config global_vars/noninteractive/local-site.yml
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -185,6 +185,34 @@ The server must be accessible using the `$HOME/id_rsa` SSH Key, and **root** is 
 
 **Note:** Running Streisand against an existing server can be a destructive action! You will be potentially overwriting configuration files and must be certain that you are affecting the correct machine.
 
+### Noninteractive Deployment (Advanced) ###
+
+Alternative scripts and configuration file examples are provided for
+noninteractive deployment, in which all of the required information is passed
+on the command line or in a configuration file.
+
+Example configuration files are found under `global_vars/noninteractive`. Copy
+and edit the desired parameters, such as providing API tokens and other choices,
+and then run the appropriate script.
+
+To deploy a new Streisand server:
+
+      deploy/streisand-new-cloud-server.sh \
+        --provider digitalocean \
+        --site-config global_vars/noninteractive/digitalocean-site.yml
+
+To run the Streisand provisioning on the local machine:
+
+      deploy/streisand-local.sh \
+        --site-config global_vars/noninteractive/local-site.yml
+
+To run the Streisand provisioning against an existing server:
+
+      deploy/streisand-existing-cloud-server.sh \
+        --ip-address 10.10.10.10 \
+        --ssh-user root \
+        --site-config global_vars/noninteractive/digitalocean-site.yml
+
 Upcoming Features
 -----------------
 * Easier setup.
--- a/deploy/streisand-existing-cloud-server.sh
+++ b/deploy/streisand-existing-cloud-server.sh
@@ -1,0 +1,116 @@
+#!/usr/bin/env bash
+#
+# Provision an existing cloud server.
+#
+# This requires an expanded extra-vars file specific to the provider type that
+# sets all of the values gathered by prompts in the interactive installation.
+# See the contents of global_vars/noninteractive for examples that can be copied
+# and modified.
+#
+# Usage:
+#
+# streisand-existing-cloud-server \
+#   --ssh-user root \
+#   --ip-address 10.10.10.10 \
+#   --site-config path/to/digitalocean-site.yml
+#
+
+set -o errexit
+set -o nounset
+
+DIR="$( cd "$( dirname "$0" )" && pwd)"
+PROJECT_DIR="${DIR}/.."
+
+export DEFAULT_SITE_VARS="${PROJECT_DIR}/global_vars/default-site.yml"
+export GLOBAL_VARS="${PROJECT_DIR}/global_vars/vars.yml"
+
+# Include the check_ansible function from ansible_check.sh.
+# shellcheck source=util/source_validate_and_deploy.sh
+source "${PROJECT_DIR}/util/ansible_check.sh"
+check_ansible
+
+# --------------------------------------------------------------------------
+# Reading options.
+# --------------------------------------------------------------------------
+
+function usage () {
+  cat <<EOF
+Usage:
+$0 \\
+  --ssh-user root \\
+  --ip-address 10.10.10.10 \\
+  --site-config path/to/site.yml
+
+If no SSH user is specified, then the root user will be used.
+
+If no configuration file is specified, then ~/.streisand/site.yml will be used
+if it exists, or global_vars/default-site.yml will be used otherwise.
+EOF
+}
+
+SSH_USER=""
+IP_ADDRESS=""
+SITE_VARS=""
+
+while [[ ${#} -gt 0 ]]; do
+  case "${1}" in
+    # Required.
+    --ip-address)    IP_ADDRESS="${2}"; shift;;
+
+    # Optional.
+    --site-config)   SITE_VARS="${2}"; shift;;
+    --ssh-user)      SSH_USER="${2}"; shift;;
+
+    # Utility.
+    -h|--help)       usage; exit 0;;
+    --)              break;;
+    -*)              echo "Unrecognized option ${1}"; usage; exit 1;;
+  esac
+
+  shift
+done
+
+# --------------------------------------------------------------------------
+# Fail if required options are not set.
+# --------------------------------------------------------------------------
+
+if [ -z "${IP_ADDRESS}" ]; then
+  usage
+  exit 1
+fi
+
+# --------------------------------------------------------------------------
+# Default if no parameter is provided.
+# --------------------------------------------------------------------------
+
+if [ -z "${SSH_USER}" ]; then
+  echo "Defaulting to SSH user: root"
+  SSH_USER="root"
+fi
+
+# --------------------------------------------------------------------------
+# Sort out the SITE_VARS value based on input and defaults.
+# --------------------------------------------------------------------------
+
+# shellcheck source=util/source_check_and_default_site_vars.sh
+source "${PROJECT_DIR}/util/source_check_and_default_site_vars.sh"
+
+# --------------------------------------------------------------------------
+# Onwards to launch and provision the server.
+# --------------------------------------------------------------------------
+
+export INVENTORY="${PROJECT_DIR}/inventories/inventory-existing"
+
+# Create an inventory file on the fly.
+cat > "${INVENTORY}" <<EOF
+[localhost]
+localhost ansible_connection=local ansible_python_interpreter=python
+[streisand-host]
+${IP_ADDRESS} ansible_user=${SSH_USER}
+EOF
+
+export PLAYBOOK="${PROJECT_DIR}/playbooks/existing-server.yml"
+
+# Run the validation and deployment. This expects the variables set here.
+# shellcheck source=util/source_validate_and_deploy.sh
+source "${PROJECT_DIR}/util/source_validate_and_deploy.sh"
--- a/deploy/streisand-local.sh
+++ b/deploy/streisand-local.sh
@@ -1,0 +1,75 @@
+#!/usr/bin/env bash
+#
+# Run a noninteractive Streisand installation on the local machine.
+#
+# This requires an expanded extra-vars file specific to the provider type that
+# sets all of the values gathered by prompts in the interactive installation.
+# See the contents of global_vars/noninteractive for examples that can be copied
+# and modified.
+#
+# Usage:
+#
+# streisand-local --site-config path/to/local-site.yml
+#
+
+set -o errexit
+set -o nounset
+
+DIR="$( cd "$( dirname "$0" )" && pwd)"
+PROJECT_DIR="${DIR}/.."
+
+export DEFAULT_SITE_VARS="${PROJECT_DIR}/global_vars/default-site.yml"
+export GLOBAL_VARS="${PROJECT_DIR}/global_vars/vars.yml"
+
+# Include the check_ansible function from ansible_check.sh.
+# shellcheck source=util/source_validate_and_deploy.sh
+source "${PROJECT_DIR}/util/ansible_check.sh"
+check_ansible
+
+# --------------------------------------------------------------------------
+# Reading options.
+# --------------------------------------------------------------------------
+
+function usage () {
+  cat <<EOF
+Usage:
+$0 --site-config path/to/site.yml
+
+If no configuration file is specified, then ~/.streisand/site.yml will be used
+if it exists, or global_vars/default-site.yml will be used otherwise.
+EOF
+}
+
+SITE_VARS=""
+
+while [[ ${#} -gt 0 ]]; do
+  case "${1}" in
+    # Optional.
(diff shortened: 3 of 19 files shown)


nopdotcom:
In general, I think it's a better user interface to be dropped into $EDITOR for an options file than to be prompted interactively.

Our data needs look like this. That connection between the provider question and provider options is just as awkward as it looks.

capto_capture 2017-10-26_12-59-27_pm

nopdotcom:
To be clear: I think we have a problem, and heading your direction sounds like where we need to go.

cpu:
> In general, I think it's a better user interface to be dropped into $EDITOR for an options file than to be prompted interactively.

@nopdotcom I'm not sure I agree - can you expand on your rationale for preferring `$EDITOR`?

global_vars/noninteractive/amazon-site.yml
```diff
@@ -0,0 +1,68 @@
...
+#
+# Copy this and edit it as needed before running streisand-new-cloud-server.
+#
+
+streisand_noninteractive: true
+confirmation: true
+
+# The SSH private key that Ansible will use to connect to the Streisand node.
+#
+# The corresponding public key must be added to the Digital Ocean control panel
```
cpu: This paragraph refers to Digital Ocean but it's an Amazon config example.

global_vars/noninteractive/amazon-site.yml
```diff
@@ -0,0 +1,68 @@
...
+# 7. EU             (Frankfurt)
+# 8. EU             (Ireland)
+# 9. EU             (London)
+# 10. South America (Sao Paulo)
+# 11. US East       (Northern Virginia)
+# 12. US East       (Ohio)
+# 13. US West       (Northern California)
+# 14. US West       (Oregon)
+#
+# Note that this must be a string representation of a number, not a number.
```
cpu: Perhaps clearer expressed as 'Note: aws_region_var must be a number in quotes, e.g. "3" not 3.'

cpu: The other defaults files should probably have their `region_var` comments updated for clarity as well.

global_vars/noninteractive/amazon-site.yml
cpu: The other defaults files (minus the actual Digital Ocean defaults file) should have this comment fixed as well.

streisand-existing-cloud-server
```diff
@@ -0,0 +1,103 @@
...
+#
+# Provision an existing cloud server.
+#
+# This requires an expanded extra-vars file specific to the provider type that
+# sets all of the values gathered by prompts in the interactive installation.
+# See the contents of global_vars/noninteractive for examples that can be copied
+# and modified.
+#
+# Usage:
+# $0 --provider digitalocean --site-config path/to/digitalocean-site.yml
```
cpu: I think it would be clearer to replace `$0` with the current script filename since these comments are static and not something that can expand vars dynamically.

exratione:
If you're good with the existence of multiple bash scripts, I can DRY them up a bit as suggested.

No objection to the `streisand-local` change to default to a file in `~/.streisand/site.yml`.

Other suggests all seem reasonable, or catch my cut-paste errors.

I put repetition into the vars files because you really want a person just to be able to pull the one file for their scenario that lists all the options right there. They serve the dual purpose of docs and utility. If the individual also has to go digging around elsewhere for other options, that is less friendly. Something of a trade-off with no good answers, I recognize.

exratione:
Updated with some moving, refactoring, and correcting.

playbooks/roles/privoxy/tasks/main.yml
```diff
@@ -0,0 +1,4 @@
+---
+- name: Install Privoxy
+  apt:
+    name: privoxy
```
alimakki: Is privoxy necessary for the scope of the changes?

exratione: Oops. I was tinkering with this, and seems I crossed the streams.

cpu:
> I put repetition into the vars files because you really want a person just to be able to pull the one file for their scenario that lists all the options right there. They serve the dual purpose of docs and utility. If the individual also has to go digging around elsewhere for other options, that is less friendly. Something of a trade-off with no good answers, I recognize.

That's fair. I think it's a reasonable trade-off to have to maintain some duplication in the effort of making these self-contained. Let's try it out and see whether it becomes a sticking point later :+1:

cpu:
This is at the top of my list to review again over the next couple of days. :mag: :watch:

nopdotcom:
"🔍⌚️" sounds like you're reviewing the displays of smart watches.

I'll 🔍 at this ⌚️ too. I need to write the manifesto for $EDITOR....
-- 

-- 
Jay

README.md
```diff
@@ -185,6 +185,34 @@ The server must be accessible using the `$HOME/id_rsa` SSH Key, and **root** is
...
+noninteractive deployment, in which all of the required information is passed
+on the command line or in a configuration file.
+
+Example configuration files are found under `global_vars/noninteractive`. Copy
+and edit the desired parameters, such as providing API tokens and other choices,
+and then run the appropriate script.
+
+To deploy a new Streisand server:
+
+      deploy/streisand-new-cloud-server \
```
cpu: Can you add a `.sh` extension to the shell scripts in the `deploy/` directory? That's how CI knows to run shellcheck. I admit it's a little but ugly compared to a bare path but I think it's a fair trade off to get bash scripts linted easily.

These example invocations in the README will have to change as well as the file names.

Thanks!

README.md
```diff
@@ -185,6 +185,34 @@ The server must be accessible using the `$HOME/id_rsa` SSH Key, and **root** is
...
+        --provider digitalocean \
+        --site-config global_vars/noninteractive/digitalocean-site.yml
+
+To run the Streisand provisioning on the local machine:
+
+      deploy/streisand-local \
+        --site-config global_vars/noninteractive/local-site.yml
+
+To rerun the Streisand provisioning an existing server:
```
cpu: "rerun" should be "run". Streisand doesn't really support re-provisioning a server at this point. The existing server mode assumes a fresh 16.04 install like the cloud provisioning modes provide.

deploy/streisand-existing-cloud-server
```diff
@@ -0,0 +1,99 @@
...
+# Provision an existing cloud server.
+#
+# This requires an expanded extra-vars file specific to the provider type that
+# sets all of the values gathered by prompts in the interactive installation.
+# See the contents of global_vars/noninteractive for examples that can be copied
+# and modified.
+#
+# Usage:
+# streisand-existing-cloud-server \
+#   --provider digitalocean \
```
cpu: I don't think the `existing-cloud-server` should take a `--provider` arg for digitalocean. Copy pasta?

deploy/streisand-existing-cloud-server
```diff
@@ -0,0 +1,99 @@
...
+#   --site-config path/to/digitalocean-site.yml
```
cpu: I think this needs to also include `--ip-address` and `--ssh-user` to be a complete example for localhost noninteractive provisioning.

deploy/streisand-local
```diff
@@ -0,0 +1,91 @@
...
+# See the contents of global_vars/noninteractive for examples that can be copied
+# and modified.
+#
+# Usage:
+#
+# streisand-local
+#
+# or
+#
+# streisand-local --site-config path/to/digitalocean-site.yml
```
cpu: `global_vars/noninteractive/local-site.yml ` not digitalocean-site.yml ?

deploy/streisand-new-cloud-server
```diff
@@ -0,0 +1,95 @@
...
+# Run a noninteractive Streisand installation that creates a new cloud server.
+#
+# This requires an expanded extra-vars file specific to the provider type that
+# sets all of the values gathered by prompts in the interactive installation.
+# See the contents of global_vars/noninteractive for examples that can be copied
+# and modified.
+#
+# Usage:
+# streisand-new-cloud-server \
+#   --provider digitalocean \
```
cpu: It might make sense to express this usage like:
`--provider [amazon|azure|digitalocean|google|linode|rackspace]` to make it clearer that you can use those additional providers.

global_vars/noninteractive/amazon-site.yml
```diff
@@ -0,0 +1,68 @@
...
+aws_vpc_id_var: ""
+aws_vpc_subnet_id_var: ""
+
+aws_instance_name: streisand
+
+# The AWS credentials to use.
+aws_access_key: ""
+aws_secret_key: ""
+
+# Definitions needed for Let's Encrypt SSH certificate setup.
```
cpu: SSH -> HTTPS (or TLS)

global_vars/noninteractive/azure-site.yml
```diff
@@ -0,0 +1,89 @@
...
+#   32: Germany Northeast   (Magdeburg)
+#   33: UK West             (Cardiff)
+#   34: UK South            (London)
+#
+# Note: azure_region_var must be a number in quotes, e.g. "1" not 1.
+azure_region_var: "1"
+
+azure_instance_name_var: streisand
+
+# Definitions needed for Let's Encrypt SSH certificate setup.
```
cpu: SSH -> HTTPS (or TLS)

global_vars/noninteractive/google-site.yml
```diff
@@ -0,0 +1,79 @@
...
+gce_zone_var: "3"
+
+gce_server_name: streisand
+
+# The full path of your unique service account credentials file. See:
+# https://docs.ansible.com/ansible/guide_gce.html#credentials
+# https://support.google.com/cloud/answer/6158849?hl=en&ref_topic=6262490#serviceaccounts
+gce_json_file_location: ""
+
+# Definitions needed for Let's Encrypt SSH certificate setup.
```
cpu: SSH -> HTTPS (or TLS)

global_vars/noninteractive/linode-site.yml
```diff
@@ -0,0 +1,55 @@
...
+#
+# Note: linode_datacenter must be a number in quotes, e.g. "7" not 7.
+linode_datacenter: "7"
+
+linode_server_name: streisand
+
+# Obtain the API key from the Linode Manager console.
+linode_api_key: ""
+
+# Definitions needed for Let's Encrypt SSH certificate setup.
```
cpu: SSH -> HTTPS (or TLS)

global_vars/noninteractive/local-site.yml
```diff
@@ -0,0 +1,37 @@
...
+streisand_ssh_forward_enabled: yes
+# By default sshuttle is disabled because it creates a `sshuttle` user that has
+# full shell privileges on the Streisand host
+streisand_sshuttle_enabled: no
+streisand_stunnel_enabled: yes
+streisand_tinyproxy_enabled: yes
+streisand_tor_enabled: yes
+streisand_wireguard_enabled: yes
+
+# Definitions needed for Let's Encrypt SSH certificate setup.
```
cpu: SSH -> HTTPS (or TLS)

global_vars/noninteractive/rackspace-site.yml
```diff
@@ -0,0 +1,53 @@
...
+# Note: rackspace_region must be a number in quotes, e.g. "1" not 1.
+rackspace_region: "1"
+
+rackspace_server_name: streisand
+
+# Obtain these credentials from the Rackspace Cloud Control Panel.
+rackspace_username: ""
+rackspace_api_key: ""
+
+# Definitions needed for Let's Encrypt SSH certificate setup.
```
cpu: SSH -> HTTPS (or TLS)

util/source_validate_and_deploy.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
```
cpu: Can you make sure to `set -o errexit` for this script?

exratione:
Hmm. Yes, the local install is indeed prompting. I will see if I can figure out why.

exratione:
@cpu Fixed. The problem was bad variable names in the global_vars/noninteractive file.

deploy/streisand-local.sh
```diff
@@ -0,0 +1,93 @@
...
+# --------------------------------------------------------------------------
+# Default if no parameter is provided.
+# --------------------------------------------------------------------------
+
+# Make sure the alleged configuration file exists.
+if [ ! -f "${SITE_VARS}" ]; then
+
+  SITE_VARS="${HOME}/.streisand/site.yml"
+
+  if [ -f "${SITE_VARS}" ]; then
```
cpu: I think this should be `if [ ! -f "${SITE_VARS}" ]; then`. Otherwise running with no `~/.streisand/site.yml` and no `--site-config` results in:
```
# deploy/streisand-local.sh
Using default config file: /root/.streisand/site.yml
 [WARNING]: Found both group and host with same name: localhost

ERROR! Unable to retrieve file contents
Could not find or access '/root/.streisand/site.yml'
```

deploy/streisand-new-cloud-server.sh
```diff
@@ -0,0 +1,97 @@
...
+  usage
+  exit 1
+fi
+
+# --------------------------------------------------------------------------
+# Fail for other reasons.
+# --------------------------------------------------------------------------
+
+# Make sure the alleged configuration file exists.
+if [ ! -f "${SITE_VARS}" ]; then
```
cpu: Can you match the `${SITE_VARS}` handling here with how it's done in `deploy/streisand-local.sh`? I don't think it will default the same way.

deploy/streisand-existing-cloud-server.sh
```diff
@@ -0,0 +1,102 @@
...
+  esac
+
+  shift
+done
+
+# --------------------------------------------------------------------------
+# Fail if required options are not set.
+# --------------------------------------------------------------------------
+
+if [ -z "${IP_ADDRESS}" ] || [ -z "${SITE_VARS}" ] || [ -z "${SSH_USER}" ]; then
```
cpu: Minor nit: I think if `${SSH_USER}` is empty/not provided we should just default to using `root`.

exratione:
@cpu Moved the `SITE_VARS` handling into a sourced script, and addressed the other two issues.

PR #1028, opened 2017-11-01
add chain.pem for OCSP

pguizeline:


cpu:
I left [a comment on #1017](https://github.com/StreisandEffect/streisand/pull/1017#pullrequestreview-74267127) - In general I think its best to have each PR be as self-contained as possible. Right now #1017 working correctly depends on this PR being merged. Reviewing this PR alone it seems like `le_chain` is being added for no reason since nothing refers to it yet (except #1017). I think its easier to reason about & review if everything required for #1017 happens in #1017. Thanks!

PR #1032, opened 2017-11-02
Windows PowerShell fragments to set up L2TP, NAT settings

### LINKED ISSUE #823: Documentation: L2TP is suggested for Windows, but it can’t deal with NAT
I think this is a *documentation* bug. 

I tried to run an L2TP client on Windows 10. I had the problems described in #291. Other people are going to run into this too, and not be able to figure out why their Windows box isn’t working. I think we should do either or both of:

1. Describe (and provide the .reg file, why not) needed to make Windows L2TP work behind NAT
2. Document that L2TP on Windows has this issue and suggest other methods.

nopdotcom:
Include a PowerShell command to fix L2TP NAT tunneling. Most people need this (or a .reg file). Fixes #823.

Instead of hand-hacking the phone book (don't ask), use the PowerShell cmdlets to set up L2TP connections.

Note that there is an interesting dance here: Windows 10 blocks script execution. But it doesn't block cut&paste into admin PowerShell windows...


playbooks/roles/l2tp-ipsec/templates/instructions.md.j2
```diff
@@ -16,6 +16,46 @@ If the connection appears to time out, first check to make sure the passwords we
 
 <a name="windows"></a>
 ### Windows ###
+
+In order to follow the quick directions, you'll need to use Windows PowerShell ISE. To start it:
+
```
alimakki: I think this can be worded better by changing it to something like:

"In order to proceed, you will need to start Windows PowerShell ISE in administrator mode"....

playbooks/roles/l2tp-ipsec/templates/instructions.md.j2
```diff
@@ -16,6 +16,46 @@ If the connection appears to time out, first check to make sure the passwords we
...
+1. Look in the *Start* menu, under *Windows PowerShell*.
+1. Right-click on *PowerShell ISE*, and choose "Run as administrator"
```
alimakki: 1. Open the *Start* menu, and navigate to (or search for) *Windows PowerShell*

playbooks/roles/l2tp-ipsec/templates/instructions.md.j2
```diff
@@ -16,6 +16,46 @@ If the connection appears to time out, first check to make sure the passwords we
...
+
+#### Windows and NAT ####
+
+If you are connecting from behind a NAT (such as a home router), you need to add a setting to tell Windows to expect the NAT.
```
alimakki: you need to -> you will need to

playbooks/roles/l2tp-ipsec/templates/instructions.md.j2
```diff
@@ -16,6 +16,46 @@ If the connection appears to time out, first check to make sure the passwords we
...
+
+Start PowerShell ISE. In the ISE command window, paste the following code, and press *Enter*.
+
+```
+Set-ItemProperty -Path HKLM:\SYSTEM\CurrentControlSet\Services\PolicyAgent -Name AssumeUDPEncapsulationContextOnSendRule -Value 2 -Type DWord
+```
+
+If you aren't using PowerShell, you could use `regedit` to add and set the value of `AssumeUDPEncapsulationContextOnSendRule`.
+
```
alimakki: I think the regedit instructions need to be more fleshed out (how do I navigate to the correct entry in regedit?). Alternatively, better to omit it entirely and could be made more thorough in an entry for the documentations folder, with pretty screenshots. 

Anyone else have thoughts on this?

cpu: Isn't there a way to load a .reg fragment of some sort into regedit so you don't have to do GUI clicking stuff?

(I might be crazy, my Windows experience is suuuper dated/hazy)

nopdotcom:
I'm regretting saying anything about `regedit` at all. You need an admin PowerShell to do the rest of my directions...

Yes, you can download `.reg` files. The primary audience for the file version is Google, I think. Not the worst thing to do.

The secondary audience is people following the "by hand" directions, although those have sorta rotted.

Speaking of rot, what's the oldest version of Windows we care about? I don't want to have anything to do with XP if I can help it. 7 is tolerable.

cpu:
> Speaking of rot, what's the oldest version of Windows we care about? I don't want to have anything to do with XP if I can help it. 7 is tolerable.

Good question. I'm not sure anyone has taken the time to decide that. I think XP is definitely out. Too old & beyond supported life

Personally I can't commit to doing much in the way of Windows support. @nopdotcom it seems like yourself and @alimakki have been doing the most w.r.t Windows. I vote that you decide what the appropriate support level is between the two of you. I think its reasonable to aim for modern Windows initially and then go from there. It's hard enough to support Windows 10, anything older will ratchet up the difficulty.

alimakki:
Just a quick search showed for Windows [market share](https://www.netmarketshare.com/operating-system-market-share.aspx?qprid=10&qpcustomd=0), shows that Windows 7 is still has the largest installation base.

I think that for the meantime we should consider Windows 7, and re-visit it periodically when the numbers change.

cpu:
@jlund How do you feel about Windows? Have any cycles for a review?

playbooks/roles/l2tp-ipsec/templates/instructions.md.j2
nopdotcom: I'm punting on the `.reg` for now in the interest of getting this merged before further iteration. I totally need to go over the UI instructions on a couple Windows variants, and I don't have the necessary excitement to do that today. (Besides, the automatic setup method works better than the UI instructions. :-)

nopdotcom:
I did a pass at Windows 7 and right now people have to use the manual directions. I'm going to work on that.

nopdotcom:
Doing a few last deploys.

PR #1037, opened 2017-11-04
Remove DigitalOcean nyc2 region due to limited availability.

CorbanR:


CorbanR:
Possibly a pointless commit.. but I feel like the ny2 region very commonly has capacity issues.

exratione:
Might be better just to add a comment line to mention as much. Otherwise six months from now, someone will see it is missing and add it back in.

cpu:
> Possibly a pointless commit.. but I feel like the ny2 region very commonly has capacity issues.

What happens when you choose NY2 and it is out of capacity? Do you have an example of what the error output looks like?

Right now I'm slightly leaning towards +1'ing @exratione's suggestion to document instead of remove but I don't fully understand what things look like in the failure case we're trying to avoid.

Thanks!

CorbanR:
The error is ```An exception occurred during task execution. To see the full traceback, use -vvv. The error was: DoError: Region is not available
fatal: [localhost]: FAILED! => {"changed": false, "failed": true, "msg": "Region is not available"}```

CorbanR:
I actually agree, removing probably isn't the best idea.

PR #1038, opened 2017-11-04
Refactor certificates role

alimakki:
This PR refactors the certificates role to do the following tasks: 
 - the underlying `client.yml` task is now responsible for generating client directories, reducing code duplication across the openvpn and ocserv roles
 - adds an additional task to generate client pkcs files, guarded by the `generate_pkcs` flag.
 - update and clean up the certificates documentation.

### DIFF
--- a/documentation/certificates.md
+++ b/documentation/certificates.md
@@ -22,16 +22,15 @@ certificate signed by the newly minted certificate authority.
 
 Client
 ------
-Guarded by `generate_client: yes`
+Generate client certificates using the CA from `generate_ca_server: yes`. 
+These client certificates can be used by VPN clients to authenticate with 
+VPN services on the server.
 
-Generate client side certificates using the same CA used for authenticating
-connecting VPN clients.
-
-Client PKCS
------------
+Client (PKCS#11 format)
+-----------------------
 Guarded by `generate_pkcs: yes`
 
-Generate client PKCS files.
+Also convert client certificates into PKCS#12 format.
 
 Infrastructure
 --------------
@@ -50,9 +49,6 @@ OpenConnect CA:
    \__ OpenConnect client 1
    \__ OpenConnect client ...
    \__ OpenConnect client n
-      \__ OpenConnect client PKCS 1
-      \__ OpenConnect client PKCS ...
-      \__ OpenConnect client PKCS n
 
 OpenVPN CA: 
    \__ OpenVPN server cert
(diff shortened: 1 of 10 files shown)


documentation/certificates.md
```diff
@@ -2,24 +2,44 @@ Streisand PKI
...
+Generate a certificate authority, and a server 
 certificate signed by the newly minted certificate authority.
 
-The latter task is responsible for generating client side certificates
-(using the same CA) used for authenticating connecting VPN clients.
+Client
+------
+Guarded by `generate_client: yes`
+
+Generate client side certificates using the same CA used for authenticating
```
cpu: I think this sentence is a little confusing because it's mildly self-referential. The CA authenticates VPN clients using the client side certificates that are generated by the CA to authenticate VPN clients :-)

How about:

"Generate client certificates using the CA from `generate_ca_server: yes`. These client certificates can be used by VPN clients to authenticate with VPN services on the server." ?

documentation/certificates.md
```diff
@@ -2,24 +2,44 @@ Streisand PKI
...
+connecting VPN clients.
 
-Each VPN service that leverages the Streisand PKI generates its
-own CA/Server certificate and client certificates.
+Client PKCS
+-----------
+Guarded by `generate_pkcs: yes`
+
+Generate client PKCS files.
```
cpu: Maybe a more descriptive comment here is: "Also convert client certificates into PKCS#12 format". They aren't distinct client certificates and "PKCS" without specifying "PKCS#12" is a little bit too vague.

documentation/certificates.md
```diff
@@ -2,24 +2,44 @@ Streisand PKI
...
+Client
+------
+Guarded by `generate_client: yes`
+
+Generate client side certificates using the same CA used for authenticating
+connecting VPN clients.
 
-Each VPN service that leverages the Streisand PKI generates its
-own CA/Server certificate and client certificates.
+Client PKCS
```
cpu: "Client (PKCS#11 format)" maybe?

documentation/certificates.md
```diff
@@ -30,6 +50,9 @@ OpenConnect CA:
    \__ OpenConnect client 1
    \__ OpenConnect client ...
    \__ OpenConnect client n
+      \__ OpenConnect client PKCS 1
```
cpu: I would omit the PKCS#11 stuff from this hierarchy - probably too complicated to fit in nicely. It's the same client certs in a diff format. The way you have it here also makes it seem like the PKCS#11 format clients certs are under the "Openconnect client n" hierarchy when instead they're adjacent leafs to each of the "client" entries.

alimakki:
@cpu all done!

alimakki:
@cpu thanks for the quick review :)

PR #1042, opened 2017-11-06
User OpenSSL for PKCS12 generation

alimakki:
Cleans up the certificates role by replacing the use of certtool to create ~~certificates~~ PKCS#12 files, opting instead for OpenSSL as used throughout the rest of the role.

This also brings us the additional benefit of passing the PKCS#12 passwords as an argument rather than relying on expect.

Tested on a Linode instance with no issues. 

Edit: clarification

### DIFF
--- a/playbooks/roles/certificates/tasks/pkcs.yml
+++ b/playbooks/roles/certificates/tasks/pkcs.yml
@@ -28,14 +28,16 @@
     label: "{{ client_name.item }}"
   changed_when: False
 
-- name: "Convert the keys and certificates into PKCS #12 format"
-  expect:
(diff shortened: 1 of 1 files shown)


playbooks/roles/certificates/tasks/pkcs.yml
```diff
@@ -28,14 +28,16 @@
...
-    creates: "{{ tls_client_path }}/{{ vpn_client_password.client_name.stdout }}.p12"
+- name: "Convert the {{ vpn_name }} client keys and certificates into PKCS #12 format"
+  command: >
+    openssl pkcs12 -export
+    -in {{ tls_client_path }}/{{ vpn_client_password.client_name.stdout }}/client.crt
+    -inkey {{ tls_client_path }}/{{ vpn_client_password.client_name.stdout }}/client.key
+    -name {{ vpn_client_password.client_name.stdout }}
+    -out {{ tls_client_path }}/{{ vpn_client_password.client_name.stdout }}.p12
+    -certfile {{ tls_client_path }}/ca.crt
+    -passout pass:"{{ vpn_client_password.stdout }}"
```
cpu: For folks who may notice the `--pkcs-cipher 3des-pkcs12` parameter from the original `command` not being reflected in the new `openssl pkcs12` command `man pkcs12` confirms that `openssl pkcs12` defaults to `-des3` for the private key encryption.

PR #1053, opened 2017-11-18
lets-encrypt: Update agreements PDF

wzyboy:
Let's Encrypt has updated their agreements.


cpu:
Thanks @wzyboy - ironically I've been filing issues across a bunch of projects to fix this over the past week and forgot Streisand! :blush:

PR #1054, opened 2017-11-18
Avoid hardcoding LE subscriber agreement URL.

cpu:
Prior to this commit the Let's Encrypt subscriber agreement URL required
for the [acmetool non-interactive response file](https://github.com/StreisandEffect/streisand/blob/bbd3f34094955f39828d0d104dd75e63f0e10497/playbooks/roles/lets-encrypt/templates/response.yaml.j2) was [hardcoded](https://github.com/StreisandEffect/streisand/blob/bbd3f34094955f39828d0d104dd75e63f0e10497/playbooks/roles/lets-encrypt/vars/main.yml#L4). This means
whenever Let's Encrypt changes their subscriber agreement URL the
lets-encrypt role will break until the agreement var is [updated to the
new value](https://github.com/StreisandEffect/streisand/commit/bbd3f34094955f39828d0d104dd75e63f0e10497).

Fortunately the ACME protocol specifies a way to learn the server's
current terms-of-service agreement URL. [Section 7.1.1 of draft-08](https://tools.ietf.org/html/draft-ietf-acme-acme-08#section-7.1.1) says:

    The object MAY additionally contain a field "meta".  If present, it
    MUST be a JSON object; each field in the object is an item of
    metadata relating to the service provided by the ACME server.

    The following metadata items are defined, all of which are OPTIONAL:

      "terms-of-service" (optional, string):  A URL identifying the current
        terms of service.

Let's Encrypt does send a "meta" element with a "terms-of-service" item.
This commit adds a small Python script that fetches the current
terms-of-service URL from the API directory for use in the acmetool
response file.

This will keep the response file current even if the agreement URL
changes.

### DIFF
--- a/playbooks/roles/lets-encrypt/tasks/main.yml
+++ b/playbooks/roles/lets-encrypt/tasks/main.yml
@@ -6,6 +6,10 @@
     state: directory
     recurse: yes
 
+- name: Find the current Let's Encrypt subscriber agreement URL
+  script: "../scripts/termsOfServiceLookup.py {{ le_api_endpoint }}"
+  register: le_agreement
+
 - name: Put reponse file in place
   template:
     src: response.yaml.j2
--- a/playbooks/roles/lets-encrypt/tasks/main.yml
+++ b/playbooks/roles/lets-encrypt/tasks/main.yml
@@ -9,6 +9,7 @@
 - name: Find the current Let's Encrypt subscriber agreement URL
   script: "../scripts/termsOfServiceLookup.py {{ le_api_endpoint }}"
   register: le_agreement
+  changed_when: False
 
 - name: Put reponse file in place
   template:
--- a/playbooks/roles/lets-encrypt/tasks/main.yml
+++ b/playbooks/roles/lets-encrypt/tasks/main.yml
@@ -7,7 +7,7 @@
     recurse: yes
 
 - name: Find the current Let's Encrypt subscriber agreement URL
-  script: "../scripts/termsOfServiceLookup.py {{ le_api_endpoint }}"
+  script: "../scripts/terms_of_service_lookup.py {{ le_api_endpoint }}"
   register: le_agreement
   changed_when: False
 
--- a/playbooks/roles/lets-encrypt/scripts/termsOfServiceLookup.py
+++ b/playbooks/roles/lets-encrypt/scripts/termsOfServiceLookup.py
@@ -1,0 +1,71 @@
+#!/usr/bin/env python
+"""
+termsOfServiceLookup.py fetches the terms-of-service URL from an ACME
+/directory URL. This can be used to avoid hardcoding the current
+terms-of-service URL somewhere and having to update it every time
+Let's Encrypt migrates to an updated ToS.
+
+@cpu - daniel@binaryparadox.net
+"""
+from __future__ import print_function
+
+import sys
+import urllib2
+import json
+
+
+def errorPrint(*args, **kwargs):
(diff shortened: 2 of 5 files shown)


cpu:
@wzyboy What do you think about something like this to keep the agreement current?

playbooks/roles/lets-encrypt/tasks/main.yml
```diff
@@ -6,6 +6,10 @@
     state: directory
     recurse: yes
 
+- name: Find the current Let's Encrypt subscriber agreement URL
+  script: "../scripts/termsOfServiceLookup.py {{ le_api_endpoint }}"
+  register: le_agreement
+
```
wzyboy: A "changed_when: False" could be used to avoid "changed" status every time the task is run.

wzyboy:
@cpu Wow it's pretty neat we can fetch the URL in advance! This should solve hard-code problem.

Just a little thing: the script might better use snake case rather than camel case to be in consistent style with other Python scripts in the project, and pythonic generally.

wzyboy:
@cpu BTW did you look into "uri" Ansible module already? It might be able to fetch the URL with the help of some Jinja filters ("from_json" or something similar) so that there's no need to ship a script.

My apartment is having an Internet breakdown so I cannot test my ideas for now.

cpu:
> Just a little thing: the script might better use snake case rather than camel case to be in consistent style with other Python scripts in the project, and pythonic generally.

Thanks! I don't write much Python these days and fell into CamelCaseHabits :-) I think I've addressed all of your feedback. I appreciate you taking a look.

> BTW did you look into "uri" Ansible module already? It might be able to fetch the URL with the help of some Jinja filters ("from_json" or something similar) so that there's no need to ship a script.

That's a neat idea. I thought about using the `url` module but I thought I would have to process the JSON with a bash one-liner or something and it felt easier to catch errors and handle things robustly with a short Python script.

Using a `from_json` filter is a great suggestion. I will try to find some time to experiment with that.

> My apartment is having an Internet breakdown so I cannot test my ideas for now.

Bad news! I hope your Internet starts working again soon! :computer: :satellite: :chart_with_upwards_trend:

wzyboy:
@cpu Good news! It turns out that "uri" module has built-in JSON support. I've written a demo playbook:

```
- hosts: localhost
  connection: local
  gather_facts: no

  vars:
    le_api_endpoint: "https://acme-v01.api.letsencrypt.org/directory"

  tasks:
    - name: Fetch HTTP body
      uri:
        url: "{{ le_api_endpoint }}"
        body_format: json
      register: http_body

    - name: Extract TOS URL
      set_fact:
        le_agreements: "{{ http_body['json']['meta']['terms-of-service'] }}"

    - name: Print the URL
      debug:
        var: le_agreements
```

The result:
```
TASK [Print the URL] ******************
ok: [localhost] => {
    "le_agreements": "https://letsencrypt.org/documents/LE-SA-v1.2-November-15-2017.pdf"
}
```

cpu:
@wzyboy Wow that's perfect! Thanks for the demo - much cleaner than my script :-D

Do you want me to close this PR so you can submit your own? Alternatively I could update my own PR and credit you for the solution :-) Up to you!

wzyboy:
@cpu I have opened a new PR for the patch: https://github.com/StreisandEffect/streisand/pull/1055 Thanks for the ACME spec :-)

cpu:
@wzyboy Perfect! You're the best.

Closing this in favour of #1055.

PR #1055, opened 2017-11-18
lets-encrypt: Query Let's Encrypt for current agreement URL

wzyboy:
As @cpu [pointed out](https://github.com/StreisandEffect/streisand/pull/1054#issue-275086994), there is an ACME API for querying the latest TOS. So we can actually fetch the TOS URL when running the playbook, instead of hardcoding it.

### DIFF
--- a/playbooks/roles/lets-encrypt/tasks/main.yml
+++ b/playbooks/roles/lets-encrypt/tasks/main.yml
@@ -6,13 +6,13 @@
     state: directory
     recurse: yes
 
-- name: Query Let's Encrypt to find out the latest agreements
+- name: Query Let's Encrypt to find the current subscriber agreement URL
   uri:
     url: "{{ le_api_endpoint }}"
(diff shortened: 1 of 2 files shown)


playbooks/roles/lets-encrypt/tasks/main.yml
```diff
@@ -6,6 +6,16 @@
     state: directory
     recurse: yes
 
+- name: Query Let's Encrypt to find out the latest agreements
```
cpu: Minor nit: I think this would be clearer as:
"Query Let's Encrypt to find the current subscriber agreement URL"

wzyboy:
@cpu The "uri" and "set_fact" module do not require "changed_when: False" because they are always "ok" if no changes are made.

cpu:
@wzyboy Perfect! I'll merge this once CI is satisfied. Thanks again.

PR #1067, opened 2017-11-22
Generate multiple WireGuard client profiles.

cpu:
This commit updates the WireGuard playbook to follow the client
generation approach used by the OpenVPN playbook. This means using the
BIP-0039 randomly generated client names & honouring the `vpn_clients`
var to determine how many client profiles to make. Like OpenVPN only one
client profile may be in use at a time.

I added un peu de mauvais français to keep the translations ~= in-sync.
Forgive my Franglais.

Thanks to @alimakki for helping me figure out some of the templating :-)

Resolves https://github.com/StreisandEffect/streisand/issues/693

### DIFF
--- a/playbooks/roles/wireguard/tasks/docs.yml
+++ b/playbooks/roles/wireguard/tasks/docs.yml
@@ -7,10 +7,14 @@
     mode: 0750
     state: directory
 
-- name: Copy the client configuration file to the WireGuard Gateway directory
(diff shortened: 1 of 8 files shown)


cpu:
Thanks for the review @alimakki :taco: 

I also ran a test provision on a vultr node with 3 WireGuard clients. No trouble having two Linux desktops and an Android phone (with a patched kernel) connected at once using the three distinct client profiles following the setup instructions.

PR #1071, opened 2017-11-26
Don't turn on HTTP Strict Transport Security on non-LE servers

nopdotcom:
Like it or not, a lot of people depend on being able to click through the "this cert sucks" page on non-LE servers. HSTS prohibits click-throughs. Perhaps "lots of people depend" is a little strong, but there are going to be people who get stuck on this step.

### DIFF
--- a/playbooks/roles/streisand-gateway/templates/vhost.j2
+++ b/playbooks/roles/streisand-gateway/templates/vhost.j2
@@ -32,7 +32,6 @@ server {
   ssl_session_tickets off;
 
   # headers
-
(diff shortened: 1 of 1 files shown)


playbooks/roles/streisand-gateway/templates/vhost.j2
```diff
@@ -29,7 +32,7 @@ server {
   ssl_session_tickets off;
 
   # headers
-  add_header Strict-Transport-Security max-age=63072000;
+
```
cpu: Can you remove this unnecessary whitespace?

PR #1075, opened 2017-11-29
Added autoindexing for the openvpn directories

Spengreb:
Commit 811bea1 broke listing .crt and .key files for openvpn directories as it now appears with a 403 error. 

I added a bit to allow auto indexing on openvpn.

### DIFF
--- a/playbooks/roles/streisand-gateway/templates/vhost.j2
+++ b/playbooks/roles/streisand-gateway/templates/vhost.j2
@@ -48,6 +48,10 @@ server {
   location / {
     autoindex off; 
   }
+  
+  location /openvpn/ {
+    autoindex on;
+  }
 
   location /monit/ {
     rewrite ^/monit/(.*) /$1 break;


cpu:
@Spengreb Is it neccessary to access `/openvpn/` by directory listing? I believe all of the `.crt` and `.key` files are linked from the openvpn instructions. Maybe I'm misunderstanding.

Spengreb:
@cpu You are mistaken at least in the case of OpenVPN -> Linux (Ubuntu) -> Step 1. then has links to a bunch of directory listings which will fail on 403 presently. 
selection_037

cpu:
>  You are mistaken at least in the case of OpenVPN -> Linux (Ubuntu) -> Step 1. then has links to a bunch of directory listings which will fail on 403 presently.

@Spengreb You're 100% correct. I was able to reproduce this. Thanks for opening the bug report/PR.

In this case I think the preferred fix is to adjust the OpenVPN Linux/NetworkManager instructions template to link directly to the files required instead of relying on the bare indexes. I opened https://github.com/StreisandEffect/streisand/pull/1076 to do that. I'm going to close your PR in favour of #1076. Thanks again!

PR #1083, opened 2017-12-05
Update L2TP instructions in French.

alimakki:


cpu:
Merci @alimakki :champagne:

PR #1097, opened 2017-12-13
Split up the README into multiple pieces, to ease translation and edits

nopdotcom:
I'm planning to break up the READMEs into multiple pages. As an interim step, split the READMEs into multiple *files*, and use a script to reassemble the READMEs from them.

Splitting up the README also makes other translation tasks easier. For example, it caught that the `ru` readme is missing the *Noninteractive Deployment* section.


cpu:
> I'm planning to break up the READMEs into multiple pages. As an interim step, split the READMEs into multiple files, and use a script to reassemble the READMEs from them.

How long do you think this interim state will last? Would it be better to hold off and tackle the README breakup in one step? I'm not sure what the advantage of the interim is (except, as you mentioned, making it easier to notice places where the fragments have drifted)

nopdotcom:
I was holding off on polishing/documenting the generation tools until there was general consensus this approach made sense. :-)

The problem I'm trying to solve (I guess I should have said this up front);

* README is too big. It's intimidating.
* It seems "obvious" to me that the right thing to do is to turn /README.md into an introduction/blurb plus links to subdocuments. (The hypothetical intro doesn't exist yet.) 
* The sections we do have now are not free-standing. Until that editing gets done, we can't go to subdocuments. If we have them split up, we can make them free-standing one-by-one, and not break the README in the process; we can make them subdocuments too.
* I should use the term "refactoring" somewhere in here.
* I'd like to avoid a mondo PR branch that changes both content and structure; it will drift.

Maybe the right thing for me to do is put up a sketch of what I think the documentation will eventually look like.

cpu:
@nopdotcom Thanks for the added background/context. This PR makes _a lot_ more sense to me now. I agree that it's a good interim step, especially in light of the editing that will need to be done to make the fragments more free-standing.

cpu:
> Maybe the right thing for me to do is put up a sketch of what I think the documentation will eventually look like.

This seems like a good case for an omnibus issue or a github project that can have subissues/associated PRs (this one included)

nopdotcom:
I'm going to close this PR, since the docs are gradually drifting, and it will be easier to redo than pretend to do mixmaster merges.

While I'm thinking about it: this may not be the *greatest* idea, but if we aren't going to physically split up the main document, we could generate subpages by splitting up the existing `README.md` along "cut here" lines.

PR #1105, opened 2017-12-15
Revert "Switch to BIP identifiers/password generation"

### LINKED ISSUE #1103: commit 36a46f7 (BIP wordlists) violates TOR config policy and breaks TOR -> breaking Streisand default deployment
### ExecSummary:
The adoption of BIP wordlists includes dashes "-" between words, which violates TOR configuration file standards. This prevents TOR from starting, and consequently Streisand from completing the default deployment. 

### Expected behavior:
TOR runs, Streisand deployment succeeds.

### Actual Behavior:
TOR fails to validate the configuration file /etc/tor/torrc and terminates. Consequently, this prevents the deployment to succeed with following error:
```
TASK [tor-bridge : Wait until obfs4proxy information has shown up in its state file] *************************
fatal: [localhost]: FAILED! => {"changed": false, "elapsed": 300, "msg": "Timeout when waiting for search string node-id in /var/lib/tor/pt_state/obfs4_state.json"}
```

### Steps to Reproduce:
./streisand
select default setup

### Additional Details:
TOR complains that the auto generated names for the 'Nickname' field in /etc/tor/torrc contain a dash "-", which violates TOR configuration file standards, as only the characters [a-zA-Z0-9] are allowed for this field. 

This regression was introduced by commit: https://github.com/StreisandEffect/streisand/commit/36a46f765b82ab10a4107b7a00193bc4cee8462f

TOR stout error is:
```
[warn] Failed to parse/validate config: Nickname 'monkey-glare', nicknames must be between 1 and 19 characters inclusive, and must contain only the characters [a-zA-Z0-9].
[err] Reading config failed--see warnings above.
```

### Workaround:
While deploying Streisand, the following message is displayed for 300 seconds before timing out:
`TASK [tor-bridge : Wait until obfs4proxy information has shown up in its state file]`
During this timeframe, open another terminal window and edit TOR config file
`nano /etc/tor/torrc `
verify that the line starting with Nickname does not include dashes "-", if so delete the dash and save the file. Start TOR: 
`systemctl start tor`

cpu:
This reverts commit 3d153287c889cc95f268e82d5807ea82b573f6a9, reversing changes made to 9baef3eceeaf94cd823eb376eee00096de2f7b15.

This commit was causing a [tor regression](https://github.com/StreisandEffect/streisand/issues/1103). 

Resolves #1103 

_Note: Happy to close this if we have a fix for the root cause of 1103 to prefer_

### DIFF
--- a/playbooks/roles/certificates/tasks/ca-server.yml
+++ b/playbooks/roles/certificates/tasks/ca-server.yml
@@ -23,7 +23,7 @@
     creates: "{{ tls_ca }}.crt"
 
 - name: Generate a random server common name
-  shell: "{{ streisand_word_gen.long_identifier }} > {{ tls_server_common_name_file }}"
+  shell: grep -v -P "[\x80-\xFF]" /usr/share/dict/american-english-huge | sed -e "s/'//" | shuf -n 2 | xargs | sed -e 's/ /./g' | cut -c 1-64 > {{ tls_server_common_name_file }}
   args:
     creates: "{{ tls_server_common_name_file }}"
 
--- a/playbooks/roles/certificates/tasks/pkcs.yml
+++ b/playbooks/roles/certificates/tasks/pkcs.yml
@@ -1,6 +1,6 @@
 ---
 - name: "Generate a random password that will be used during the PKCS #12 conversion"
-  shell: "{{ streisand_word_gen.weak_password }} > {{ tls_client_path }}/{{ vpn_name }}-{{ client_name.stdout }}-pkcs12-password"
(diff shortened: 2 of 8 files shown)


nopdotcom:
@cpu This was fixed in 2c69f04717600d2baa7ad2a42ef37dc5ed0cb76f, right before this revert. Got a preferred way to unrevert this?

cpu:
@nopdotcom A new PR would be great - doesn't need to be a revert of the revert - a new commit would be fine IMO.

PR #1118, opened 2017-12-25
Update ./streisand documentation

thorncorona:
I think it's a good idea to mention to the user that the IP is separate from the domain name they use to setup, and that they can set it up later. I spent a couple hours trying to use a FQDN for my IP only to realize that's a setting that comes up later...

### DIFF
--- a/streisand
+++ b/streisand
@@ -128,7 +128,7 @@ Please enter the word 'streisand' to continue: "
 }
 
 function existing_server() {
-  read -r -p "What is the IP of the existing server: " SERVER_IP
+  read -r -p "What is the IP of the existing server (you may choose to set a domain name later): " SERVER_IP
 
   be_careful_friend "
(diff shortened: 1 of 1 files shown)


cpu:
Hi @thorncorona,

Thanks for submitting this PR! I appreciate you making an effort to save others from the same frustration you experienced.

Are you open to another fix? I think the PR you submitted may confuse slightly. The only thing that ends up using the domain name accepted later in provisioning is the HTTPS NGINX gateway for documentation/client configs and mirrored software. None of the client profiles or other serverside daemons will reference the domain name or otherwise use it. I worry that accepting this PR would give the wrong impression and lead to issues if a user changes the address the domain name points to.

I think a better approach might be to try and validate that the SERVER_IP is an IPv4 address and fail with a descriptive error message if it isn't.

@thorncorona Would you be interested in closing this PR and submitting another for the above approach?

thorncorona:
@cpu Yeah sure.

PR #1121, opened 2017-12-27
Added the 'Mode' setting under 'Additional TLS Auth' settings for OpenVpn for Ubuntu 17.10

Frichetten:
I just noticed on Ubuntu 17.10, when setting up OpenVPN, it was missing a 'Mode' setting under 'Additional TLS authentication or encryption'. As a result I pulled the file, added the line instructing the user to select TLS-Auth for their mode (that is the only viable option) and have submitted a pull request.

I checked to make sure there wasn't already an existing issue or pull request, so I apologize if I missed it. Let me know if you need screenshots or anything else for confirmation.

**As a side note, I'm noticing that Travis CI is failing this, and other pull requests because of a missing file at https://images.linuxcontainers.org/1.0/images/ubuntu%2Fxenial%2Famd64. If someone could fix that, that would be great :)**


cpu:
Thanks @Frichetten !

Frichetten:
@cpu Not a problem! I've been a fan of the project for a long time. Glad I could help in some small way. :)

PR #1124, opened 2017-12-29
fixing inconsistency in apache-libcloud requirement for GCE

adityarpillai:
The French README already had this change, but instead of installing "apache-libcloud>=1.5.0", the user should install a version greater than 1.17.0. Although using the >=1.5.0 function will install apache-libcloud 2.2.1, it seemed better for both the sake of consistency and to avoid any bugs to ensure that the installed version is above the specified minimum in Ansible for use with Google's JSON credentials, which Streisand uses.

Look here for Ansible's documentation:
http://docs.ansible.com/ansible/latest/gce_module.html

### DIFF
--- a/README-chs.md
+++ b/README-chs.md
@@ -140,7 +140,7 @@ Streisand 运行在**你自己的计算机上时（或者你电脑的虚拟机�
         sudo pip install dopy==0.3.5
   * Google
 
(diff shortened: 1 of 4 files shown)


cpu:
Thanks @adityarpillai!

PR #1134, opened 2018-01-04
Update ss-libev from tag: 3.1.1 to 3.1.2

denoza:
Upstream developers have updated the `shadowsocks-libev` version from `3.1.1` to `3.1.2`. [Changelog](https://github.com/shadowsocks/shadowsocks-libev/blob/master/debian/changelog) can be accessed here.


alimakki:
Thank you @denoza

PR #1140, opened 2018-01-07
 Update venv builder for upstream module changes, more compatibility

nopdotcom:
Update venv builder for upstream module changes, more compatibility  …
Apparently something bad happened with pycurl, so we need more
dependencies than just libcurl4-dev to build stuff with it.


nopdotcom:
This is here because I've run out of energy to test it; help appreciated. Works on Vultr and Google Cloud Shell, which are usually the weird ones.

util/venv-dependencies.sh
```diff
@@ -36,6 +36,7 @@ requires sudo/root access.
 'new-directory' must be somewhere you can write to. A good place may be
 $HOME/streisand-deps. If it already exists,
 please delete the directory, or use a different name.
+
```
cpu: Intentional whitespace diff?

util/venv-dependencies.sh
```diff
@@ -199,7 +211,8 @@ fi
 # In case we have a new virtualenv executable.
 hash -r
 
-if ! virtualenv --python=python2 $virtualenv_no_site "$1"; then
+if ! virtualenv --python=python2 $NO_SITE_PACKAGES "$1"; then
+    dn="$(dirname "$1")"
```
cpu: What is the `dn="$(dirname "$1")"` change here doing? I'm likely missing something but maybe that's a sign that there needs to be a comment here.

util/venv-dependencies.sh
```diff
@@ -59,12 +60,20 @@ if [ "$#" -ne 1 ]; then
...
+
+On your system, 'pip' appears to invoke Python 3's pip. This might cause problems.
+This script will try switching to 'pip2', but your pip2 install might be broken too.
+
+"
+    PIP="pip2"
+fi
 
-if ! pip2 --version >/dev/null 2>&1; then
+if ! "$PIP" >/dev/null 2>&1; then
```
cpu: Will running "$PIP" without args do the expected thing with this conditional? I thought it might return non-zero without args. Should it be "$PIP" --version?

util/venv-dependencies.sh
nopdotcom: Yeah. The style for all the other diagnostics has a trailing blank line.

nopdotcom:
So, this is broken because setting `--no-site-packages` is morally the right thing, but python-apt is only available outside the virtualenv. Grr.

cpu:
> So, this is broken because setting --no-site-packages is morally the right thing, but python-apt is only available outside the virtualenv. Grr.

Is there a path forward? I'm not sure what to recommend. Do you think the virtualenv approach is salvageable or should we just ship the util script that installs the deps to the host directly?

nopdotcom:
Sorry, I forgot to hit send!

The good news is that this just affects localhost installs. On the remote systems, Ansible invokes `python`, and gets `/usr/bin/python`. Running inside a virtualenv, it picks up our isolated environment, which can't have `python-apt` in it. 

The remedy is to tell Ansible, "here's the explicit path to python on the machine being provisioned".

nopdotcom:
Oh yeah, @cpu asked to be @-mentioned when this was ready again, I think.

PR #1148, opened 2018-01-20
Adding docker support

hammady:
Simple Dockerfile so that the local environment is not polluted.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,8 @@
+FROM python:2.7
+LABEL maintainer="Hossam Hammady <github@hammady.net>"
+WORKDIR /root
+COPY ./requirements.txt /root/requirements.txt
+RUN pip install -r requirements.txt
+COPY / /root/
+CMD ./streisand
+
--- a/README.md
+++ b/README.md
@@ -213,6 +213,20 @@ To run the Streisand provisioning against an existing server:
         --ssh-user root \
         --site-config global_vars/noninteractive/digitalocean-site.yml
 
+Docker Installation
+-------------------
+
+If you don't want to pollute your local environment, you can use docker.
+Just build the image, then run it with your ssh key and output folder
+mounted:
+
+    # build
+    docker build . -t streisand
+
+    # run
+    docker run -it -v ~/.ssh:/root/.ssh -v $PWD/generated-docs:/root/generated-docs streisand:latest
+
+
 Upcoming Features
 -----------------
 * Easier setup.
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,6 +4,9 @@
 #
 # If you have problems running this, try the util/venv-builder.sh script.
 
+ansible
+markupsafe
+
 # Core with Azure dependencies
 ansible[azure]
 


cpu:
Hi @hammady,

Thanks for the PR. Have you seen https://github.com/StreisandEffect/streisand/pull/980/ ? @nopdotcom had started a proof of concept Dockerfile that I think was a little bit more fleshed out than this one.

@hammady @nopdotcom Do the two of you want to coordinate on a combined effort to add Docker support? I think some of my feedback on 980 will apply here (I'd prefer not to mount the whole ssh directory for one).

Thanks!

hammady:
@cpu thanks for your reply. I looked quickly at the other PR and will look more. From the first glance, I can see this is much simpler. But you are right on the ssh thing.

nopdotcom:
I think it would be cool to wake up the Docker effort. The place I got stuck last time was trying to support choosing SSH keys.

I'll clean up that branch a little, so it can at least work with the "default" `~/.ssh/id_rsa`.

Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM python:2.7
```
nopdotcom: I'm not sure we want to do this. We do most of our testing inside Ubuntu. Part of the point of Dockerization is to make debugging for us easier, and adding another OS/platform as an execution engine has testing costs.

Also, the python image, while perhaps saving some time on building prerequisites, doesn't save any space.

Image | version  | size
---- | ----  | ----
nop-image|latest|726MB
python-image|latest|946MB

Dockerfile
```diff
@@ -0,0 +1,8 @@
...
+LABEL maintainer="Hossam Hammady <github@hammady.net>"
+WORKDIR /root
+COPY ./requirements.txt /root/requirements.txt
+RUN pip install -r requirements.txt
```
nopdotcom: I tend to upgrade pip in the Ubuntu image before installing anything with it. I'm not sure if there is a huge difference, but there is a lot of complaining if you aren't using an up-to-date version of pip.

README.md
```diff
@@ -213,6 +213,20 @@ To run the Streisand provisioning against an existing server:
...
+
+If you don't want to pollute your local environment, you can use docker.
+Just build the image, then run it with your ssh key and output folder
+mounted:
+
+    # build
+    docker build . -t streisand
+
+    # run
+    docker run -it -v ~/.ssh:/root/.ssh -v $PWD/generated-docs:/root/generated-docs streisand:latest
```
nopdotcom: This is much cleaner than what I did, although you'll get the same pushback on `~/.ssh` mounting (don't expose all the user's keys). Until the two of us come up with something clever or cleaner to do about keys, we should at least mount them read-only. 

Come to think of it, we can make the user's choice of SSH keys *appear* at `/root/.ssh/id_rsa`, through (read-only) bind mounting. This might resolve @cpu's issues with mounting the whole `~/.ssh`.

I think we need to at least do the run phase in a script, in order to validate a few things. For starters, we do need to pick an SSH keypair that exists, or else Docker might do something really *dumb* with that bind. (The directory `/path/to/conf` exists on my machine, because Docker was happy to create the directory mentioned in the cut&pasted line. :‐(

requirements.txt
```diff
@@ -4,6 +4,9 @@
 #
 # If you have problems running this, try the util/venv-builder.sh script.
 
+ansible
+markupsafe
+
 # Core with Azure dependencies
 ansible[azure]
```
nopdotcom: We had problems installing both `ansible` and `ansible[azure]`. Although it looks correct, we need to take that out. :-(

nopdotcom: ...and maybe we should add that as a comment, since it does look odd not to specify things the normal way.

alimakki: Removing `ansible` and leaving `ansible[azure]` should work. I think that changing `ansible[azure]` to just `azure` should do the trick as well, though I haven't tested it.

asomov:
Hi all,
the local environment does not work for me (Pythons are fighting...)

The Dockerfile works as is. 
Thank you !

nopdotcom:
@asomov If it works for you, I am happy.

I need to document the `utils/venv-dependencies.sh` script. It should be able to handle the warring-Pythons problem too. But again, you have a working system :-)

nopdotcom:
When we're going to do this, we probably need a separate branch or repository. I'd volunteer mine, but I'm trying to clean up a Git disaster and will be deleting and re-forking it soon. We could use @hammady's repository, but there's a time commitment to host it (mostly dealing with merges). I know I wouldn't have much energy for it if I was just trying to get things running.

I'm going to take this out of the PR list now, because I (and @hammady?) should probably start fresh, and use #980 as cut&paste fodder

@alimakki, @cpu, @jlund: what should we be doing with StreisandEffect/streisand branches?

Dockerfile
hammady: Python image is based on Debian, so I do not see any problem here. One should not reinvent the wheel, and if you need Ubuntu/Debian + Python then you should directly use the Python image.

Dockerfile
hammady: Updating the pip won't harm.

README.md
hammady: I suggest that we add a step in the `Dockerfile` to generate a brand new key pair, then place it in the `generated_docs` directory. The user is then free to copy it back to his `~/.ssh` or use it from there using the `-i` parameter.

requirements.txt
hammady: What kind of problems did you face? This worked for me flawlessly. And because it is Docker, then it should always work the same way.

hammady:
@nopdotcom I am waiting for some replies on this thread, and then I can do all the fixes. But I don't agree with starting fresh as the change was minimal anyway. I guess the main feature I will be adding is the brand new ssh key generation.

Dockerfile
nopdotcom: Yeah, I'm not worried about the size at all, except to say that there is more stuff included with library/python. What I'm worried about is reproducibility of results.

I'm willing to try it both ways (ubuntu and python), but using Debian may stress out install scripts a little. I'm trying to keep Debian operational, if for no other reason than the free Google Cloud Shell has a Debian environment.

requirements.txt
alimakki: @hammady IIRC, I faced problems trying to provision to Azure when installing `ansible` followed by `ansible[azure]`

README.md
nopdotcom: What I *really* want to do is use the ssh-agent to hold the private keys, and extract the public key from the agent for an `id_rsa.pub`. My primary SSH key is held on a Yubikey, and it simply does not *have* an `id_rsa`.

Generating the `id_rsa` key is great for people who don't do this very much. But I have about a zillion machines I need to log into, and every additional `id_rsa` means more complexity for me--I'm already at the five cert limit, so I need to add `IdentitiesOnly` clauses to `~/.ssh/config` to pick which key goes where.

Another user's requirement was solved in Streisand in a way that made using `ssh-agent` harder, so this has been on my back burner.

I think the best thing is to read-only bind-mount the keys specified on the host to `/root/.ssh/id_rsa*` `id_rsa` and `id_rsa.pub` in. This exposes the private `id_rsa` to the container, but without the agent, that's inevitable. :‐(

requirements.txt
nopdotcom: It's possible it's fixed now. I'll take a look.

nopdotcom:
@hammady Thanks for commenting on this. I'm not a super Docker person, so I'm learning this as I go along.

I'll do a cleanup pass on my `Dockerfile` in this PR in any case, and I will test the `ansible` vs `ansible[azure]` stuff again. It's possible they've fixed this upstream.

PR #1149, opened 2018-01-20
Updated Wording for OpenVPN - Ubuntu for Ubuntu 17.10

Frichetten:
This may be a little nit-picky but I noticed that some of the wording is off for installing OpenVPN on Ubuntu. I think some changes were made in 17.10 and now things are laid out a little differently. 

I made 2 changes to the instructions. First, Instead of telling the user to check a box (it not longer exists), I mentioned that they should view a particular section (so they can update that information).

Second, there is no 'Save' button anymore, it is now referred to as 'Add'. I've linked some screenshots below.

[No Check Box](http://frichetten.com/images/z-9.png)
[No Save Button](http://frichetten.com/images/z-10.png)

PR #1152, opened 2018-01-22
Updated Wording for OpenVPN - Ubuntu for Ubuntu 17.10 (Fixed)

Frichetten:
This may be a little nit-picky but I noticed that some of the wording is off for installing OpenVPN on Ubuntu. I think some changes were made in 17.10 and now things are laid out a little differently.

I made 2 changes to the instructions. First, Instead of telling the user to check a box (it no longer exists), I mentioned that they should view a particular section (so they can update that information).

Second, there is no 'Save' button anymore, it is now referred to as 'Add'. I've linked some screenshots below.

[No Check Box](http://frichetten.com/images/z-9.png)
[No Save Button](http://frichetten.com/images/z-10.png)

If this looks like a repeat, it is. This was originally Pull Request #1149 . I wanted to make a quick fix so I did it on my Master branch. After that, I found something different to work on. To save the headache, I just closed it and did it the proper way :P

### DIFF
--- a/playbooks/roles/openvpn/templates/instructions.md.j2
+++ b/playbooks/roles/openvpn/templates/instructions.md.j2
@@ -124,12 +124,12 @@ It's preferable to configure Ubuntu using the OpenVPN plugin for NetworkManager.
 1. Go to the *TLS Authentication* tab.
    * Under *Server Certificate Check* choose `verify name exactly` and enter `{{ openvpn_server_common_name.stdout }}` as its value.
    * Check *Verify peer (server) certificate usage signature*.
-   * Check *Use additional TLS authentication*.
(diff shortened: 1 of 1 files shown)


alimakki:
@Frichetten you're probably right, if I remember correctly I made some changes based off of Ubuntu 16.04.

cpu:
Thanks again @Frichetten :trophy:

Frichetten:
@alimakki Does the documentation target a specific version of Mac OS or Linux? I'm sure there a little variances here and there that may make updatings the documentation tedious, but then again it's probably not radically different :P

alimakki:
@Frichetten not that I'm aware of, but I take into account the Serrias for mac, though I'd defer to the better judgment of others when it comes to Linux :)

Issue #1138, opened 2018-01-07
Clarify SSH key instructions for beginners unfamiliar with SSH/publickey auth


baimafeima:
I found these instructions not very clear and omitting several steps, they are certainly not written for total beginners: 

Make sure an SSH key is present in ~/.ssh/id_rsa.pub.
If you do not have an SSH key, you can generate one by using this command and following the defaults: ssh-keygen

1. No information is given how to make sure whether an SSH key is present. 
2. Why must it be in exactly this folder and not somewhere else? 
3. What is id_rsa.pub? Does the file have to be named exactly like this in order for it to work or can I just name it sshkeys.pub?
4. When generating an SSH key with the command ssh-keygen, no further information is given what name the file ideally should have and where it should be stored and how to do that.

cpu:
Thanks for the feedback!

nopdotcom:
I'm going to close this; the original itch has been scratched. My...really @cpu's...requests for improvement are filed separately in #11668.

PR #1153, opened 2018-01-22
Updated README to Clarify Information Regarding SSH Keys

Frichetten:
Added clarifying information in the README in regards to SSH keys. This was requested under Issue #1138. 

I'm happy to discuss ways for this to be improved! My primary goal was to explain a basic synopsis of what SSH keys are for, how to check if you have one, and mention that you can use a custom one when customizing your streisand instance.

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,8 +97,8 @@ Complete all of these tasks on your local home machine.
 * Streisand requires a BSD, Linux, or macOS system. As of now, Windows is not supported. All of the following commands should be run inside a Terminal session.
 * Python 2.7 is required. This comes standard on macOS, and is the default on almost all Linux and BSD distributions as well. If your distribution packages Python 3 instead, you will need to install version 2.7 in order for Streisand to work properly.
 * Make sure an SSH key is present in ~/.ssh/id\_rsa.pub.
-  * SSH keys use public-key cryptography to prove your identity to a server or service.
-  * To check if you already have an SSH key, please enter the following command at a Bash prompt.
+  * SSH keys are a more secure alternative to passwords that allow you to prove your identity to a server or service built on public key cryptography.
+  * To check if you already have an SSH key, please enter the following command at a command prompt.
   
         ls ~/.ssh
   * If you see an id_rsa.pub file, then you have an SSH key.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -97,9 +97,15 @@ Complete all of these tasks on your local home machine.
 * Streisand requires a BSD, Linux, or macOS system. As of now, Windows is not supported. All of the following commands should be run inside a Terminal session.
 * Python 2.7 is required. This comes standard on macOS, and is the default on almost all Linux and BSD distributions as well. If your distribution packages Python 3 instead, you will need to install version 2.7 in order for Streisand to work properly.
 * Make sure an SSH key is present in ~/.ssh/id\_rsa.pub.
+  * SSH keys use public-key cryptography to prove your identity to a server or service.
```
cpu: I'm not sure that someone unfamiliar with SSH keys will find much clarity in a reference to public-key cryptography. In counter-balance, we don't have a ton of space in a README to really teach a whole course on anything.

How about:
"SSH keys are a more secure alternative to passwords that allow you to prove your identity to a server or service built on public key cryptography."?

I think that maintains the intent of your edit but also makes it clear that conceptually a new user can think of it as a more secure password without really losing much.

README.md
```diff
@@ -97,9 +97,15 @@ Complete all of these tasks on your local home machine.
...
+  * To check if you already have an SSH key, please enter the following command at a Bash prompt.
```
cpu: Instead of "at a Bash prompt" maybe it would be better to say something along the lines of "in a terminal at a command prompt". I think most users unfamiliar with SSH keys won't know about specific shells and would be using a GUI to navigate to something like Konsole or another _terminal_ program. What do you think?

Frichetten:
Thanks for the awesome feedback @cpu! I agree, it's tricky to balance writing for beginners without being overly verbose. I think your edit covers the issue very well. I will role these in and submit a new pull request tomorrow :)

nopdotcom:
The only thing I can think of is that users should save a copy of the RSA private key if they want to log in later. I'm not concerned enough about this to file a PR though. :-)

cpu:
@nopdotcom Good thoughts. Something mentioning that the private key is A) important to keep secret (like a password) and B) required for all future access (like a password) would be a good follow up

PR #1169, opened 2018-01-29
Switch passwords back to BIP generation

nopdotcom:
Now that the Tor bridge name is fixed, we can go back to the newer password generator.


nopdotcom:
I think the Tor bridge name generator *has* been using the new name generation code...

PR #1177, opened 2018-02-02
Make LE email address optional

emirpprime:
- Allow LE role to run when no email is provided
- Update copy to reflect this but emphasise desirability

### DIFF
--- a/playbooks/lets-encrypt.yml
+++ b/playbooks/lets-encrypt.yml
@@ -30,10 +30,13 @@
 
     - name: "streisand_admin_email_var"
       prompt: |
-        Which email address do you want to use as contact of Streisand server?
+        Which email address do you want to use as contact for the Streisand
+        server's Let's Encrypt certificate?
 
-        This is an optional question. The email is only used as contact method
-        of your Let's Encrypt account.
+        This is an optional question. Let's Encrypt will issue a certificate
+        without providing an email address, however it is highly recommended
+        you do. The email is only used as contact method of your Let's
+        Encrypt account.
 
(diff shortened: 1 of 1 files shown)


playbooks/lets-encrypt.yml
```diff
@@ -30,10 +30,13 @@
 
     - name: "streisand_admin_email_var"
       prompt: |
-        Which email address do you want to use as contact of Streisand server?
+        Which email address do you want to use as contact for the Streisand
+        server's Let's Encrypt certificate?
 
-        This is an optional question. The email is only used as contact method
-        of your Let's Encrypt account.
+        This is an optional question. Let's Encrypt will issue a certificate
```
nopdotcom: Here's some text to steal from:

"The Let's Encrypt service issues HTTPS certificates automatically. If you supply an email address, Let's Encrypt will send you important (but very infrequent) notifications about using the certificate. This mail includes any upcoming certificate expirations, and important changes to the Let's Encrypt service."

Does Let's Encrypt have accounts?

Thanks for hitting this, btw. The user interview is really important.

emirpprime: On Accounts: https://letsencrypt.org/docs/account-id/

emirpprime: Thanks - that text looks good, and official. I've used most bar the first sentence, and added left the 'won't be used for anything else' at the end.

cpu:
Will merge this momentarily. Just waiting on a test provision with an empty ACME contact email to complete.

PR #1193, opened 2018-02-15
Add OpenVPN tlscrypt support

alimakki:
- configure OpenVPN to use the tls-crypt directive.
 - remove key-direction as it is no longer relevant

Tested on:

- [x] iOS
- [x] Linux cli (Ubuntu 17.10)
- [ ] Windows
- [x] Android
- [ ] macOS


cpu:
@alimakki I'll review & test from Android shortly. Thanks @alimakki !

alimakki:
@cpu thanks.

It just occurred to me that enabling this will likely break compatibility with 2.3x clients.

Something to consider.

cpu:
> It just occurred to me that enabling this will likely break compatibility with 2.3x clients.

I'm having a hard time deploying weak google-fu to find out when 2.3.0 was released. @alimakki  Any thoughts on how much this matters?

cpu:
I was able to successfully test this on Android using OpenVPN 0.6.73. Sorry for the delay!

cpu:
>> It just occurred to me that enabling this will likely break compatibility with 2.3x clients.
> I'm having a hard time deploying weak google-fu to find out when 2.3.0 was released. @alimakki Any thoughts on how much this matters?

My current inclination is to merge as-is. If there are complaints about 2.3.x incompatible we can revisit and perhaps template `tls-crypt` by default with an option to switch to the legacy format if you prefer supporting old clients vs having better security/privacy.

alimakki:
@cpu, the 2.3 branch was released in Jan 2 2013, see https://github.com/OpenVPN/openvpn/tree/v2.3.0

cpu:
@alimakki Wow that's really old! OK, let's merge this :-) Thanks!

PR #1197, opened 2018-02-18
Updated DigitalOcean Size id

samitok:
Old value sets-up a 10$ droplet, it is shown as legacy size after creation.


alimakki:
Thank you @samitok :+1:

PR #1198, opened 2018-02-19
Modify the Gateway documentation to reflect updates to passphrase generation and Tor modularity

jlund:
* We're now using the BIP-0039 wordlist.
* Role modularity means that many users will not have Tor installed and there won't be multiple connection options.


jlund:
Oh yeah, of course. Done!

PR #1209, opened 2018-03-05
Enable TCP BBR congestion control

pguizeline:
From the discussion https://github.com/StreisandEffect/discussions/issues/6
Description: https://research.google.com/pubs/pub45646.html

This PR enables BBR support on Ubuntu 16 LTS, the downside(?) is that it bumps the kernel to 4.13, and the HWE branch suffers a lot of updates, and it could break installs on some providers.

After Streisand is installed, the user needs to reboot so the new kernel is used and BBR is enabled. I've tested on a localhost install on DigitalOcean and everything is running, it would be great to run some benchmarks on the actual gain from using BBR.

### DIFF
--- a/playbooks/roles/common/vars/main.yml
+++ b/playbooks/roles/common/vars/main.yml
@@ -33,6 +33,8 @@ streisand_common_packages:
   # A UUID generator with an explicit random function.
   # Used for generating UUIDs for mobileconfig files
   - uuid
+  # Install Ubuntu LTS Enablement Stacks (HWE or Hardware Enablement) kernel
+  - linux-generic-hwe-16.04
 
 # Services that are running by default but not needed by Streisand
 streisand_unneeded_packages:
--- a/playbooks/roles/sysctl/vars/main.yml
+++ b/playbooks/roles/sysctl/vars/main.yml
@@ -27,3 +27,5 @@ sysctl_values:
   - { key: fs.suid_dumpable, value: 0 }
   - { key: fs.protected_hardlinks, value: 1 }
   - { key: fs.protected_symlinks, value: 1 }
+  - { key: net.core.default_qdisc, value: fq }
+  - { key: net.ipv4.tcp_congestion_control, value: bbr }


z0rc:
I'm not going to comment on PR in general, but

> user needs to reboot

This can be and should be done via ansible.

nopdotcom:
I'm reluctant to bump to HWE without *at least* anecdotal evidence that BBR is significant. 

My guess is all the protocols with TCP ingress could probably use it the most, given the network is probably much smoother on the server<->Internet side.

The question after that is, "do we make BBR standard, optional, or a hidden option".

cpu:
@pguizeline I agree with @nopdotcom's assessment. Moving to HWE is a fairly drastic change for this performance optimization. Having to factor a reboot into the provisioning process will also be a headache, whether it is done by Ansible or by manual prompting.

Overall my feeling is that the Streisand project is in a place right now where maintainer time is low and stability should be preferred over features. In this case the feature is a performance optimization that, while nice, won't fundamentally improve access to the unfiltered internet for most users.

@pguizeline Are you open to adding a `.patch` diff and a short "HOWTO" to the Wiki section of the Streisand project about how to enable TCP BBR if desired? I think that would benefit power users that are interested in the optimization while also removing the maintenance burden.

pguizeline:
I agree with you guys! Going to the HWE kernel is probably going to do more harm than good. In fact even writing a patch will only create more problems along the way. I can write a PR for the WIKI on a quick how to, but even then I feel like it's going to cause more harm than good. In the end, I think that we should just scrape this whole thing and wait for the new Ubuntu, when enabling BBR is trivial.

cpu:
@pguizeline Sounds like everyone is in agreement. I'm going to close this PR for now and you can work on a Wiki page if you feel up to it. I hope we'll be on a version of Ubuntu with consistent support in the near future. I'm also keeping an eye on how other projects are addressing the same question (e.g. https://github.com/Jigsaw-Code/outline-server/issues/15 ).

PR #1219, opened 2018-03-21
Convert SSLH to use config file vs cmd line params.

### LINKED ISSUE #827: Convert SSLH systemd defaults to use SSLH config file
### Expected behavior:

The SSLH configuration Streisand uses should be generated in `/etc/sslh.cfg`. See the [upstream example config](https://github.com/yrutschle/sslh/blob/master/example.cfg) for more information.

### Actual Behavior:

Streisand deploys a [Systemd defaults file](https://github.com/jlund/streisand/blob/6f0a4ea725e4202a0469b7457054675d822bee0f/playbooks/roles/sslh/templates/sslh.j2) that overrides the command line `DAEMON_OPTS` for SSLH configuration instead of relying on `/etc/sslh.cfg`.

cpu:
This commit updates the `sslh` playbook such that the
`/etc/default/sslh` systemd defaults specify a `-F` flag with
a configuration file instead of a bunch of individual command line
options. The new configuration is read from `/etc/sslh.cfg` and contains
the same options as were previously specified by command line.

This is slightly cleaner overall and will be easier to extend in the
future. This also unblocks the usage of `log_level` when SSLH is
upgraded to a version > 1.18.

Note: specifying `probe: "builtin"` for each protocol is only 
required [pre 1.18](https://github.com/yrutschle/sslh/issues/121#issuecomment-341899139). For the version shipped in Streisand currently (1.17-2) 
it is required.

Resolves https://github.com/StreisandEffect/streisand/issues/827


cpu:
Ran a test deploy on Vultr. Confirmed that Port 443 muxing still functions. SSH, HTTPS, and OpenVPN were all able to connect to :443 and reach their respective services.

alimakki (approved):
LGTM 👍

cpu:
Closing and re-opening. I think CI got caught up when I deleted a branch I pushed to the StreisandEffect remote instead of my fork.

f1h0:
2 years have passed and you did not realize it for some reason? Why so long

PR #1241, opened 2018-03-28
Search for "my ip address", since "ip address" sucks now

nopdotcom:
I think Google used to give you your connection's current address when you searched for `ip address`. That seems to be sorta broken now. Searching for `my ip address` gives you an info box with the address right at the top.

I am totally up for a better place to put the variable.

### DIFF
--- a/global_vars/vars.yml
+++ b/global_vars/vars.yml
@@ -8,3 +8,5 @@ streisand_client_test: no
 streisand_site_vars: "{{ lookup('env','HOME') }}/.streisand/site.yml"
 
 gpg_key_server_address: "x-hkp://pool.sks-keyservers.net"
+
+streisand_my_ip_url: https://encrypted.google.com/search?hl=en&q=my%20ip%20address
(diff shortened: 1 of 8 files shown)


global_vars/vars.yml
```diff
@@ -8,3 +8,5 @@ streisand_client_test: no
 streisand_site_vars: "{{ lookup('env','HOME') }}/.streisand/site.yml"
 
 gpg_key_server_address: "x-hkp://pool.sks-keyservers.net"
+
+streisand_my_ip_url: https://encrypted.google.com/search?hl=en&q=my%20ip%20address
```
cpu: Echoing [Ali's comment](https://github.com/StreisandEffect/streisand/pull/1241#issuecomment-376991079) as a formal requested change. Can you use the DDG link instead?

alimakki: +1 on putting setting the url as a var, however (for the sake of consistency), if we put the ip url in global_vars, then all the site vars would need to be updated as well. I think a better place for this would be in the common/vars rather than global_vars.

PR #1242, opened 2018-03-28
Install ocserv from universe repository

alimakki:
This PR updates the OpenConnect role to now install via apt as opposed to compiling from source.
In addition to a speedup in installation time and a simplified codebase, OpenConnect now benefits from automatic upgrades through unattended-upgrades.

The init scripts previously being used have now been replaced wholesale with `systemd` unit files:
 - `ocserv.service` to manage the OpenConnect server
 - `ocserv-iptables.service` to handle firewall rules.

An additional configuration, `tunnel-all-dns = true` has been added to the generated configuration file; better be explicit.

A minor issue in the generated instructions were corrected where anchor links for the openconnect-gui/cli were broken.

Tests:
 - [ ] macOS
 - [x] Linux
 - [ ] Windows
 - [x] Android
 - [x] iOS

Updates #1220, #1223

PR #1248, opened 2018-03-31
Convert OpenVPN/stunnel to use systemd unit files

alimakki:
#### OpenVPN
3 unit files are created:
 - `openvpn@server.service` to manage the tcp server
 - `openvpn@servier-udp.service` to manage the udp server
 - `openvpn-iptables.service` to manage our OpenVPN firewall rules

As the apt repository will install be default an `openvpn.service`, it must be explicitly stopped and disable to avoid issues during a server reboot.

#### stunnel
1 unit file is created
 - `stunnel.service`

By default during stunnel installation, an init script (`/etc/init.d/stunnel`) is created and thus must be explicitly stopped and removed entirely.

This PR builds off of #1242


playbooks/roles/openconnect/tasks/firewall.yml
```diff
@@ -12,16 +12,15 @@
...
-    dest: /etc/init.d/streisand-openconnect
-    mode: 0755
+    src: ocserv-iptables.service.j2
+    dest: /etc/systemd/system/ocserv-iptables.service
+    mode: 0644
 
-- name: "Enable the streisand-openconnect init service"
-  service:
-    name: streisand-openconnect
+- name: Enable the ocserv-iptables service
```
cpu: I have the distinct memory of finding that the `systemd` task didn't handle specifying `enabled` and `state` at the same time very well. It seems like this works correctly in CI so perhaps I'm remembering wrong?

alimakki: That could be, however running it against a real server seemed to work fine in my case. Don't really have an issue with making them separate either.

cpu: I think its fine to leave as-is. I'm not sure why I thought it was useful to comment with superstitious observations :laughing:

playbooks/roles/openvpn/handlers/main.yml
```diff
@@ -1,5 +1,8 @@
 ---
 - name: Restart OpenVPN
-  service:
-    name: openvpn
+  systemd:
+    name: "{{ item }}"
     state: restarted
+  with_items:
```
cpu: The two service names in this `with_items` are duplicated in a couple other places. Can you move them into a var in `playbooks/roles/openvpn/vars/`?

playbooks/roles/openvpn/tasks/main.yml
```diff
@@ -126,13 +126,37 @@
...
+    dest: "{{ openvpn_path }}/server.conf"
 
 - name: Copy OpenVPN UDP configuration file into place
   template:
     src: etc_openvpn_server_udp.conf.j2
-    dest: /etc/openvpn/server-udp.conf
-  notify: Restart OpenVPN
+    dest: "{{ openvpn_path }}/server-udp.conf"
+
+- name: Stop and disable openvpn.service
```
cpu: It might avoid some confusion if this task name specified that it was stopping and disabling the stock/upstream service.

playbooks/roles/openvpn/tasks/main.yml
```diff
@@ -126,13 +126,37 @@
...
+    name: openvpn.service
+    state: stopped
+    enabled: no
+
+- name: Copy the OpenVPN system unit files
+  template:
+    src: openvpn.service.j2
+    dest: /etc/systemd/system/{{ item }}
+    mode: 0644
+  with_items:
```
cpu: this `with_items` can reference the earlier mentioned var

playbooks/roles/openvpn/tasks/main.yml
```diff
@@ -126,13 +126,37 @@
...
+    - "openvpn@server.service"
+    - "openvpn@server-udp.service"
+
+- name: Enable the OpenVPN services
+  systemd:
+    daemon_reload: yes
+    name: "{{ item }}"
+    enabled: yes
+    state: restarted
+  with_items:
```
cpu: this `with_items` can reference the var as well.

playbooks/roles/stunnel/templates/stunnel.service.j2
```diff
@@ -0,0 +1,19 @@
...
+[Install]
+WantedBy=multi-user.target
+Alias=stunnel.target
+
+[Service]
+Type=forking
+ExecStart=/usr/bin/stunnel {{ stunnel_path }}/stunnel.conf
+ExecStop=/usr/bin/killall -9 stunnel
+
+# Give up if ping don't get an answer
```
cpu: Typo: "if ping don't get an answer" -> "if ping doesn't get an answer within the timeout"

playbooks/roles/stunnel/templates/stunnel.service.j2
```diff
@@ -0,0 +1,19 @@
...
+TimeoutSec=600
+
+Restart=always
+PrivateTmp=false
```
cpu: Out of curiosity: have you tested stunnel with `PrivateTmp=true`? It looks like `PrivateTmp=false` is present in the linked source unit but without a comment. Maybe there's a good reason for it? To me without justification it seems like a mild red flag since the global `/tmp` is untrustworthy.

playbooks/roles/openvpn/templates/openvpn-iptables.service.j2
```diff
@@ -0,0 +1,14 @@
+[Unit]
+Description=Set the firewall rules required for openvpn
```
cpu: I think `Description=Firewall rules required for OpenVPN` would be a little bit clearer. Right now this shows up a little wonky in `journalctl` output:
>`systemd[1]: Started Set the firewall rules required for openvpn`

alimakki:
@cpu thanks for the review, I got all the requested changes in. I think its ready for another go.

playbooks/roles/openvpn/templates/openvpn-iptables.service.j2
```diff
@@ -0,0 +1,14 @@
+[Unit]
+Description=Firewall rules required for OpenVPN
+After=network.target
+Before={{ openvpn_service_names | join(' ') }}
+
```
alimakki: This one almost slipped by

playbooks/roles/stunnel/templates/stunnel.service.j2
alimakki: Good catch, I’ll test this.

playbooks/roles/openvpn/templates/openvpn-iptables.service.j2
cpu: Good catch!

Elsopuro:
You need to fix service name in this file _streisand/playbooks/roles/stunnel/tasks/main.yml_

\- name: Stop (init.d's) stunnel4
  systemd:
    name: **stunnel.service**
    state: stopped

Issue #1202, opened 2018-02-23
Clarify purpose of multiple OpenVPN profiles & random names


cpu:
Multiple users have been confused by the generated OpenVPN instructions, in particular having confusion about the purpose of the OpenVPN profiles, the differences between them, and the randomly generated names.

We should take a crack at updating the generated docs to clarify these points.

nopdotcom:
A potential bandaid: never use the BIP words for names without a prefix.

`openvpn-profile-poem-walk.ovpn` might help some.

PR #1255, opened 2018-04-07
Allow to use the same certificates for OpenVPN by default

yamaha252:
It's disabled by default although it is written that it is available
> For security reasons, OpenVPN clients typically have their own unique certificate and private key. The server you will be connecting to **has been configured to allow clients to share the same certificate and private key**, but you may still wish to give your phone and laptop different keys, for example.

### DIFF
--- a/playbooks/roles/openvpn/templates/etc_openvpn_server_common.j2
+++ b/playbooks/roles/openvpn/templates/etc_openvpn_server_common.j2
@@ -16,7 +16,7 @@ remote-cert-tls client
 # Allow multiple clients with the same common name to concurrently connect.
 # In the absence of this option, OpenVPN will disconnect a client instance
 # upon connection of a new client having the same common name.
-# duplicate-cn
+duplicate-cn
 
 keepalive 1800 3600
 tls-crypt ta.key # This file is secret


yamaha252:
Hi! I'm sorry for a long reply.
I think the doc I quoted is correct. You cannot share the server's web page for people when `duplicate-cn` is disabled because it might be difficult to explain each person what profile they should use to connect.
In this way I suggest fixing the common readme doc to explain for an admin what profiles are (now it's difficult to understand) and suggest them enabling `duplicate-cn` option if they'd like to share their connection web page.
What do you think?

cpu:
> In this way I suggest fixing the common readme doc to explain for an admin what profiles are (now it's difficult to understand)

@yamaha252 Good point about making it easier to understand what the profiles are. We have an open issue for that: https://github.com/StreisandEffect/streisand/issues/1202 Hearing you say the same thing certainly puts it more towards the front of the TODO list :memo: :top: 

> suggest them enabling duplicate-cn option if they'd like to share their connection web page.

I'm hesitant to make this suggestion in the docs since it degrades security. In the short-term I think addressing #1202 and helping admins understand that profiles shouldn't be shared is the best bet. In the long-term I think we could do some work to treat client profiles as something more like API keys: a user clicks to get a profile, it downloads, and after that it isn't available for anyone else to use or view again.

I'm going to close this PR for now based on the reasoning above. If you're interested in submitting a PR to help with #1202 it would certainly be welcome!!!

Thanks for contributing to Streisand :-)

PR #1262, opened 2018-04-11
Common: use explicit --random-source for shuf

cpu:
Resolves https://github.com/StreisandEffect/streisand/issues/1259

Note: I changed the `shuf` calls to use a variable, which in turn caused quoting issues until I switched to the `>` "Folded block scalar" YAML syntax. Doing _that_ caused an extra `\n` to get tacked on the end and this caused places we used the var in `shell` pipeline expression to break unless there was a `| trim`. Phew! I'm certainly open to cleaner alternatives/suggestions but I want to land a fix to 1259 ASAP.

### DIFF
--- a/playbooks/roles/certificates/tasks/ca-server.yml
+++ b/playbooks/roles/certificates/tasks/ca-server.yml
@@ -23,7 +23,7 @@
     creates: "{{ tls_ca }}.crt"
 
 - name: Generate a random server common name
-  shell: "{{ streisand_word_gen.long_identifier }} > {{ tls_server_common_name_file }}"
+  shell: "{{ streisand_word_gen.long_identifier | trim }} > {{ tls_server_common_name_file }}"
   args:
     creates: "{{ tls_server_common_name_file }}"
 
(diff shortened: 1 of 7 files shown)


cpu:
@nopdotcom Any qualms about seeing this merged?

nopdotcom:
I'm glad I didn't have to figure that out! TBH, I'd probably have punted,
and just cut&pasted into the vars, but this looks better.

What you went through sounds complicated enough that you ought to put some
of the commit message in as a comment; the next person to change this may
be puzzled.
-- 

-- 
Jay

cpu:
> I'm glad I didn't have to figure that out! TBH, I'd probably have punted, and just cut&pasted into the vars, but this looks better.

Thanks! I was awfully tempted by this idea towards the end of YAML struggling...

> What you went through sounds complicated enough that you ought to put some of the commit message in as a comment; the next person to change this may be puzzled.

Good feedback! Thank you. Addressed in d436d99

cpu:
I ran a couple test provisions of this branch using the full-default site-vars and had no issues. Going to merge when CI completes :+1:

PR #1270, opened 2018-04-15
nginx systemd

alimakki:
Create a systemd unit file and deprecate usage from monit.

### DIFF
--- a/playbooks/roles/monit/templates/monitrc.j2
+++ b/playbooks/roles/monit/templates/monitrc.j2
@@ -11,17 +11,6 @@ set httpd port 2812 and
     # and requiring basic auth (through nginx) should be sufficient to protect this interface.
     allow 0.0.0.0/0.0.0.0 # allow any ip to connect to monit.
 
-# Nginx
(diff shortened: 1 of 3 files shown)


nopdotcom:
Nginx doesn't have this problem, I think, but at least (the ghost of) xl2tpd will spin at 100%, and monit helps. But I've never seen nginx die that way, anywhere.

alimakki:
It's mostly for having `Restart=on-failure`, the default one doesn't have it.

z0rc:
At this case, it's possible to add a drop-in configuration to override the options you need, instead of overriding the whole unit file. See https://coreos.com/os/docs/latest/using-systemd-drop-in-units.html for example.

In general it's better to stick to upstream unit file.

alimakki:
@z0rc thanks for the link, and appreciate the tip.

alimakki:
@z0rc I've taken into consideration your suggestion, would you mind taking a quick peek?

z0rc:
Yeah, LGTM :+1:

PR #1277, opened 2018-04-17
Always customize installation

### LINKED ISSUE #1225: Make custom setup instructions the default
Streisand's modularization effort allows a user to pick and choose which VPN daemons get installed and where supported, how many client configurations to generate.

This option only presents itself if a user elects to customize the install, which many seem to be unaware of and opt for a default installation.

The objective would be to make modular options part of the interactive process thus making it more prominent during Streisand's initial setup.

alimakki:
Make customizing the install a default action, answering `no` will read from the defined site-vars.

Resolves #1225

### DIFF
--- a/playbooks/customize.yml
+++ b/playbooks/customize.yml
@@ -9,7 +9,7 @@
       default: "~/.ssh/id_rsa"
       private: no
     - name: vpn_clients
-      prompt: "How many VPN client profiles should be generated per-service? Press enter for default "
+      prompt: "How many VPN client profiles should be generated per-service (min: 1 max: 20)? Press enter for default "
       default: 5
       private: no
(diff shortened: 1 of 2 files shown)


playbooks/customize.yml
```diff
@@ -9,7 +9,7 @@
       default: "~/.ssh/id_rsa"
       private: no
     - name: vpn_clients
-      prompt: "How many VPN client profiles should be generated per-service? Press enter for default "
+      prompt: "How many VPN client profiles should be generated per-service (min: 1 max: 20)? Press enter for default "
```
cpu: Gosh I wish we could use vars in prompt strings :'(

alimakki: If everything was perfect life would be boring ;)

PR #1287, opened 2018-04-22
bring back setting permissions to ssh key

zlatin:
Looks like it was removed accidentally.


cpu:
Good catch, thanks @zlatin!

PR #1299, opened 2018-04-24
Temporary pins of Ansible, SecretStorage versions; add pycrypto

nopdotcom:
Ansible 2.5.1 appears to be broken for us. 2.5.2 looks like it will have a fix. Pin to version 2.5.0 until then.

The SecretStorage module was just bumped to 3.0.0, and now requires Python 3. Until we're ready for Python 3, specify the earlier rev.

GCE requires pycrypto for authentication now.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -29,8 +29,7 @@ dopy==0.3.5
 
 # Google Compute Engine
 apache-libcloud>=1.17.0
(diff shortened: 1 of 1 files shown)


alimakki:
@nopdotcom I actually hit the pycrypto dependency for GCE. I think it would be a good idea to put it in.

cpu:
Thanks for fixing this @nopdotcom!!!

PR #1305, opened 2018-04-25
Add `-v` flag to user instructions for SSH port forwarding

raabrp:
As per [issue 1284](https://github.com/StreisandEffect/streisand/issues/1284), a flag for verbose output is added to the SSH port-forwarding command which a user is instructed to run. This allows for verifying fingerprints without branching instructions for different users.

In addition, the French instructions lacked the `-N` flag for the same command, which has been added.

### DIFF
--- a/playbooks/roles/ssh-forward/templates/instructions-fr.md.j2
+++ b/playbooks/roles/ssh-forward/templates/instructions-fr.md.j2
@@ -114,7 +114,7 @@ Sshuttle est une simple solution de tunneling VPN qui fonctionne sur le transpor
 
 1. `SSH`ez dans le serveur et transférer un port SOCKS dynamique:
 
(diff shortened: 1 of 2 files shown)


playbooks/roles/ssh-forward/templates/instructions-fr.md.j2
```diff
@@ -114,7 +114,7 @@ Sshuttle est une simple solution de tunneling VPN qui fonctionne sur le transpor
 
 1. `SSH`ez dans le serveur et transférer un port SOCKS dynamique:
 
-   `ssh -D {{ ssh_default_socks_port }} forward@{{ streisand_server_name }}`
```
cpu: Nice work catching the missing `-N`! :eagle: :mag:

Issue #1295, opened 2018-04-23
Existing server: Provisioning should not require SSH public key


Lexy2:
Expected behavior:
Streisand should not need a public key for existing server if connection via private key succeeds.
It should rather check remote server's ~/.ssh/authorized_keys and pull the public keys from there.

Actual Behavior:
Existing server connection fails.

Steps to Reproduce:
Set up an existing server.
Set up certificate authentication for SSH logon.
Run server provisioning.
Installation fails because id_rsa.pub is missing.
[ contents of streisand-diagnostics.md here ]

Additional Details:
Log output from Ansible or other relevant services (link to Gist for longer output):
PLAY [Perform global variables validation] *****************************************************************************

TASK [validation : Stat the Streisand SSH private key] *****************************************************************
ok: [localhost]

TASK [validation : Fail if the Streisand SSH private key file doesn't exist] *******************************************
skipping: [localhost]

TASK [validation : Stat the Streisand SSH public key] ******************************************************************
ok: [localhost]

TASK [validation : Fail if the Streisand SSH public key file doesn't exist] ********************************************
fatal: [localhost]: FAILED! => {"changed": false, "msg": "The Streisand SSH public key \"~/.ssh/id_rsa.pub\" does not exist."}

TASK [validation : fail] ***********************************************************************************************
fatal: [localhost]: FAILED! => {"changed": false, "msg": "Ensure you specified an existing SSH private key file (not public).\n Try using `ssh-keygen -f ~/.ssh/id_rsa to generate your key if it does not exist\n"}
        to retry, use: --limit @/home/lexy/streisand/playbooks/validate.retry

PLAY RECAP *************************************************************************************************************
localhost                  : ok=2    changed=0    unreachable=0    failed=2

Target Cloud Provider:
custom

Operating System of target host:
Ubuntu 16.04

Operating System of client:
Linux Mint 18.3

Version of Ansible, using ansible --version :
ansible 2.5.1
config file = /home/lexy/streisand/ansible.cfg
configured module search path = [u'/home/lexy/.ansible/plugins/modules', u'/usr/share/ansible/plugins/modules']
ansible python module location = /usr/local/lib/python2.7/dist-packages/ansible-2.5.1-py2.7.egg/ansible
executable location = /usr/local/bin/ansible
python version = 2.7.12 (default, Dec 4 2017, 14:50:18) [GCC 5.4.0 20160609]

Output from git rev-parse HEAD in your Streisand directory :
[...]

PR #1306, opened 2018-04-25
Disabled pub key existence checking when provisioning existing server…

Lexy2:
… and localhost.

Done by adding a parameter to ~/.streisand/site.yml which is checked at validation and pub key checking is skipped if parameter is defined.


Lexy2:
Fix for #1295

Lexy2:
Hi @cpu , 

Thank you very much for the comments. Yes, your approach looks prettier. Let me amend the PR.

Regards, Lexy

Lexy2:
Hi @cpu,

I have pushed the amendments here. Travis fails because the no_check_pub_key (renamed to streisand_new_server_provisioning) is not define per se in any vars and is only passed in ./streisand script.

Would you like me to test if the var is defined or add a default value to default inventory?

Thanks.

Lexy

cpu:
Hi @Lexy2,

Thanks for pushing those changes. It looks great :+1:  I like the new var name.

> Would you like me to test if the var is defined or add a default value to default inventory?

Historically I think we've handled this by adding a default value but its a bit messy and needs to be done in a couple spots. Let's try the alternative you suggest and consider `no_check_pub_key` enabled if its defined at all. I think that's reasonable for this case .

Lexy2:
Thanks, I suppose when you said
> consider no_check_pub_key enabled if its defined at all
you meant
consider `streisand_new_server_provisioning` enabled if undefined.

cpu:
> you meant consider streisand_new_server_provisioning enabled if undefined.

@Lexy2 Oops, yes :-) Sorry to confuse things. You got it.

Lexy2:
Hi @cpu ,

Please kindly review the proposed changes.
I tried your variant of testing if `streisand_new_server_provisioning` undefined but I think setting a default value is a better option:

- code is prettier
- the change is still local to validation task
- easier overridden since defaults have the lowest priority

Thank you!

Regards, Lexy

cpu:
> Please kindly review the proposed changes.

Hi @Lexy2 - great! Thank you. I'll try to take a look at this PR this evening after work.

> I tried your variant of testing if streisand_new_server_provisioning undefined but I think setting a default value is a better option

Ok! I'm not opposed to handling it this way :+1:

Lexy2:
Thanks for guiding me through my first one, @cpu !

Appreciate your feedback.

PR #1314, opened 2018-04-29
GPG: explicitly wait for dirmngr and gpg-agent sockets.

cpu:
#### GPG: explicitly wait for dirmngr and gpg-agent sockets.

A user provisioning Streisand on Azure [reports a persistent failure](https://github.com/StreisandEffect/streisand/issues/1308#issue-318237314) during the Streisand GPG keyring import task with the following error msg:
> "gpg: can't connect to the agent: IPC connect call failed"

This is (likely?) caused by the gpg-agent not having started by the time we import the keys. I can't reproduce the bug myself so this may be a red herring.

This PR changes the gpg playbook to explicitly start both the gpg-agent and the dirnmgr instead of letting the first gpg operation do so. This PR also adds an explicit `wait_for` task for both the dirnmgr and gpg agent control sockets to make sure we don't proceed until both daemons are available.

#### GPG: Use explicit English locale for gpgv2.

The download-and-verify playbook invokes `gpg2` to check GPG signatures. In addition to checking the return code for success we regex match the key ID that was used to successfully validate the signature. The successful match of this regex depends on the `gpgv2` output being in English. If a user has override their `LANG` to something else then the regex will fail to match! See: https://github.com/StreisandEffect/streisand/issues/1199#issuecomment-378677563 for more.

This PR addresses this case by explicitly forcing the locale to "en_US.UTF-8" when invoking `gpgv2`, ignoring the system locale configuration.

Its likely bugs like this exist other places in Streisand but at least this one is fixed!

Updates https://github.com/StreisandEffect/streisand/issues/1308

### DIFF
--- a/playbooks/roles/download-and-verify/tasks/main.yml
+++ b/playbooks/roles/download-and-verify/tasks/main.yml
@@ -23,6 +23,15 @@
 
 - name: "Verify the {{ project_name }} download signatures with the Streisand GPG keyring"
   command: "gpgv2 --keyring {{ streisand_gpg_keyring }} {{ project_download_location }}/{{ item.sig }} {{ project_download_location }}/{{ item.file }}"
+  environment:
+    # We need to explicitly override the configured LANG, LC_MESSAGES and LC_ALL
+    # env vars to ensure that we can match on English output even on systems
+    # that have a non-english locale configured. It's _probably_ enough to
+    # override just LANG but it sounds like there are some cases where LC_ALL
+    # takes precedence and setting env vars is cheap!
+    LANG: "en_US.UTF-8"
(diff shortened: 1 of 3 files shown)


cpu:
> @cpu gonna give this a quick test

Great, thanks. I didn't mention it in the PR desc but I'd also prefer to wait on merging this until @ekifox has a chance to see if it actually helped their problem or not. I'll put a label on this PR to remind myself not to merge it early :-)

cpu:
This was confirmed to fix the GPG IPC error for Azure deployment. As soon as it's been reviewed and CI is happy it's ready to merge :ballot_box_with_check:

PR #1267, opened 2018-04-14
Limit TLS ciphersuites

alimakki:
Limit TLS ciphersuites within Streisand to a more modern and explicit configuration.

OpenVPN:
A list of available cipher suites supported by OpenVPN can be obtained by invoking: `openvpn --show-tls` which prints a list of supported suites in order of preference. As Streisand doesn't generate ECDSA certificates we can limit Streisand's configuration to the strongest available one `TLS-ECDHE-RSA-WITH-AES-256-GCM-SHA384` 

Stunnel:
 - Add an additional configuration to disable TLSv1.1 (therefore only TLSv1.2 is allowed)
 - Limit ciphers based on [Mozilla's modern compatibility  profile](https://wiki.mozilla.org/Security/Server_Side_TLS#Modern_compatibility), and removing ECDSA ciphersuites as Streisand's generated certificates utilize RSA.

Nginx:
 - Only allow TLSv1.2
 - Use the same ciphersuite  list based on the modern compatibility profile.

OpenConnect:
 - Only allow TLSv1.2
 - Use the Secure 192 profile from the [gnutls priority strings](https://gnutls.org/manual/html_node/Priority-Strings.html)

Tests:
* OpenVPN direct
   - [x] iOS
   - [ ] Android
   - [ ] macOS
   - [ ] Windows
   - [x] Linux

* OpenVPN stunnel
   - [ ] Android
   - [ ] macOS
   - [ ] Windows
   - [x] Linux

* OpenConnect
   - [x] iOS
   - [ ] Android
   - [ ] macOS
   - [ ] Windows
   - [x] Linux

Addresses  #1261

Edit: Fixed Mozilla URL

PR #1315, opened 2018-04-29
Remove diffie-hellman role

alimakki:
As of PR #1267 all TLS ciphersuites leverage ECDHE for key agreement, obsoleting the need to generate dh primes, with the added benefit of speeding up a Streisand build.

### DIFF
--- a/playbooks/lets-encrypt.yml
+++ b/playbooks/lets-encrypt.yml
@@ -14,32 +14,34 @@
   vars_prompt:
     - name: "streisand_domain_var"
       prompt: |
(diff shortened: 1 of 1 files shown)


alimakki:
FWIW tested on DO with OpenVPN only.

Can access the gateway, no problems with OpenVPN either.

cpu:
> FWIW tested on DO with OpenVPN only.

:thumbsup: I'm reasonably confident in this PR from a :eyes: only pass. If you and CI are both happy feel free to merge.

Issue #1272, opened 2018-04-17
Digital Ocean / access tokens text is confusing


dandaka:
### Expected behavior:

Text in wizard is consistent with DO website text.

New API access tokens → New Personal Access Tokens (or New API Personal Access Tokens)

### Actual Behavior:

Text in wizard:

```
New API access tokens can be generated in the DigitalOcean control panel.
https://cloud.digitalocean.com/settings/applications
  * When generating a new token, please note that the Write Scope is necessary
    in order to create Droplets.
```

Text in DO website:

```
Personal Access Tokens
Personal access tokens function like a combined name and password for API Authentication. Generate a token to access the DigitalOcean API
```

### Steps to Reproduce:

1. Launch Streisand
2. Select Digital Ocean
3. Find a step, where it asks about API key

cpu:
@dandaka Do you think the updated text proposed by @Frichetten in #1317 resolves your observation? Thanks!

dandaka:
@cpu Yes, thank you!

PR #1317, opened 2018-05-01
Improved DO Personal Access Token Documentation for #1272

Frichetten:
I tried to explain what the Personal Access Token is, as well as fix the naming convention (before we were just calling it an API Access Token. This was for issue #1272


playbooks/digitalocean.yml
```diff
@@ -46,7 +46,7 @@
       private: no
 
     - name: "do_access_token_entry"
-      prompt: "\n\nNew API access tokens can be generated in the DigitalOcean control panel.\nhttps://cloud.digitalocean.com/settings/applications\n  * When generating a new token, please note that the Write Scope is necessary\n    in order to create Droplets.\n\nIf this field is left blank, the environment variable DO_API_KEY\nwill be used.\n\nWhat is your DigitalOcean Access Token?\n"
+      prompt: "\n\nPersonal Access Tokens allow Streisand to create a droplet for you.\nNew Personal Access Tokens can be generated in the DigitalOcean control panel.\nTo generate a new token please do the following:\n  * Go to https://cloud.digitalocean.com/settings/applications\n  * Click 'Generate New Token'\n  * Give the token a name (it is arbitrary)\n  * Be sure to select the 'Write' scope as well (this is not optional)\n  * Click 'Generate Token'\n  * Copy the long string that is generated and paste it below.  \n\nIf this field is left blank, the environment variable DO_API_KEY\nwill be used.\n\nWhat is your DigitalOcean Personal Access Token?\n"
```
cpu: @Frichetten This diff looks good but I'd love if we could also take this as a chance to fix the long line. 

I _think_ you could do something like:
```
prompt: |


Personal Access Tokens allow Streisand to create a droplet for you.
New Personal Access Tokens can be generated in the DigitalOcean control panel.
....
```

This is the "Literal Block Scalar" notation described in [the Ansible YAML docs](https://docs.ansible.com/ansible/latest/reference_appendices/YAMLSyntax.html). You'd want literal (`|`) and not folded (`>`) in this case to preserve the newlines.

Are you open to giving that a shot? Thanks!

Frichetten:
@cpu thanks for the feedback! I wasn't aware that YAML had that kind of Syntax. Would cutting long lines into something like this be desirable for other parts of the project? I'd be happy to go in and fix them up?

PR #1318, opened 2018-05-02
Center the Streisand logo in the Readme

Frichetten:
Okay, I know this is a really weird PR. But for a while now It has bugged me that the Streisand logo in the Readme was not centered. It is aligned to the left. Feel free to reject this PR but GitHub supports HTML in MD files so I just centered the image. I also retained the alt text if that is a requirement

Before:
image

After:
image

Additionally, if you want to see a live render of it visit https://github.com/Frichetten/streisand/tree/center_readme_logo which should be my branch with this change.


Frichetten:
@alimakki thanks for the feedback! I completely forgot about the other READMEs. I have applied the same centering to the logos in the other READMEs. :)

alimakki:
Thank you @Frichetten

PR #1322, opened 2018-05-05
Add route command to drop packets to ec2 metadata instance

### LINKED ISSUE #1260: EC2 instance metadata accessible through authenticated VPN access
### Expected behavior:

While Streisand **does not deploy any IAM roles with metadata accessible secrets** a Streisand instance deployed on AWS should not allow authenticated VPN clients access to instance metadata.

Streisand should block this access in the event advanced users configure IAM roles above-and-beyond the Streisand configuration with secrets that could be leaked. 

### Actual Behavior:

[AWS instance metadata](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-metadata.html) is accessible at `http://169.254.169.254` through the Streisand VPN services.

### Steps to Reproduce:
1. Deploy to AWS
2. Create some kind of IAM role
3. Authenticate to a VPN service (e.g. Shadowsocks)
4. Access  http://169.254.169.254/latest/meta-data/iam/security-credentials/<ec2IamRole>.

### Proposed Remediation

Streisand should persistently null-route the EC2 metadata host on the Streisand server during the EC2 instance provisioning: e.g. with `route add -host 169.254.169.254 reject`

alimakki:
Resolves #1260


alimakki:
should be verified for correctness

nopdotcom:
I think this needs to be a persistent firewall rule, at least after provisioning finishes.

playbooks/ec2-metadata-instance.yml
```diff
@@ -0,0 +1,10 @@
+---
+- name: Drop packets to the Amazon EC2 metadata instance
+  hosts: streisand-host
+  gather_facts: no
+
+  remote_user: "root"
+  become: true
+
+  tasks:
+    - raw: route add -host 169.254.169.254 reject
```
cpu: > @nopdotcom  I think this needs to be a persistent firewall rule, at least after provisioning finishes.

@alimakki I believe @nopdotcom is correct about this. I don't think this will persist across reboots. I think you can achieve persistence ifupd or systemd.

alimakki:
@cpu, taken your feedback, this PR is ready for another look 🔎

playbooks/amazon.yml
```diff
@@ -114,5 +114,7 @@
 
 - import_playbook: ssh-setup.yml
 - import_playbook: cloud-status.yml
+- import_playbook: python.yml
```
nopdotcom: I was tearing my hair out trying to find this one bit of magic--I couldn't get the control flow right. Whew.

alimakki: Yeah that's originally why I used raw commands since python.yml will be included as part of streisand.yml, and I wanted to avoid doing things twice, kind of a necessary evil.

PR #1327, opened 2018-05-09
Change "Localhost" to "localhost"

imcdona:
Use "localhost" instead. I got confused and thought there was a cloud provider called "Localhost" so I selected option 8 (Existing Server) when I should have selected option 7.  Everyone is used to seeing localhost when referring to localhost. Localhost with a capital L is a proper name.

### DIFF
--- a/streisand
+++ b/streisand
@@ -187,7 +187,7 @@ read -r -p "Which provider are you using?
   4. Google
   5. Linode
   6. Rackspace
-  7. Localhost (Advanced)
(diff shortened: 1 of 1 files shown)


nopdotcom:
I couldn't figure out what "SSHD" was (solid state hybrid disk?) so no objection here.

cpu:
I will merge this shortly. I have one PR ahead of this to merge. Afterwards I will pull master into this branch and merge once CI finishes.

PR #1335, opened 2018-05-17
ensure gpgagent is dead before trying to start it Fixes #1333

### LINKED ISSUE #1333: starting gpg agent fails if it is already running
### Expected behavior:
To restart when attempting to start the GPG agent and it's already running
### Actual Behavior:
exits with non zero return value
### Steps to Reproduce:
1. start the setup for existing server
2. break the setup after gpg agent is started
3. retry to setup

### Ansible Information

* Ansible version: 2.5.2
* Ansible system: Linux
* Host OS: Archlinux
* Host OS version:  NA
* Python interpreter: python
* Python version: 3.6.5

### Streisand Information

* Streisand Git revision: c81b4f4500248ad91f480b54a49705c4eed34efe
* Streisand Git clone has untracked changes: no
* Genesis role: existing-server
* Custom SSH key: False

### Enabled Roles

* Shadowsocks enabled:  True
* Wireguard enabled: True
* OpenVPN enabled: True
* stunnel enabled: True
* Tor enabled: True
* Openconnect enabled: True
* TinyProxy enabled: True
* SSH forward user enabled: True
* Configured number of VPN clients: 5

hamidzr:
checks for the status of gpg-agent to see if it is running or not and then act accordingly

### DIFF
--- a/playbooks/roles/gpg/tasks/main.yml
+++ b/playbooks/roles/gpg/tasks/main.yml
@@ -39,6 +39,9 @@
     group: root
     mode: 0750
 
+- name: "Ensure GPG agent is dead"
+  command: "gpgconf --kil gpg-agent"
+
 - name: "Start the GPG agent"
   command: "gpg-agent --homedir {{ root_gpg_dir }} --use-standard-socket --daemon"
 
--- a/playbooks/roles/gpg/tasks/main.yml
+++ b/playbooks/roles/gpg/tasks/main.yml
@@ -39,7 +39,7 @@
     group: root
     mode: 0750
 
-- name: "Start the GPG agent if it's not already started"
+- name: "Ensure gpg-agent is started and running the latest config"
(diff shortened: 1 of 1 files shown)


hamidzr:
yea I'll fix it. The reason I opted for killing it every time through was because I assumed configuration is being changed and like most daemons, it needs to restart/reload to pick up those (potentially) new configs.

cpu:
> yea I'll fix it. The reason I opted for killing it every time through was because I assumed configuration is being changed and like most daemons, it needs to restart/reload to pick up those (potentially) new configs.

@hamidzr That's a good point. I see you've pushed some commits that handle this aspect better. Thank you! I'll try to give this branch a test run & review shortly. Thanks for your patience!

nopdotcom:
I stole your bug in #1387; my apologies.

PR #1340, opened 2018-05-18
Added missing setuptools to pip command

krehl:


nopdotcom:
(There's a minor brain-o in the PR; change `setuptools` to `python-setuptools`.)

OK, I see what's going on, I think. Python setuptools is needed, but it's only `Recommend`ed by `python-pip`.  If you're in an environment where `apt-get` is following `--no-install-recommends` you won't get setuptools. 

This PR isn't a bad idea; it's obsoleted by #1347, but if #1347 isn't accepted we should merge this one.

BTW `./util/venv-dependencies.sh` does check for this situation:

```
# ./util/venv-dependencies.sh ./venv

This system appears to be running Ubuntu or Debian. Checking
for critical packages.

*** Missing package: build-essential
*** Missing package: python-dev
*** Missing package: python-setuptools
Found: python-pip
*** Missing package: python-cffi
*** Missing package: libffi-dev
*** Missing package: libssl-dev
*** Missing package: libcurl4-openssl-dev
-------
Setup will fail without these packages. To install them:

 apt-get install build-essential python-dev python-setuptools python-cffi libffi-dev libssl-dev libcurl4-openssl-dev
```

PR #1345, opened 2018-05-21
wireguard: remove awkward client/server notation

zx2c4:
This had with it a few problems. Some versions of android don't like
config files with a dot in them, it would increase the length of the
interface beyond what it should be, and generally this belies the real
architecture of wireguard. Instead, call test interfaces and the server
interfaces 'wg0' and call each client interface by its profile name
alone.

While we're at it, we clean up the instructions a bit.

----------

This commit is entirely untested.

### DIFF
--- a/playbooks/roles/test-client/tasks/wireguard-profiletest.yml
+++ b/playbooks/roles/test-client/tasks/wireguard-profiletest.yml
@@ -1,17 +1,17 @@
 ---
 - block:
-    - name: "Replace wg0-client with the {{ profile_name }} configuration"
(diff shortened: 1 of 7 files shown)


cpu:
> This commit is entirely untested.

I have an instance coming up now for testing. With that + green CI I'm confident in merging as-is.

cpu:
> I have an instance coming up now for testing. With that + green CI I'm confident in merging as-is.

My instance came up as expected and the client setup instructions in this branch worked without error.

PR #1361, opened 2018-05-27
Support simple IPv6 ingress when Ansible detects existing IPv6

nopdotcom:
If a system has *existing* IPv6 connectivity, support some services over IPv6. (In other words, let's postpone work on *provisioning* IPv6 on servers; that can happen in parallel.)

* sslh: Adding another listen line opens up the gateway (and ssh).

* shadowsocks: Adding another listen address opens the service; obfs works too.

* gateway: Adding the IPv6 address to the self-signed gateway SSL cert works.

Note that systems with IPv6 ingress get IPv6 egress for connection-oriented services for free; check with https://ipv6.google.com/


nopdotcom:
In a branch, I've played with configuring shadowsocks with QR on Android and iOS. It's slick. Just needs another PNG...

nopdotcom:
Consider #1382 instead, which adds user-configurability.

cpu:
@nopdotcom It seems like this should be closed in favour of #1382 - is that reasonable?

nopdotcom:
Yes. This was just two different jumping-off points; if people didn't like the way the other one was going, they could start back from here.

PR #1366, opened 2018-06-04
Adds gcp west europe netherlands

RaymondKroon:
Also adds support for named zones for robust noninteractive deployments. Otherwise it will break after every zone adddition, unless we do not sort the list...

### DIFF
--- a/playbooks/google.yml
+++ b/playbooks/google.yml
@@ -133,11 +133,6 @@
         gce_zone: "{{ zones[gce_zone_idx] }}"
       when: gce_zone is not defined
 
-    - name: Set the Google Compute Engine Zone noninteractive
-      set_fact:
-        gce_zone: "{{ gce_zone }}"
-      when: gce_zone is defined
-
     - name: Register JSON file contents
       command: cat {{ gce_json_file_location }}
       register: gce_json_file_contents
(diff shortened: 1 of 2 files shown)


RaymondKroon:
Ok, I will add some comments in the file, and rename the _var variable
I am really new to Ansible, and I would have liked to remove the var from the config, but if I do that Ansible is complaining the value is missing. And I still don't know how to fix that without the "do not press the button" solution.

nopdotcom:
Don't worry, I'm not very fluent in Ansible either. I'd like the documentation for *my* sake when I fail to understand what's going on here later. :-) Thanks for following up.

playbooks/google.yml
```diff
@@ -121,9 +128,15 @@
...
 
   pre_tasks:
-    - name: Set the Google Compute Engine Zone fact
+    - name: Set the Google Compute Engine Zone interactive
       set_fact:
-        gce_zone: "{{ zones[gce_zone_var] }}"
+        gce_zone: "{{ zones[gce_zone_idx] }}"
+      when: gce_zone is not defined
+
+    - name: Set the Google Compute Engine Zone noninteractive
```
nopdotcom: Hm, wait, do we need this clause at all? Directly above it, we've set gce_zone, so this clause should have the effect of `gce_zone := gce_zone`?

Do I need more coffee?

PR #1369, opened 2018-06-06
Generate OpenWrt WireGuard profiles and instructions

nopdotcom:
There's been some interest in OpenWrt/LEDE support for WireGuard. This adds profiles and full (?) documentation. Works on OpenWrt devices from LEDE 17.01.4 and newer.

### DIFF
--- a/playbooks/roles/wireguard/templates/client-openwrt.txt.j2
+++ b/playbooks/roles/wireguard/templates/client-openwrt.txt.j2
@@ -18,7 +18,7 @@
 
 ############
 
-# If aren't using LuCI, you need the "kmod-wireguard" and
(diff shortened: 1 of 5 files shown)


playbooks/roles/wireguard/templates/client-openwrt.txt.j2
```diff
@@ -0,0 +1,116 @@
...
+# success/failure message in the "Local Startup" box.
+
+# If you're using the web admin UI, you need the "luci-app-wireguard"
+# and "luci-proto-wireguard" packages installed.
+
+# LuCI instructions end here.
+
+############
+
+# If aren't using LuCI, you need the "kmod-wireguard" and
```
alimakki: If aren't -> If you aren't

Issue #1374, opened 2018-06-11
WireGuard not fully upgraded on the fly


nopdotcom:
### Expected behavior:

WireGuard kernel modules reloaded and units restarted on unattended upgrade.

### Actual Behavior:

On upgrade, I see this:

```
You appear to have just upgraded WireGuard from version v0.0.20180513 to v0.0.20180531.
However, the old version is still running on your system. In order to use the
new version, you will need to remove the old module and load the new one.
You can accomplish this with the following commands:

    # sudo rmmod wireguard
    # sudo modprobe wireguard

Do note that doing this will remove current WireGuard interfaces, so you may want
to gracefully remove them yourself prior.

In order to have all wg-quick@.service units restarted and the new module inserted
when a new version is installed, run `touch /etc/wireguard/.reload-module-on-update'.
```

One system told me I needed to `sysctl daemon-reload` too.

### Steps to Reproduce:
1. Force an upgrade from the wireguard PPA.
### Ansible Information

* Ansible version: 2.5.4
* Ansible system: Darwin
* Host OS: MacOSX
* Python interpreter: python2.7
* Python version: 2.7.15

### Streisand Information

* Streisand Git revision: f8d96387c0861c59600b0a947cbeeac0350600f8
* Streisand Git clone has untracked changes: yes
* Genesis role: genesis-amazon
* Custom SSH key: False

### Enabled Roles

* Shadowsocks enabled:  True
* Wireguard enabled: True
* OpenVPN enabled: True
* stunnel enabled: True
* Tor enabled: False
* Openconnect enabled: True
* TinyProxy enabled: True
* SSH forward user enabled: True
* Configured number of VPN clients: 5

tiliarou:
I confirm this behavior, it always ask to remove old module.
Would adding `touch /etc/wireguard/.reload-module-on-update' to the streisand installation help ?

nopdotcom:
@tiliarou: Yeah, adding it would help I think--do you have time to make a PR for it?

nopdotcom:
It might be better to do it with the `file:` task. https://docs.ansible.com/ansible/latest/modules/file_module.html#examples

tiliarou:
I can try to make PR but I'm really not that proficient in coding, I barely know any language to be honest, although ansible seems simple enough.
I assume an ansible `file:` command should come after having installed wireguard and generated all clients configurations ? Or should it come at the very end when we restart all services ?

tiliarou:
[...]

PR #1378, opened 2018-06-13
Added .reload-module-on-update to make wireguard upgrades fully unattended

tiliarou:
Related to WireGuard not fully upgraded on the fly #1374


nopdotcom:
I poked at `/var/lib/dpkg/info/wireguard-dkms.postinst` and this does appear to trigger it. Resolves #1374 .

PR #1382, opened 2018-06-16
Explicit IPv6 config option: sslh/shadowsocks/gateway, plus DO IPv6

nopdotcom:
I'm open to the default being "no".

This PR adds configurability to #1361. In addition to systems where IPv6 is auto-detected, it requests DigitalOcean IPv6 provisioning. Original description follows.

----

If a system has *existing* IPv6 connectivity, support some services over IPv6. (In other words, let's postpone work on *provisioning* IPv6 on servers; that can happen in parallel.)

* sslh: Adding another listen line opens up the gateway (and ssh).

* shadowsocks: Adding another listen address opens the service; obfs works too.

* gateway: Adding the IPv6 address to the self-signed gateway SSL cert works.

Note that systems with IPv6 ingress get IPv6 egress for connection-oriented services for free; check with https://ipv6.google.com/

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -52,7 +52,7 @@ of required questions without having to do any work besides copy & pasting!
 
 #### What Other Information to Include in Your Issue
 
-Please try your best to answer all of the questions in [the Streisand Issue template](https://github.com/StreisandEffect/streisand/blob/master/.github/ISSUE_TEMPLATE.md]. Answering these questions help us skip a lot of "Github telephone tag" that takes up time that could be used fixing the bug.
+Please try your best to answer all of the questions in [the Streisand Issue template](https://github.com/StreisandEffect/streisand/blob/master/.github/ISSUE_TEMPLATE.md). Answering these questions help us skip a lot of "Github telephone tag" that takes up time that could be used fixing the bug.
 
 When describing your issue focus on:
 
@@ -72,7 +72,7 @@ Issues are generally closed for one of three reasons:
 1. There wasn't enough information or nobody has been replying to questions
    asked.
 2. The issue was created on the wrong repository and should have been created as
-   a [streisand-discussions](https://github.com/StreisandEffect/streisand-discussions]
+   a [streisand-discussions](https://github.com/StreisandEffect/streisand-discussions)
    issue or vice-versa.
 3. After discussion and evaluation it was determined the problem was fixed or
    wasn't appropriate for Streisand
--- a/global_vars/default-site.yml
+++ b/global_vars/default-site.yml
@@ -11,6 +11,8 @@ streisand_ssh_private_key: "~/.ssh/id_rsa"
 
 vpn_clients: 5
 
+streisand_ipv6_enabled: yes
+
 streisand_openconnect_enabled: yes
 streisand_openvpn_enabled: yes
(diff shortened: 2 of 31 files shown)


clenga:
Interested in seeing this happen and will contribute in anyway i can. I run openvpn server through streisand and recently switched my phone carrier to tmobile which uses ipv6 for highspeed 4glte data connection. Ive briefly attempted manual config of ipv6 using my linode VPS but still working through the limited documentation/support for openvpn ipv6.

alimakki:
@clenga I have a branch with OpenVPN, WireGuard and OpenConnect running on ipv6 if you'd like to give it a try: https://github.com/alimakki/streisand/tree/wireguard_ipv6

clenga:
@alimakka great! I will give it a try tonight and report back to you with my results

clenga:
@alimakka , how far does your branch go into configuring ipv6? For example, i am using linode and with linode, when you request an ipv6 block , everything must be manually configured with static ip associations. Does your branch configure ipv6 by taking the  ipv6 address from /etc/network/interfaces or '

alimakki:
@clenga my branch is based off of this current branch - so if the server does have an IPv6 address it should be detected by the script if IPv6 is enabled. The relevant line can be found [here](https://github.com/StreisandEffect/streisand/blob/b65dc8027f83435fc17763cc923b95da1e9f180d/playbooks/roles/common/tasks/set-default-variables.yml#L43). I believe it does essentially populate the ipv6 address from the default interface.

I just ran it against Linode using my API key, and everything worked through and through. As per Linode's documentation, [all new servers are provisioned with IPv6 address by default](https://www.linode.com/docs/networking/an-overview-of-ipv6-on-linode/), which works in our favor.

clenga:
@alimakki First and foremost I appologize that it took so long to report back... Your IPV6 branch does indeed work well. Initially i had some issues getting the correct fork to install using termux but I believe it was mostly from my inexperience installing a specific fork rather than the git master... Anyways, after some trial an error and learning, I installed your fork, streisand prompted me for ipv6 and everything then fell into place seemlessly. I was able to successfully run both ipv4 and ipv6 configurations using my linode. I use a galaxy s9 running android 8.0 on tmobile network (tmobile uses ipv6 for LTE), configured my Linode VPS using termux for android and successfully connected using 2 different android openvpn clients, OpenVPN Connect and OpenVPN for Android by Arne Shwab. I felt that the OpenVPN client by Arne accepted the ipv6 config with more ease and with minimum user adjustments. I look forward to the future of streisand and all the developers like yourself @alimakki who dedicate the time, work, and knowledge into these projects. I will report back again, as soon as my linode is back up and running, if you have the need for more testing/logging/debugging.

alimakki:
@clenga Thank you for your valuable feedback, I'm glad that the fork ended up working out for you.  Also there's no need to apologize, we're all volunteers here :)

OpenVPN has always been somewhat of a point of contention - differences between clients on mobile and desktop machines (as well as free vs proprietary) makes it a bit tougher to get right. I've had issues with the iOS OpenVPN client switching back and from with IPv6 enabled cellular to IPv4-only WiIFi for example with the only solution being a manual disconnect/reconnect.

In any case should you encounter any further issues I've created a WIP PR #1471 for full IPv6 support , so feel free to submit comments or feedback should you have any.

Cheers!

Issue #1383, opened 2018-06-18
Tinyproxy not running after reboot


nopdotcom:
Reproducible #1381 

### Expected behavior:

`systemctl show tinyproxy` shows running tinyproxy processes

```
$ systemctl status tinyproxy
● tinyproxy.service - tinyproxy - a light-weight HTTP/HTTPS proxy daemon for POS
   Loaded: loaded (/etc/systemd/system/tinyproxy.service; enabled; vendor preset
   Active: active (running) since Mon 2018-06-18 00:33:13 UTC; 55s ago
     Docs: man:tinyproxy(8)
           https://www.banu.com/tinyproxy/
  Process: 17566 ExecStop=/usr/bin/killall -9 tinyproxy (code=exited, status=0/S
  Process: 17572 ExecStart=/usr/sbin/tinyproxy -c /etc/tinyproxy.conf (code=exit
 Main PID: 17577 (tinyproxy)
    Tasks: 11
   Memory: 3.2M
      CPU: 14ms
   CGroup: /system.slice/tinyproxy.service
           ├─17577 /usr/sbin/tinyproxy -c /etc/tinyproxy.conf
           ├─17578 /usr/sbin/tinyproxy -c /etc/tinyproxy.conf
           ├─17579 /usr/sbin/tinyproxy -c /etc/tinyproxy.conf
           ├─17580 /usr/sbin/tinyproxy -c /etc/tinyproxy.conf
           ├─17581 /usr/sbin/tinyproxy -c /etc/tinyproxy.conf
           ├─17582 /usr/sbin/tinyproxy -c /etc/tinyproxy.conf
           ├─17583 /usr/sbin/tinyproxy -c /etc/tinyproxy.conf
           ├─17584 /usr/sbin/tinyproxy -c /etc/tinyproxy.conf
           ├─17585 /usr/sbin/tinyproxy -c /etc/tinyproxy.conf
           ├─17586 /usr/sbin/tinyproxy -c /etc/tinyproxy.conf
           └─17587 /usr/sbin/tinyproxy -c /etc/tinyproxy.conf
```

### Actual Behavior:

After reboot:

```
# systemctl status tinyproxy
● tinyproxy.service - tinyproxy - a light-weight HTTP/HTTPS proxy daemon for POSIX operating systems
   Loaded: loaded (/etc/systemd/system/tinyproxy.service; enabled; vendor preset: enabled)
[...]

PR #1391, opened 2018-06-27
Ensure tinyproxy can write PID file

### LINKED ISSUE #1381: can not start tinyproxy error PID file. Failing.
tinyproxy error
how to start?

```
 root@streisand-sgr:~# systemctl start tinyproxy.service
Job for tinyproxy.service failed because a configured resource limit was exceeded. See "systemctl status tinyproxy.service" and "journalctl -xe" for details.
root@streisand-sgr:~# systemctl status tinyproxy.service
● tinyproxy.service - tinyproxy - a light-weight HTTP/HTTPS proxy daemon for POSIX operating systems
   Loaded: loaded (/etc/systemd/system/tinyproxy.service; disabled; vendor preset: enabled)
   Active: activating (auto-restart) (Result: resources) since Thu 2018-06-14 18:30:01 UTC; 2s ago
     Docs: man:tinyproxy(8)
           https://www.banu.com/tinyproxy/
  Process: 3465 ExecStart=/usr/sbin/tinyproxy -c /etc/tinyproxy.conf (code=exited, status=0/SUCCESS)

Jun 14 18:30:01 streisand-sgr systemd[1]: tinyproxy.service: Daemon never wrote its PID file. Failing.
Jun 14 18:30:01 streisand-sgr systemd[1]: Failed to start tinyproxy - a light-weight HTTP/HTTPS proxy daemon for POSIX operating syJun 14 18:30:01 streisand-sgr systemd[1]: tinyproxy.service: Unit entered failed state.
Jun 14 18:30:01 streisand-sgr systemd[1]: tinyproxy.service: Failed with result 'resources'.
```

alimakki:
- add a wrapper startup script to ensure pid directory is created prior to tinyproxy startup
- template out tinyproxy settings into a configuration file
- invoke startup script from the systemd unit file
- credit the original init.d script author for code re-use

Resolves #1381 #1383


z0rc:
You don't need a wrapper to handle temporary directory creation. The proper solution here is to use [systemd tmpfiles](https://www.freedesktop.org/software/systemd/man/tmpfiles.d.html).

alimakki:
@z0rc appreciate the heads up, wasn't aware of systemd tmpfiles. I'll give it a go and see what the results look like.

z0rc:
Actually, given the current use case, seems that [RuntimeDirectory](https://www.freedesktop.org/software/systemd/man/systemd.exec.html#RuntimeDirectory=) should be sufficient for this case.
```
Example: if a system service unit has the following,

RuntimeDirectory=foo/bar baz

the service manager creates /run/foo (if it does not exist), /run/foo/bar, and /run/baz. The directories /run/foo/bar and /run/baz except /run/foo are owned by the user and group specified in User= and Group=, and removed when the service is stopped.
```

playbooks/roles/tinyproxy/templates/tinyproxy.conf.j2
```diff
@@ -0,0 +1,142 @@
...
+# can be used for signalling purposes.
+#
+PidFile "{{ tinyproxy_pid_file }}"
+
+#
+# MaxClients: This is the absolute highest number of threads which will
+# be created. In other words, only MaxClients number of clients can be
+# connected at the same time.
+#
+MaxClients {{ vpn_clients }}
```
nopdotcom: Each client machine can chew up multiple threads due to browser parallelism, so this seems low. (Plus, we're starting 10 threads, so we should at least use them. :-) On the other hand, we don't want to blow out some poor t2.micro node. How about something like:

```
{{ [vpn_clients * 4,50]|min() }}
```

Or am I just overthinking this?

playbooks/roles/tinyproxy/templates/tinyproxy.conf.j2
```diff
@@ -0,0 +1,142 @@
...
+#
+# ConnectPort: This is a list of ports allowed by tinyproxy when the
+# CONNECT method is used.  To disable the CONNECT method altogether, set
+# the value to 0.  If no ConnectPort line is found, all ports are
+# allowed (which is not very secure.)
+#
+# The following two ports are used by SSL.
+#
+ConnectPort 443
+ConnectPort 563
```
nopdotcom: Up to you, but I don't see a problem in setting this to 0; it adds the functionality of an arbitrary CONNECT proxy to Streisand, which might be useful in a lot of cases.

playbooks/roles/tinyproxy/templates/tinyproxy.service.j2
```diff
@@ -6,8 +6,8 @@ Documentation=https://www.banu.com/tinyproxy/
 
 [Service]
 Type=forking
-PIDFile=/var/run/tinyproxy/tinyproxy.pid
-ExecStart=/usr/sbin/tinyproxy -c /etc/tinyproxy.conf
+PIDFile={{ tinyproxy_pid_file }}
+ExecStart={{ tinyproxy_startup_script }} {{ tinyproxy_conf_file }}
 ExecStop=/usr/bin/killall -9 tinyproxy
```
nopdotcom: I think we can get along without an `ExecStop` at all, now that I read [systemd.kill](https://www.freedesktop.org/software/systemd/man/systemd.kill.html). It might be worthwhile as an `ExecStopPost`.

playbooks/roles/tinyproxy/templates/tinyproxy.conf.j2
```diff
@@ -0,0 +1,142 @@
...
+#
+# Allow: Customization of authorization controls. If there are any
+# access control keywords then the default action is to DENY. Otherwise,
+# the default action is ALLOW.
+#
+# The order of the controls are important. All incoming connections are
+# tested against the controls based on order.
+#
+Allow 127.0.0.1
+Allow {{ streisand_ipv4_address }}
```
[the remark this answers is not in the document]
alimakki: You might be right, I'll have to test it out since that was the how it was previously configured.

playbooks/roles/tinyproxy/templates/tinyproxy.conf.j2
alimakki: The default previously was 100, so I'm unsure myself.

playbooks/roles/tinyproxy/templates/tinyproxy.conf.j2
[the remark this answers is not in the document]
alimakki:
```
- name: Reconfigure TinyProxy to accept connections from the Streisand server IP
  lineinfile:
    dest: /etc/tinyproxy.conf
    line: "Allow {{ streisand_ipv4_address }}"
    insertafter: "^Allow 127\\.0\\.0\\.1$"
    state: present
```

alimakki:
@z0rc the tmpfiles solution ended up working perfectly, thanks for the suggestion.

PR #1403, opened 2018-07-13
While asking for LE FQDN, tell the user what address it should point at

nopdotcom:
If you let Streisand create a server for you, it will not have a FQDN pointing at it. For new Ansible users, it isn't immediately obvious where to find the IP address of the new server. Be explicit.

If there is a better way to get the IP address at this stage, please fix.

### DIFF
--- a/playbooks/lets-encrypt.yml
+++ b/playbooks/lets-encrypt.yml
@@ -11,6 +11,9 @@
   vars_files:
     - roles/lets-encrypt/vars/main.yml
 
+  # FIXME: "ansible_play_hosts[0]" is a confusing way of writing
+  # "the address streisand-host points at".
+
   vars_prompt:
     - name: "streisand_domain_var"
       prompt: |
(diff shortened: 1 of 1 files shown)


playbooks/lets-encrypt.yml
```diff
@@ -26,6 +29,9 @@
         address. Make sure the fully qualified domain resolves to the correct IP
         address before proceeding.
 
+        If Streisand has created a new server for you, its IP address is
+        {{ ansible_play_hosts }}
```
jlund: Does this need to be `ansible_play_hosts[0]` (matching the comment on line 14)?

nopdotcom: For some reason I thought this was broken. It isn't. Thanks!

nopdotcom: Apparently it blows up CI if I do that.

Issue #1296, opened 2018-04-23
OpenVPN Ubuntu 16.04 instructions doesn't match UI


shotor:
### Expected behavior:

- Ubuntu OpenVPN extension to have "verify name exactly" + "additional TLS authentication mode"
- Initiate OpenVPN connection

### Actual Behavior:

- Missing fields in UI
- Connection fails

### Steps to Reproduce:
1. Install OpenVPN plugin `sudo apt-get install network-manager-openvpn-gnome`
2. Create new OpenVPN connection
3. Fill in fields
3. Click advanced
4. Go to TLS Authentication tab
5. Enter client name in "subject match"
6. Add key file to additional TLS authentication without setting mode
7. Save and connect => failure

![image](https://user-images.githubusercontent.com/6153419/39142291-82101dcc-472a-11e8-85b6-11c306077433.png)

Relevant section from manual:

Go to the TLS Authentication tab.
- Under Server Certificate Check choose verify name exactly and enter x-y-z as its value. (missing)
- Check Verify peer (server) certificate usage signature. (ok)
- Go to Additional TLS authentication or encryption. (ok)
- Select TLS-Crypt as the Mode. (missing)
- Select the ta.key file you downloaded from the client-files directory for the Key File. (ok)
- Click OK.

Frichetten:
![image](https://user-images.githubusercontent.com/10386884/39390677-8cc9acde-4a5d-11e8-8497-7a8ca03944c3.png)
Hey @shotor, thanks for submitting an issue! I'm running Ubuntu 17.10 (I'm excited to upgrade to 18.04 later :]) and I have posted a screenshot of my configuration. It looks like some changes were made to NetworkManager between 16.04 and 17.10. In particular, it looks like in 17.10 you need to set a 'Mode' but 16.04 does not need this. 

I'd like to help with this issue, but I'd likely need a comment from @cpu. @cpu, Any guidance for how different versions of Ubuntu should be documented? I know that OpenVPN has a dedicated section to Ubuntu, but not the particular versions? Would you want Streisand to support all the LTS versions of the OS?

cpu:
>  @cpu, Any guidance for how different versions of Ubuntu should be documented? I know that OpenVPN has a dedicated section to Ubuntu, but not the particular versions? Would you want Streisand to support all the LTS versions of the OS?

[...]

PR #1408, opened 2018-07-19
Updated Documentation for Ubtuntu 17.10, 18.04 and 16.04.

Frichetten:
There have been some issues with versions of Ubuntu. See Issue #1296 for information about problems with 16.04. 16.10 and 17.04 are now end of life (see https://wiki.ubuntu.com/Releases). I have updated the documentation (minor changes) to reflect 17.10 and 18.04 which are currently supported. Going forward it may be a good idea to update these entries for each Ubuntu release (there are usually some minor changes).

### DIFF
--- a/playbooks/roles/openvpn/templates/instructions.md.j2
+++ b/playbooks/roles/openvpn/templates/instructions.md.j2
@@ -13,6 +13,7 @@ For security reasons, this OpenVPN has been configured to only allow one client 
   * [Windows](#windows)
   * [macOS](#macos)
   * [Linux](#linux)
+  * [Linux (Ubuntu 16.04)](#linux-ubuntu-16.04)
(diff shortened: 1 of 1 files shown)


nopdotcom:
Thank you for the contribution! Good documentation is always a pain point. Ubuntu 16.04 is still supported by Canonical—are you interested in writing it up?

Frichetten:
Hey @nopdotcom, thanks for the feedback! I would like to writeup the documentation to connect with 16.04. However, issue #1296 shows that there are some problems with the default Ubuntu 16.04 network manager. I plan to do some research to see if I can get that working and I will submit a PR with the updated documentation at that time. 

Alternatively, if the NetworkManager can't be modified or fixed, we may have to ask the user to use OpenVPN from the command line after updating to 2.4.

PR #1409, opened 2018-07-22
WIP: Refactor OpenConnect VPN client instructions into user selectable client profiles

alimakki:
Streisand's current strategy of generating instructions for OpenConnect poses a number of usability issues when attempting to follow client specific instructions:

 - CLI/Android: In order to configure a device, a user would have to fetch the appropriate credentials from a list, and follow a generically generated snippet where per-client variables such as the pkcs12 certificate and password would have to be substituted manually making the process of setting up a device much more involved.
- VPN client size: The layout of the instructions page doesn't scale well when client numbers increase. Streisand's current limit of 20 clients would have multitudes of unordered lists of 20 elements throughout the page, increasing difficulty in navigation and visual pollution.

This PR attempts to address these issues by displaying per-client instructions where relevant, selectable via a drop down menu for each device that requires it.

The following platforms have been refactored:
 * macOS CLI
 * Linux CLI 
 * Android
 * iOS

In order to continue leveraging our current markdown templates and provide support for selectable instructions, the use of `<div>` elements were necessary; the original `markdown` only operates on `<span>` level elements and caused rendering issues, thus needed to be replaced with a [python implementation](https://github.com/Python-Markdown/markdown) of markdown.  While the final result does appear to be faithful, some quirks are apparent and warrants discussion.

Screenshots:

screen shot 2018-07-21 at 9 02 32 pm
---
screen shot 2018-07-21 at 9 03 39 pm

Comments and feedback appreciated!


nopdotcom:
I think the drop-down boxes boxes shouldn't start off blank; they should say something like "- Client profiles...".

alimakki:
That sounds reasonable, will change.

PR #1412, opened 2018-07-29
Don't print long apt keys to the log

nopdotcom:
This is mostly for the 200-line Tor key, but the other (25-line) keys were hidden for consistency.

### DIFF
--- a/playbooks/roles/lets-encrypt/tasks/install.yml
+++ b/playbooks/roles/lets-encrypt/tasks/install.yml
@@ -3,6 +3,7 @@
     id: 9862409EF124EC763B84972FF5AC9651EDB58DFA
     data: "{{ item }}"
   with_file: acmetool_ppa.pem
+  no_log: True
 
 - name: Add the official acmetool repository
   apt_repository:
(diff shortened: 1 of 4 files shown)


playbooks/roles/lets-encrypt/tasks/install.yml
```diff
@@ -1,4 +1,4 @@
-- name: Add the APT key for acmetool
+- name: "Add the APT key for acmetool; hiding 25 lines of log..."
   apt_key:
     id: 9862409EF124EC763B84972FF5AC9651EDB58DFA
     data: "{{ item }}"
```
jlund: Missing `no_log`.

playbooks/roles/nginx/tasks/main.yml
```diff
@@ -5,7 +5,7 @@
     state: absent
   with_items: "{{ apache_packages_to_remove }}"
 
-- name: Add the official Nginx APT key
+- name: "Add the official Nginx APT key; hiding 25 lines of log..."
   apt_key:
     id: 7BD9BF62
     data: "{{ item }}"
```
jlund: Missing `no_log`.

PR #1423, opened 2018-08-09
Add detect-public-ip.yml and integrate into set-default-variables.yml

speedy-beaver:
Challenge: Streisand VPN services not reachable due to external IP not visible on streisand-host for non-identified VM/host/cloud providers
Solution: Compare external and host IP address and update streisand_ipv4_address if neccessary

This will allow installation of streisand VPNs on a larger range of cloud instance providers. The challenge was discovered when installing streisand on a [Scaleway](httpw://www.scaleway.com/) 1-XS instances.


speedy-beaver:
Apologies, will close this one as well. I found more stuff that I would like to fix, will start to work off branches, will recreate this PR from a branch.

PR #1424, opened 2018-08-10
Add detection of public IP addresses for localhost and existing-server genesis

speedy-beaver:
Challenge: Streisand VPN services not reachable due to external IP not visible on streisand-host for non-identified VM/host/cloud providers
Solution: Compare external and host IP address and update streisand_ipv4_address if neccessary

This will allow installation of streisand VPNs on a larger range of cloud providers. The challenge was discovered when installing streisand on a [Scaleway](https://www.scaleway.com/) 1-XS instances.

### DIFF
--- a/playbooks/roles/common/tasks/detect-public-ip.yml
+++ b/playbooks/roles/common/tasks/detect-public-ip.yml
@@ -12,7 +12,7 @@
     external_ipv4_address: "presumed_failed"
 
 - name: "Check external IP Address through Google"
-  command: dig -4 o-o.myaddr.l.google.com @ns1.google.com TXT +short
+  command: dig +short myip.opendns.com @resolver1.opendns.com A
   register: dig_output
 
 - name: "Set the variable to the value"
--- a/playbooks/roles/common/tasks/detect-public-ip.yml
+++ b/playbooks/roles/common/tasks/detect-public-ip.yml
@@ -1,0 +1,51 @@
+---
+# detect-public-ip.yml will attempt to identify whether the server's public
+# IP address is different from what is visible on the host and, if
+# successfully detected, ask to update the address for documentation and
+# configuration profiles
+- name: "Install dns module"
+  apt:
+    name: dnsutils
+
+- name: "Initialize lookup variable"
+  set_fact:
+    external_ipv4_address: "presumed_failed"
+
+- name: "Check external IP Address through Google"
+  command: dig o-o.myaddr.l.google.com @ns1.google.com TXT +short
+  register: dig_output
+
+- name: "Set the variable to the value"
+  set_fact:
+    external_ipv4_address: "{{ dig_output.stdout | regex_replace('\"', '') }}"
+    when: (dig_output.rc == 0)
+
+# Enter this block only when when the IPs are different and query user for updating
+# to public ip
+- block:
+    - name: "Initialize the prompt"
+      set_fact:
+        prompt_external_ip: |
+          We have found another public IP address of your server
+
+          Some cloud providers use load balancers or SDN to make servers externally
+          reachable. It seems
+          - {{ external_ipv4_address }} is publicly visible
+          - {{ streisand_ipv4_address }} is visible on the server
+
+          Type 'yes' to use {{ external_ipv4_address }} for the VPN
+          Hit 'enter' to skip and use {{ streisand_ipv4_address }}.
+
+          Skip with 'enter' if you do not know {{ external_ipv4_address }}
+
+    - name: "Ask user to update to public IP address"
+      pause:
+        prompt: "{{ prompt_external_ip }}"
+      register: publish_external
+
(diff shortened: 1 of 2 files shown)


speedy-beaver:
CI build failing seems to be due to external factors. Is it possible to trigger another test run?

nopdotcom:
Sorry about that. If it happens again, comment again to get our attention.

playbooks/roles/common/tasks/detect-public-ip.yml
```diff
@@ -0,0 +1,51 @@
...
+- name: "Install dns module"
+  apt:
+    name: dnsutils
+
+- name: "Initialize lookup variable"
+  set_fact:
+    external_ipv4_address: "presumed_failed"
+
+- name: "Check external IP Address through Google"
+  command: dig o-o.myaddr.l.google.com @ns1.google.com TXT +short
```
nopdotcom: Is this behavior documented?

nopdotcom: This also returns an IPv6 address when available.

playbooks/roles/common/tasks/detect-public-ip.yml
```diff
@@ -0,0 +1,51 @@
+---
+# detect-public-ip.yml will attempt to identify whether the server's public
+# IP address is different from what is visible on the host and, if
+# successfully detected, ask to update the address for documentation and
+# configuration profiles
+- name: "Install dns module"
```
nopdotcom: Ansible has [the `ipify_facts` module](https://docs.ansible.com/ansible/latest/modules/ipify_facts_module.html), but it has a flaw: it returns IPv6 addresses when available.

Consider using https://v4.ident.me/.json via [the `uri` module](https://docs.ansible.com/ansible/latest/modules/uri_module.html).

nopdotcom: (BTW, this was just the first service I found; I have no special preference for it.)

speedy-beaver:
Thank you for the quick feedback.

## IPv6 address in DNS response
Thank you for noticing. The dig method with the Google servers can be limited to IPv4 by telling dig explicitly to use IPv4 by calling it with ```dig -4```. The pull request has been updated to include this option. When adding IPv6 to streisand, the explicit ```-6``` option will be helpful.

## Google and alternatives, documentation
[Here is a list][1] of lookup-your-own-IP-over-http providers from 2016, over 40% of them fail today. Challenging to pick a stable one based on the names - apart from Akamai maybe.

The best "documentation" for the currently used method available is a person, claiming to work for Google Public DNS, mentioning this method in the [Google Public DNS support forum][2].

Regarding privacy, Google have a comprehensive [privacy statement][3] regarding the collected data.

Given:
a) the limited data provided - essentially we are asking to echo our public IP address back
b) Google promising not to correlate/combine personal information from other services
c) Google running this on top one of the largest distributed global infrastructures
I would propose to stay with this form of resolution and provider until Google stops that service. I am open for feedback.

The CI seems to have problems to reach some download hosts, would be great if you can trigger it again.

[1]: https://unix.stackexchange.com/questions/254328/get-the-external-ip-address-in-shell-without-dig-in-2016
[2]: https://groups.google.com/forum/#!topic/public-dns-discuss/uyzmMcHQBE0 
[3]: https://developers.google.com/speed/public-dns/privacy

nopdotcom:
Sometimes when I resolve this, I get something else.

```
$ dig -4 o-o.myaddr.l.google.com @ns1.google.com TXT +short
"74.12x.xx.xx"
"edns0-client-subnet 192.0.2.72/32"
$
```

Apparently some middlebox is at 74.12x.xx.xx; I'm at "192.0.2.72". There were some reports that the `edns0-client-subnet` field may not be authoritative either. :-(

I think you're right that Akamai is the most likely to be live from that list, but ident.me allows explicit IPv4 queries. How about this—whether you pick https://v4.ident.me/.json or something else that has an explicit IPv4 feature, I'll put a smoke test in CI, and when that service starts blowing up, we'll know it?

nopdotcom:
(I realize all of these alternatives suck. I'm sorry for being picky about this one thing. `dig +short myip.opendns.com @resolver1.opendns.com A` may be everything we're looking for except for the privacy policy, and personally I'm not troubled by that, but I'm not the sole decider.)

speedy-beaver:
No worries, I am very happy that you take this serious and that you are helping. Your additional comments and tests will make the final solution more robust. 

Regarding the privacy policy, my argument is still that we are asking to echo our own IP back, so very little is exposed - I admit there might be aspects even with that which I have not grasped.

I like the DNS based solution, it feels more elegant and slimmer. I am happy to update it to the opendns one for now. 

If you feel one of the http-based lookups is a better fit, I can update that without too much additional efforts (famous last words :) as I have prototyped that solution as well.

speedy-beaver:
Great, thank you for the feedback and help!

Issue #847, opened 2017-07-30
AnyConnect Android app asks for username/password


sammoth:
I have followed the Android instructions on the OpenConnect page on Streisand. After step 21 (Tap Connect on the group selection screen), the app prompts me for a username/password and won't connect if I enter nothing. Have I done something wrong?

Thanks

cpu:
@alimakki @jlund Have either of you used AnyConnect on Android (or otherwise) recently?

alimakki:
@cpu Recently when I tested the revert to 0.11.7, seemed to work fine on my end.
Looking back setting up AnyConnect is a somewhat involved processes and its easy to overlook a step which I have done several times.

OP may also need to spin up a new instance in case the version of ocserv installed is 0.11.8

sammoth:
I'm using a fresh install and it's definitely the right version:
```# ocserv --version
ocserv 0.11.7```
I reset the app and ran through the steps again carefully and got the same, so I definitely think it is not working as expected.

This isn't important to me as I have OpenVPN running now instead.

alimakki:
@sammoth seems odd. In any case I'll give it another go tomorrow.

HougeLangley:
Same issue happen in my phone. I used linux client configure user name and password connect success.

alimakki:
Very odd. I was testing today and it appears to be working on my end.

jrodrigosm:
Hi, I just installed AnyConnect on my Android phone today, and I'm having the same issue when starting the AnyConnect VPN: I'm being prompted for user/password.

The streisand server was installed last week, so it's quite recent.

I don't really know what to do, but if you give me some pointers I'm willing to help. I can even (privately) share my server credentials (I will spin up a new one afterwards).

I'm running version 4.0.09030 of AnyConnect, on Android 7.1.1.

alimakki:
@jrodrigosm can you let me know the result of `git rev-parse HEAD` on your streisand directory?

nopdotcom:
I feel your pain.

![smartselectimage_2017-08-07-16-34-41 copy 4](https://user-images.githubusercontent.com/1357866/29045337-3edaa032-7b91-11e7-9437-9b851f152b2d.png)

(This is not exactly a stock Samsung 6.0 ROM, so I’ll try this in the simulator later.)

This is a deployment against 9cc77832 in nopdotcom/streisand, which is a null diff to jlund/streisand 24841095.

jrodrigosm:
[...]

PR #1440, opened 2018-09-04
Add instructions for OpenConnect app on Android.

### LINKED ISSUE #1143: Android instructions for OpenConnect instead of AnyConnect
Currently the instructions for Android include one for Cisco AnyConnect. Why didn't you choose OpenConnect instead which is also available in Google Play?

jsha:
The OpenConnect app has a few advantages over the Cisco AnyConnect:
 - Doesn't require agreeing to a EULA.
 - GPL'ed.
 - Can install a CA certificate rather than clicking through a
   certificate mismatch prompt.
 - Forward secrecy.
 - It doesn't appear to tickle the bug in #847 (prompting for a
   username).

For now, I kept the AnyConnect instructions, but I think if a number of
people are successful with these instructions, it would make sense to
provide instructions only for OpenConnect.

Fixes #1143.

### DIFF
--- a/playbooks/roles/openconnect/templates/instructions.md.j2
+++ b/playbooks/roles/openconnect/templates/instructions.md.j2
@@ -117,6 +117,34 @@ Client certificates are a mechanism by which clients can authenticate themselves
 <a name="android"></a>
 ### Android ###
 
+1. Download the [server certificate](/openconnect/ca.crt) file.
+1. You will be prompted to enter your phone's PIN. Enter that.
+1. You will be prompted to *Name the certificate*. Tap the line under *Certificate name* and enter `{{ streisand_server_name }}`.
+1. Tap *OK*.
+1. Download a [client certificate file](#clientcerts) from the list above.
+1. Download [OpenConnect](https://play.google.com/store/apps/details?id=app.openconnect) from Google Play.
+1. Launch the application.
+1. Tap the *+* icon to add a new VPN.
+1. Enter `{{ streisand_ipv4_address }}:{{ ocserv_port }}`.
+1. Tap *User certificate*.
+1. Find the client certificate file you downloaded and tap it. Most likely it is in the Download folder if you downloaded it using the browser on your phone.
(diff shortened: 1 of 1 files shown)


playbooks/roles/openconnect/templates/instructions.md.j2
```diff
@@ -117,6 +117,33 @@ Client certificates are a mechanism by which clients can authenticate themselves
 <a name="android"></a>
 ### Android ###
```
alimakki:
```
<a name="android-openconnect"></a>
### Android OpenConnect ###
```

playbooks/roles/openconnect/templates/instructions.md.j2
```diff
@@ -117,6 +117,33 @@ Client certificates are a mechanism by which clients can authenticate themselves
...
+1. Tap that entry.
+1. First time only:
+  1. Accept the Connection Request dialog that Android displays.
+  1. You will be prompted *Enter PKCS#12 pass phrase:*. Enter the password for the client certificate that you downloaded. Note: You may receive an error during this phase. If so, try again a couple of times.
+  1. You will be prompted *Certificate warning*. Tap *Always connect.*
+1. Each time you connect:
+  1. You will be prompted *Please select your group.* The correct default has already been chosen. Tap *OK*.
+1. You should be good to go! You can verify that your traffic is being routed properly by [looking up your IP address on DuckDuckGo]({{ streisand_my_ip_url }}). It should say *Your public IP address is {{ streisand_ipv4_address }}*.
+
+Alternate instructions using Cisco AnyConnect, in case the above instructions fail:
```
alimakki:
```
<a href="android-anyconnect"></a>
### Android AnyConnect ###
Alternate instructions using ....
```

baimafeima:
Thank you @jsha for the PR. I think it would still be good to keep both OpenConnect and AnyConnect so that the widest possible user group can meaningfully follow the instructions.

PR #1452, opened 2018-09-28
Note Incompatible API keys from Linode Cloud Manager

patthiel:
* The Linode Ansible module uses the now deprecated Linode APIv3. APIv3 keys can still be generated from the page specified: https://manager.linode.com/profile/api , however, users of the beta Linode Manager (https://cloud.linode.com) may be unaware of the incompatibility of APIv4 tokens and Streisand. (https://github.com/StreisandEffect/streisand/issues/1449).  This PR adds a note to help avoid confusion.

### DIFF
--- a/playbooks/linode.yml
+++ b/playbooks/linode.yml
@@ -40,7 +40,7 @@
       private: no
 
     - name: "linode_api_key"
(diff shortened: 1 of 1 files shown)


alimakki:
Thanks for clarifying @patthiel 👍

PR #1455, opened 2018-10-01
Clear `ANSIBLE_LIBRARY` plus minor doc change

nopdotcom:
People may have stuff installed in `$ANSIBLE_LIBRARY` (by default, `$HOME/.ansible/plugins/modules` and `/usr/share/ansible/plugins/modules`. (Yes, `/usr/share` even on macOS!) To reduce variability in the install environment, set `ANSIBLE_LIBRARY` to the empty string.

In the documentation, mention the warnings about `pbr`, which we should be looking at, eventually...


README.md
```diff
@@ -130,7 +135,8 @@ Complete all of these tasks on your local home machine.
 2. Run the installer for Ansible and its dependencies.
 
         ./util/venv-dependencies.sh ./venv
-  * The installer will detect missing packages, and print the commands needed to install them.
+  * The installer will detect missing system packages, and print the commands needed to install them.
+  * Expect warnings about `pbr` and `azure-cli-core`.
```
cpu: Can you clarify that the warnings are safe to ignore? I think telling the user to expect them is a good idea but there should also be clarity about the impact the warnings will have (hopefully none!)

nopdotcom: I need to look long and hard at what the warnings actually are. It's looking like we can't have a venv which is 

1. identical for all the providers, and
2. without significant mutually-contradictory requirements

Somehow things seem to work, even with broken requirements, but I may have to change the venv installer to request "which provider do you plan to use?"

nopdotcom: OK, I've moved the "ignore these warnings" into `Installation.md`. That should be enough for this PR.

PR #1472, opened 2018-10-21
18.10 openvpn doc update

Frichetten:
The Streisand OpenVPN configuration uses network-manager-openvpn-gnome as a nice plugin. Due to some minor variances between the different Ubuntu versions we have slight instruction changes between them. I have tested Ubuntu 18.10 and confirmed that it is the same as 18.04. This Pull Request is just updating the documentation to reflect that.


alimakki:
Thank you 👍

Issue #1434, opened 2018-08-28
Issue connecting to Wireguard Linux ubuntu 18.04


ilyas95:
When I try to connect to the server by using sudo wg-quick up configfile I get the following as output:
[#] ip link add configfile type wireguard
[#] wg setconf configfile /dev/fd/63
[#] ip address add 10.192.122.2/32 dev configfile
[#] ip link set mtu 1420 dev configfile
[#] ip link set configfile up
[#] resolvconf -a configfile -m 0 -x
/usr/bin/wg-quick: line 31: resolvconf: command not found
[#] ip link delete dev configfile

I tried to look inside the file /usr/bin/wg-quick but I cannot find the problem
Someone had this problem?
I'm sure the problem is not the server because I'm able to connect with my android phone

nopdotcom:
Which version of which Linux?

ilyas95:
Ubuntu 18.04 LTS (Bionic Beaver)

ilyas95:
@Frichetten 
It worked, thank you
I installed openresolv as you suggested

Frichetten:
I wouldn't recommend it. (I had problems while testing with 'resolvconf'). Do me a favor, could you run `cat /etc/resolve.conf` and see what the outout is? (Don't post it online because it should contain your servers IP address).

If it contains anything other than the IP of your server, we need to fix that. If it does have the right IP the next step would be to run the same command on the server and see what that output is. 

As a side note, what VPS provider are you using?

z0rc:
127.0.0.53 is local systemd-resolved daemon, that installed by default in Ubuntu. I'm not sure that `resolvconf` supports it or ever will. And in general I believe it's `wg-quick` missing support systemd-resolved and shoudn't entirely rely on `resolvconf`.

z0rc:
As alternative, you don't need a wg-quick and can enable Wireguard in systemd directly. See https://wiki.debian.org/Wireguard#Step_2_-_Alternative_C_-_systemd, please be aware, that this isn't a full guide and it might not work in your case.

Frichetten:
@ilyas95, after you run the wg-quick command (with your configuration) what does /etc/resolv.conf look like? When I have Wireguard turned off my /etc/resolv.conf looks exactly like you mentioned. But when I turn Wireguard on it shows the internal 10.x.x.x IP address of the server. 

If you are leaking DNS, we'd need to ensure your system is properly routing DNS queries to the server.

[...]

PR #1473, opened 2018-10-22
Updated documentation to fix issue found in Issue #1434

Frichetten:
This PR was made in response to Issue #1434. Ubuntu/Debian users will need to use the `openresolv` package in order to run Wireguard as noted [here](https://lists.zx2c4.com/pipermail/wireguard/2017-May/001351.html) and [here](https://mullvad.net/en/guides/wireguard-and-mullvad-vpn/).

The fix is simple for the user, just install `openresolv` and Wireguard will be able to use it.

### DIFF
--- a/playbooks/roles/wireguard/templates/instructions.md.j2
+++ b/playbooks/roles/wireguard/templates/instructions.md.j2
@@ -42,6 +42,10 @@ An [experimental configuration for OpenWrt/LEDE](#openwrt) 17.01.4 (or later) is
 
    `sudo sh -c 'umask 077; mkdir -p /etc/wireguard; cat > /etc/wireguard/{{ vpn_client_names.results[0].stdout }}.conf' < ~/Downloads/{{ vpn_client_names.results[0].stdout }}.conf`
 
+1. (Ubuntu/Debian) For Ubuntu and Debian users you will need to install the `openresolv` package:
+
+   `sudo apt-get install openresolv`
+
 1. Use the `wg-quick` utility to bring up the WireGuard interface:
 
    `sudo wg-quick up {{ vpn_client_names.results[0].stdout }}`


alimakki (approved):

z0rc:
https://github.com/systemd/systemd/blob/master/NEWS#L176-L179

Systemd 239 is packaged in Ubuntu 18.10. I didn't test this, but seems, that all you have to do, is to symlink `resolvectl` to `resolvconf` to enable compatability interface, if this symlink isn't created by package itself.

z0rc:
As for older systemd versions, I'd recommend to drop `wg-quick` in favor to using native systemd wireguard configuration, as shown at https://wiki.archlinux.org/index.php/WireGuard#Example_configuration_for_systemd-networkd.

Frichetten:
@z0rc, Thank you for you insight on this, and thank you for providing sources with some additional context. I have not personally run into any issues with this setup (on my personal machine for several months), but it is interesting to hear that others have. 

I would be very interested in researching this further and submitting a pull request relating to using systemd rather than wg-quick.

alimakki:
@Frichetten, could you close and re-submit this PR? It looks like it's stuck somewhere in CI purgatory 

Thanks 👍

bitinerant:
And a link here to the new PR would be helpful. :smiley:

sarnold:
Yes please; also I have some related comments for here and issue #1434 

As far as dns leakage, it seems most likely related to the version of systemd on bionic.  Also, things should still work on bionic using another method than systemd-resolvd, even with the "latest" version on groovy which is supposed to route correctly (ie, with the associated vpn/tunnel traffic).

I've been wrestling with these issues a lot lately, on multiple distros/releases, in order to support yet another vpn tool with default dns privacy, etc. First, there should be nothing inherently wrong in using (and even depending on) openresolv/resolvconf or systemd-resolvd, just bear in mind not everyone will use the same thing.  The main issue with the first two is most likely the package fails to (re)set the ``/etc/resolv.conf`` symlink.  Resetting it to a static file turns systemd-resolvd into a client that just reads the contents of ``/etc/resolv.conf``.

Also, according to the getdns/dns privacy people systemd-resolvd also has implementation issues with how its version of dns-over-tls works; given that and other issues I would recommend working out an easy solution using the static file method with openresolv.

UPDATE:  installing openresolv on updated bionic definitely does *not* touch the ``/etc/resolv.conf`` symlink, so *if* you want to use a different tool than systemd-resolvd, then you *must* manually delete the symlink and touch a new file and edit it.  Eg, if you want to use the ``stubby`` resolver or dnsmasq, you'll need to set ``nameserver 127.0.01`` or whatever your config needs.

sarnold:
Aaand now that openresolv is installed I can see there is only one binary named ``resolvconf`` but it gets installed under sbin and would appear "missing" from a normal user env.

PR #1495, opened 2018-12-13
Correctly detect public IPv4 when system has v6

quite:
Without -4, the lookup might be done over v6, and but we are indeed requesting `A`.

(To get our public IPv6 the same way, we'd do `dig -6 … AAAA`)

The reason for this is that dig returned an empty string, and the role asked if I wanted to use that as public address...


alimakki:
Thank you @quite

PR #1497, opened 2018-12-20
 Fix typo: certifcate -> certificate

FreedomBen:
Also yum -> dnf for Fedora instructions (yum is gone in Fedora for a while now)


alimakki:
Thank you @FreedomBen

PR #1520, opened 2019-02-23
Split README.md into smaller, easier-to-read pieces

nopdotcom:
The existing `README.md` is intimidating, and doesn't immediately present the information that the reader wants first.

I couldn't resist rewriting a little, but the other sections are largely the same, and should be relatively easy to re-translate.


Installation.md
```diff
@@ -0,0 +1,66 @@
...
+
+### Execution ###
+1. Clone the Streisand repository and enter the directory.
+
+        git clone https://github.com/StreisandEffect/streisand.git && cd streisand
+
+2. Run the installer for Ansible and its dependencies.
+
+        ./util/venv-dependencies.sh ./venv
+  * The installer will detect missing packages, and print the commands needed to install them.
```
cpu: I think this line being rendered as a bullet point is a little bit strange looking. Can you fold it into the numbered point sentence, e.g. something like:

> 2. Run the installer for Ansible and its dependencies. If the installer detects missing packages it will print the commands needed for you to install them.

README.md
```diff
@@ -16,180 +18,72 @@ Streisand
...
-		sudo yum -y update && sudo yum install -y git gcc python-devel python-crypto python-pycurl python-pip libcurl-devel
-  * On macOS, `git` is part of the Developer Tools, and it will be installed the first time you run it. If there isn't already a `pip` command installed, install it with:
+There's a [list of supported cloud providers](#cloud-providers); experts may be able to use Streisand to install on many other cloud providers.
 
-        sudo python2.7 -m ensurepip
+## VPN services
 
-### Execution ###
-1. Clone the Streisand repository and enter the directory.
+People usually refer to tools to avoid network censorship as Virtual Private Networks (VPNs).
```
cpu:
```suggestion
One tool people use to avoid network censorship are Virtual Private Networks (VPNs). 
```

README.md
```diff
@@ -16,180 +18,72 @@ Streisand
...
-noninteractive deployment, in which all of the required information is passed
-on the command line or in a configuration file.
+#### Other providers
+If you are an expert and can set up a *fresh Ubuntu 16.04 server* elsewhere, Streisand includes code for:
 
-Example configuration files are found under `global_vars/noninteractive`. Copy
-and edit the desired parameters, such as providing API tokens and other choices,
-and then run the appropriate script.
+* Using `ssh` to connect to your remote server from your local machine
+* After connecting to your remote server, you can run Streisand *in place*, called "localhost deployment". This converts the server to a Streisand node.
```
cpu: I think this bullet and the prev could be condensed into one like:

* Using `ssh` to connect to a remote server to run the Streisand builder *in place*, converting the remote server into a Streisand VPN server. We call this "localhost deployment".

I would avoid the term "Streisand node" since I don't think its used elsewhere in the README to this point.

README.md
```diff
@@ -16,180 +18,72 @@ Streisand
...
 
-To deploy a new Streisand server:
+"Localhost deployment" often works better than `ssh` deployment.
```
cpu: I think this comment is missing some context (otherwise why isn't it the default method the README recommends to everyone?).

Adding the context will probably make this too lengthy so I'd recommend just striking this sentence and letting advanced users that need it find out about it via the other pages of docs you link out to.

README.md
```diff
@@ -16,180 +18,72 @@ Streisand
...
-        --provider digitalocean \
-        --site-config global_vars/noninteractive/digitalocean-site.yml
+For more information, see [the advanced installation instructions](Advanced%20installation.md).
 
-To run the Streisand provisioning on the local machine:
+## Installation
 
-      deploy/streisand-local.sh \
-        --site-config global_vars/noninteractive/local-site.yml
+You need command-line access to a Unix system. You can use Linux, BSD, or macOS; on Windows 10, the Windows Subsystem for Linux (WSL) counts as Linux.
```
cpu: Earlier in the README Windows is called out as unsupported and WSL is said to maybe work. This part of the README makes it sound like it will probably work and is supported.

README.md
```diff
@@ -16,180 +18,72 @@ Streisand
...
 
-To run the Streisand provisioning against an existing server:
+Once you're ready, see the [full installation instructions](Installation.md).
 
-      deploy/streisand-existing-cloud-server.sh \
-        --ip-address 10.10.10.10 \
-        --ssh-user root \
-        --site-config global_vars/noninteractive/digitalocean-site.yml
 
 Upcoming Features
```
cpu: I would vote to remove this whole section. At best its optimistic and at worst its inaccurate :laughing: Either way it probably doesn't belong in the top level README as opposed to somewhere else.

README.md
```diff
@@ -16,180 +18,72 @@ Streisand
...
 -----------------
 * Easier setup.
+* Faster adoption of new censorship-avoidance tools
+
+We're looking for help with both.
 
-If there is something that you think Streisand should do, or if you find a bug in its documentation or execution, please file a report on the [Issue Tracker](https://github.com/StreisandEffect/streisand/issues).
+If there is something that you think Streisand should do, or if you find a bug in its documentation or execution, please file a report on the [Issue Tracker](https://github.com/StreisandEffect/streisand/issues). We appreciate
```
cpu: typo: This ends abruptly with just "We appreciate"

Services.md
```diff
@@ -0,0 +1,39 @@
+# Services
+
+Services Provided
+-----------------
+
+* [OpenSSH](https://www.openssh.com/)
+  * Windows and Android SSH tunnels are also supported, and a copy of the keypair is exported in the .ppk format that PuTTY requires.
```
cpu:
```suggestion
  * Windows and Android SSH tunnels are also supported, and a copy of the keypair is exported in the `.ppk` format that PuTTY requires.
```

Services.md
```diff
@@ -0,0 +1,39 @@
...
+  * [Tinyproxy](https://banu.com/tinyproxy/) is installed and bound to localhost. It can be accessed over an SSH tunnel by programs that do not natively support SOCKS and that require an HTTP proxy, such as Twitter for Android.
```
cpu:
```suggestion
  * [Tinyproxy](https://banu.com/tinyproxy/) can be installed and bound to localhost. It can be accessed over an SSH tunnel by programs that do not natively support SOCKS and that require an HTTP proxy, such as Twitter for Android.
```

Services.md
```diff
@@ -0,0 +1,39 @@
...
+  * An unprivileged forwarding user and SSH keypair are generated for [sshuttle](https://github.com/sshuttle/sshuttle) and SOCKS capabilities.
```
cpu:
```suggestion
  * An unprivileged forwarding user and SSH keypair can be generated for [sshuttle](https://github.com/sshuttle/sshuttle) and SOCKS capabilities.
```

Services.md
```diff
@@ -0,0 +1,39 @@
...
+* [OpenConnect](https://ocserv.gitlab.io/www/index.html) / [Cisco AnyConnect](https://www.cisco.com/c/en/us/products/security/anyconnect-secure-mobility-client/index.html)
+  * OpenConnect (ocserv) is an extremely high-performance and lightweight VPN server that also features full compatibility with the official Cisco AnyConnect clients.
+  * The [protocol](https://ocserv.gitlab.io/www/technical.html) is built on top of standards like HTTP, TLS, and DTLS, and it's one of the most popular and widely used VPN technologies among large multi-national corporations.
+    * This means that in addition to its ease-of-use and speed, OpenConnect is also highly resistant to censorship and is almost never blocked.
+* [OpenVPN](https://openvpn.net/index.php/open-source.html)
+  * Self-contained "unified" .ovpn profiles are generated for easy client configuration using only a single file.
```
cpu:
```suggestion
  * When enabled, self-contained "unified" .ovpn profiles are generated for easy client configuration using only a single file.
```

Services.md
```diff
@@ -0,0 +1,39 @@
...
+  * Both TCP and UDP connections are supported.
+  * Client DNS resolution is handled via [Dnsmasq](http://www.thekelleys.org.uk/dnsmasq/doc.html) to prevent DNS leaks.
+  * TLS Authentication is enabled which helps protect against active probing attacks. Traffic that does not have the proper HMAC is simply dropped.
+* [Shadowsocks](https://shadowsocks.org/en/index.html)
+  * The high-performance [libev variant](https://github.com/shadowsocks/shadowsocks-libev) is installed. This version is capable of handling thousands of simultaneous connections.
```
cpu:
```suggestion
  * When enabled, the high-performance [libev variant](https://github.com/shadowsocks/shadowsocks-libev) is installed. This version is capable of handling thousands of simultaneous connections.
```

Services.md
```diff
@@ -0,0 +1,39 @@
...
+  * [AEAD](https://shadowsocks.org/en/spec/AEAD-Ciphers.html) support is enabled using ChaCha20 and Poly1305 for enhanced security and improved GFW evasion.
+  * The [simple-obfs](https://github.com/shadowsocks/simple-obfs) plugin is installed to provide robust traffic evasion on hostile networks (especially those implementing quality of service (QOS) throttling).
+* [sslh](https://www.rutschle.net/tech/sslh/README.html)
+  * Sslh is a protocol demultiplexer that allows Nginx, OpenSSH, and OpenVPN to share port 443. This provides an alternative connection option and means that you can still route traffic via OpenSSH and OpenVPN even if you are on a restrictive network that blocks all access to non-HTTP ports.
+* [Stunnel](https://www.stunnel.org/index.html)
+  * Listens for and wraps OpenVPN connections. This makes them look like standard SSL traffic and allows OpenVPN clients to successfully establish tunnels even in the presence of Deep Packet Inspection.
```
cpu:
```suggestion
  * When enabled, listens for and wraps OpenVPN connections. This makes them look like standard SSL traffic and allows OpenVPN clients to successfully establish tunnels even in the presence of Deep Packet Inspection.
```

Services.md
```diff
@@ -0,0 +1,39 @@
...
+  * Unified profiles for stunnel-wrapped OpenVPN connections are generated alongside the direct connection profiles. Detailed instructions are also generated.
+  * The stunnel certificate and key are exported in PKCS #12 format so they are compatible with other SSL tunneling applications. Notably, this enables [OpenVPN for Android](https://play.google.com/store/apps/details?id=de.blinkt.openvpn) to tunnel its traffic through [SSLDroid](https://play.google.com/store/apps/details?id=hu.blint.ssldroid). OpenVPN in China on a mobile device? Yes!
+* [Tor](https://www.torproject.org/)
+  * A [bridge relay](https://www.torproject.org/docs/bridges) is set up with a random nickname.
```
cpu:
```suggestion
  * If chosen by the user a [bridge relay](https://www.torproject.org/docs/bridges) is set up with a random nickname.
```

README.md
nopdotcom: This is split up because of "existing server" vs "localhost" deployment. I've rewritten.

nopdotcom: OK, I took your advice and just punted everybody to the advanced instructions.

Installation.md
nopdotcom: Yeah, that works.

README.md
nopdotcom: I think I resolved the opinions on this.

README.md
nopdotcom: Fixed.

README.md
nopdotcom: I changed to "Things we want to do better"; I'd like to present ourselves as imperfect and looking for help.

PR #1528, opened 2019-03-17
Update README-ru.md

DmytroLitvinov:
Fix error:
https://en.wikipedia.org/wiki/Parenthesis_(rhetoric)


alimakki:
Thank you @DmytroLitvinov

PR #1534, opened 2019-03-21
Small documentation fix caused by block quotes

fmakdemir:
There was a small error causing styles in last section to be messed up a little bit


alimakki:
Thank you @fmakdemir

PR #1539, opened 2019-03-25
Use correct name for private SSH key

jmillxyz:
When starting up a new server, I got the following error (formatted for readability):

```
fatal: [localhost]: FAILED! => 
{
  "msg": 
  "The task includes an option with an undefined variable. The error was: 'streisand_ssh_key' is undefined

  The error appears to have been in '/<path_redacted>/streisand/playbooks/roles/genesis-rackspace/tasks/main.yml': line 10, column 3, but may
  be elsewhere in the file depending on the exact syntax problem.

  The offending line appears to be:

  - name: Create the server  
  ^ here"
}
```

I resolved this error by renaming `streisand_ssh_key` to `streisand_ssh_private_key`, in order to match the value in `~/.streisand/site.yml`


CorbanR:
@jmillxyz thanks!

PR #1548, opened 2019-04-02
Add a hack to make AWS Roles usable

MikaelSmith:
Enable using a manually-generated session token for AWS Roles.

A better solution would be to use https://docs.ansible.com/ansible/latest/modules/sts_assume_role_module.html, but I'm not sure how to add the conditional logic required to add that to the Streisand setup workflow.


MikaelSmith:
Actually, I may be able to use AWS_SECURITY_TOKEN. Not very discoverable though.

playbooks/amazon.yml
```diff
@@ -98,6 +98,16 @@
       prompt: "\nWhat is your AWS Secret Access Key?\n"
       private: no
 
+    - name: "aws_session_token"
+      prompt: |
+        If you use AWS Roles, see https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-role.html
+        then get a session token by running 'aws sts get-caller-identity --profile <profile>' and finding
+        it in ~/.aws/cli/cache/*.json.
+
+        What is your AWS Session Token? Press enter for default (no token).
```
nopdotcom: The details sound like the kind of thing that could go in the main or advanced documentation. It would let you have less text for the user to read in the dialog and more in the instructions.

Suggested rewrite of the prose:

> If you don't use AWS Roles, you should leave this blank.
>
> If you use AWS Roles, see https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-role.html , then get a session token by running 'aws sts get-caller-identity --profile <profile>' and finding it in ~/.aws/cli/cache/*.json.
> 
> If you use AWS Roles, what is your AWS Token? Press enter for none.

MikaelSmith: I'm not sure where that documentation should go.

PR #1553, opened 2019-04-06
Take out LimitNPROC

lunarthegrey:
Causes more issues than it's worth

https://github.com/StreisandEffect/streisand/issues/1540#issuecomment-480033766


a0s:
18.04 need this fix too!

patthiel:
This is impacting Linode on Ubuntu 16 as well. 

Here's an explanation of why a bump or removal of LimitNPROC is necessary: https://github.com/systemd/systemd/issues/6011#issuecomment-304617744

🚢 :shipit:

lunarthegrey:
@jlund @cpu can you merge this PR.

cpu:
> @jlund @cpu can you merge this PR.

@lunarthegrey Sorry I'm not involved in Streisand anymore. Unfortunately I think realistically the project should be considered unmaintained at this point.

lunarthegrey:
> > @jlund @cpu can you merge this PR.
> 
> @lunarthegrey Sorry I'm not involved in Streisand anymore. Unfortunately I think realistically the project should be considered unmaintained at this point.

Hmm, I see. PRs are still getting merged though. Maybe @nopdotcom @nickolasclarke @alimakki @CorbanR can merge it?

patthiel:
Sounds like there needs to be a call for maintainers/contributors. I'm happy to be a contributor, although I would say I lack enough domain knowledge in networking/ansible to be a full-fledged maintainer (could be a good way to force myself to learn though!). cc. @alimakki

chrisb001:
Changing LimitNPROC doesn't seem to work in all cases. Please see https://github.com/StreisandEffect/streisand/issues/1579

lunarthegrey:
I agree with @patthiel, people are still having this issue, and will continue until this PR is merged. If the maintainers are unable to keep up with the backlog we need to add/replace maintainers.

CorbanR:
@lunarthegrey have you tried just increasing the limit instead of removing it? Completely removing it may have some security implications based on a few things i've read. FWIW I recently ran into this issue as well, and am a  heavy user of OpenVPN (although loving WireGuard at the moment)

nsuchy:
> @lunarthegrey have you tried just increasing the limit instead of removing it? Completely removing it may have some security implications based on a few things i've read. FWIW I recently ran into this issue as well, and am a heavy user of OpenVPN (although loving WireGuard at the moment)

The problem is every server has a different configuration and available hardware and resources. The limit that works on one server might not work on another. Should we be setting a limit at all or leave that to the server owner to choose afterwards?

CorbanR:
> > @lunarthegrey have you tried just increasing the limit instead of removing it? Completely removing it may have some security implications based on a few things i've read. FWIW I recently ran into this issue as well, and am a heavy user of OpenVPN (although loving WireGuard at the moment)
> 
> The problem is every server has a different configuration and available hardware and resources. The limit that works on one server might not work on another. Should we be setting a limit at all or leave that to the server owner to choose afterwards?

While completely removing the setting may "fix" the issue, I worry that it possibly makes the server less secure. What concerns me is the #3 on https://github.com/systemd/systemd/issues/6011#issuecomment-304617744(I realize we are not running containers but nonetheless). There are other open (or closed.) issues around this bug as well https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=819919 being another example. Before removing the setting I think the full ramifications need to be understood and looked at(at least I for me, I need to understand it better before approving and merging. Others may feel differently)

nopdotcom:
Does it work with the limit set to 20?

patthiel:
@nopdotcom I set it to 20 and still encountered the issue. I incremented it and found the lowest number that resolves the issue is `LimitNPROC=23` (Linode provider)

nopdotcom:
OK, I thought it should work at either 20 or 30, since we are running a cloned service, which should logically need a multiple of process IDs over a single service.

I don't want to get rid of the process limit; running out of pids is Bad. Let's try `LimitNPROC=30`.

PR #1557, opened 2019-04-07
Add a simple /healthz path to the frontend without basic auth enforced.  This can be used for remotely monitoring the uptime of a Streisand box.

alexmbird:
This adds a simple, unauthenticated `/healthz` path which can be used to monitor the availability of a Streisand box using a service like [uptime robot](https://uptimerobot.com/).  It returns a HTTP `200` message with body `ok`.  The body is intentionally nondescript so as not to expose the fact that the host runs Streisand to unauthenticated visitors.

### DIFF
--- a/playbooks/roles/streisand-gateway/templates/vhost.j2
+++ b/playbooks/roles/streisand-gateway/templates/vhost.j2
@@ -48,4 +48,11 @@ server {
     autoindex off; 
   }
 
+  # Simple healthcheck path with no auth
+  location /healthz {
+    auth_basic off;
+    default_type text/plain;
+    return 200 "ok";
+  }
+
 }


nopdotcom:
I think this would have to be optional. A lot of people are paranoid.

Have you considered using `https://streisand.example/robots.txt`, delivering this file? 
```
User-agent: *
Disallow: /
```

alexmbird:
@nopdotcom I've spruced up this PR so that:

1. It serves a `robots.txt` (which is a great idea, it's more standard than a `/healthz` and just as useful)
2. Makes that optional with a default of `no`.  I've rolled out a couple of Streisand instances to test this works as expected.

Let me know if you'd like to see any other changes.

PR #1561, opened 2019-04-11
WIP: Add support for Scaleway

remyleone:


remyleone:
We need to have image facts filtering available to do it: https://github.com/ansible/ansible/pull/56016

PR #1572, opened 2019-05-11
changing upstream DNS servers to Cloudflare and OpenDNS

sfoerster:
Cloudflare and OpenDNS are both speedy and privacy-oriented


Woodham:
+1 I think this would be useful, but cloudflare actually has a secondary DNS server at `1.0.0.1` - see here https://developers.cloudflare.com/1.1.1.1/setting-up-1.1.1.1/

sfoerster:
The upstream DNS servers have been set to the Cloudflare primary (1.1.1.1) and secondary (1.0.0.1) servers.

CorbanR:
Nice I agree, and I like the changes.

CorbanR:
At some point it would be nice to get DoH, and DoT enabled.

CorbanR:
@sfoerster thanks!

PR #1574, opened 2019-05-12
Adding preshared keys to Wireguard server and client configurations

sfoerster:
Adding to the public keypairs already present, pre-shared keys augment privacy confidence with an added layer of (post-quantum resistant) symmetric cryptography. Tested on DigitalOcean.


cpu:
:wave: I'm not involved in Streisand anymore but wanted to point out to whoever might evaluate this PR that it should **not** be merged.

There's no additional security to be gained by adding pre-shared keys given Streisand's deployment model. You can see the commit where we **removed** pre-shared keys at the advice of the principle Wireguard developer here: https://github.com/StreisandEffect/streisand/commit/770deaa3fde5630a52cd79a3d7cc709882d45d43

sfoerster:
Thanks for the insight! I'll copy the commit explanation below:

"Since Streisand is deployed over the internet through a non PQ-secure
mechanism, the preshared key mechanism isn't really useful, so just
don't configure it. Users who have some way of exchanging PSKs in a
secure way -- local LAN, NewHope, NTRU, whatever -- then they can
configure that themselves."

Clearly there are no drawbacks to the Streisand users having preshard keys. I would argue there are still a few good reasons to maintain preshared keys in the Wireguard config:
- The "PresharedKey" fields exist in the configs making the users aware of their availability and opportunity to change them if they desire.
- The preshared keys themselves do provide an extra layer of cryptographic protection except only in cases where an attacker intercepts the client config exchange (over the non-PQ ssh or QR code exchange methods) and even then there is exactly the same level of protection that is currently provided only with the keypairs.
- Additionally, populating the preshared keys in configs on the Streisand server empowers the user to have the most secure Wireguard configuration available if the user chooses to pull the client configs to a client with a PQ-safe method. Streisand does not setup the clients itself, so not making these configs available cheats the Streisand users.

There are a number of PQ-safe methods to share preshared keys or even whole files. I wrote one of them myself: a kivy client implementing an NTRU library (https://github.com/sfoerster/valens) with Django channels (websocket) backend. If the Streisand generated Wireguard configs are pulled to clients over this NTRUEncrypt channel then the clients get the full measure of Wireguard privacy.

cpu:
Honestly there are *so* many issues with Streisand in its current form that need addressing before it makes even a lick of sense to care about post-quantum key exchange. 

> Clearly there are no drawbacks to the Streisand users having preshard keys.

I disagree. Every change is a maintenance burden in some form. Adding these keys also gives the false impression of post quantum security to the vast majority of users (100%?) who will deploy Streisand not using some not-implemented-NRTU-magic but with the supported SSH deployment model and HTTPS downloads that all of the documentation suggests. 

I would not bet against @zx2c4 with respect to understanding the value of this configuration option within the reality of Streisand's operating parameters. Bluntly speaking I feel as though you're discussing the F1 racing car fuel mix that you think will work best when the car we're talking about is a minivan currently sitting on cinderblocks.

Having moved on from maintenance of Streisand I'll drop out of the conversation now. Good luck with your PR.

nopdotcom:
I don't want to say, "I'm not interested in security improvements", because I am. I also think there's a non-zero audience for this code, since some people may do things right. But my personal risk/cost evaluation is that including it in mainline is not justified for most users. 

Even though I'm closing this, I'd like to keep this PR around, and make it findable by people interested in the feature as a branch/patch. Maybe a posting in discussion pointing to it?

zx2c4:
Streisand usually uses SSH, right? Exchanging preshared keys over a non PQ secure method won't make those preshared keys PQ secure.

I could imagine baking some additional Classic McEliece thing into Streisand at a later date though.

Issue #1577, opened 2019-05-22
v2ray-plugin breaks SS clients that do not support it


hjiang:
From the generated docs, `v2ray-plugin` should be optional, since it is only mentioned in the last section of the ShadowSocks instructions page. But after I set up a fresh Streisand instance, ShadowSocks did not work out of box. I tested on Surge (macOS and iOS) and Potatso LIte (iOS). Surge reported decryption error. After I removed the two plugin lines from `/etc/shadowsocks-libev/config.json` and restarted shadowsocks-libev, all the clients work correctly.

akakoori:
Same here. 
It seems that even the latest version of shadowsocksX-NG **DOES NOT** implement v2ray-plugin by default:

![image](https://user-images.githubusercontent.com/15867066/58366669-5be96880-7f08-11e9-9700-f59558a5f9ed.png)

Thus, there are two solutions:
+ On your server
1. Disable v2ray-plugin 
2. Restart shadowsocks-libev. 
Not recommended as this is not safe.
+ On your client
1. Manually download (and build if you clone it from github) v2ray-plugin from [here](https://github.com/shadowsocks/v2ray-plugin) 
2. Rename it as `v2ray-plugin`
3. Move it to `~/Library/Application Support/ShadowsocksX-NG/v2ray-plugin_master/v2ray-plugin`
4. Edit the setting of your server in ShadowsocksX-NG as follow:
![image](https://user-images.githubusercontent.com/15867066/58366900-3a3db080-7f0b-11e9-83f1-cf2c65a33969.png)
5. Hooray!

This issue probably wont fix until ShadowsocksX-NG released a new version embed with v2ray-plugin.
However the doc should be updated.

hjiang:
Most mobile clients probably wont support v2ray-plugin due to the client-side complexity. I think it's a good idea to make it an option during installation (disabled by default), or make vanilla shadowsocks available on another port.

B451YbUz75Rn:
为什么你的使用会这么复杂，手机端安装shadowckes, 去谷歌应用商店下载 v2ray-plugin （此程序是个插件）无法单独运行，在你的shadowckes 【配置文件设置】最下【插件】选择【v2ray】 即可连接。在不使用插件的情况下shadowckes 是无法使用的，即便你能链接成功。

akakoori:
@B451YbUz75Rn Google App Store is not available on iOS.

hjiang:
[...]

PR #1585, opened 2019-06-07
Make V2ray-plugin optional

denoza:
Few users actually need the installation of `v2ray-plugin` as various firewalls do not incorporate the quality of service (QOS) throttling. Due to this, this feature can be turned off by default, and if required it can be turned on upon the installation. Additionally, this pull request also addresses #1577.

Some important fixes in this pull request include:

1. `V2ray-plugin` service has been made optional. The users now have an option to install `shadowsocks` in vanilla configuration or with `v2ray-plugin`.

2. Change in Services.md to reflect the updation of `v2ray-plugin`.

~~3. Various warnings regarding "evaluation of x as a bare variable" have been addressed.~~


nopdotcom:
Could you split out the "evaluation of x as a bare variable" patches into a separate PR? Otherwise, this looks good.

Would it make sense to have a single config file, with the plugin stuff wrapped in

```
{%- if streisand_shadowsocks_v2ray_enabled -%}
,
    "plugin":"{{ shadowsocks_location }}/v2ray-plugin",	
    "plugin_opts":"{{ v2ray_options }}"
{% endif %}
```
?

denoza:
Sure can do, I have updated this PR as per your comments @nopdotcom 

Thanks

playbooks/customize.yml
```diff
@@ -28,6 +28,10 @@
       prompt: "Enable Shadowsocks? Press enter for default "
       default: "yes"
       private: no
+    - name: streisand_shadowsocks_v2ray_enabled
+      prompt: "Enable v2ray-plugin for Shadowsocks? Press enter for default "
```
nabilfreeman: Perhaps we could add a note similar to:

```
Enable v2ray-plugin for Shadowsocks? (Note: v2ray helps bypass Deep Packet Inspection, but you must configure your client device to support it). Press enter for default 
```

nodje:
Had to deal with this when upgrading my install recently.

`v2ray-plugin` is useful for my usecase, but I realised few ss clients supports it easily.
(For the records, still can't get it working in an automatic fashion on Mac, didn't try on Windows, linux was a breeze...)

Because of this I had to copy the shadowsocks systemd service, and just remove the v2ray config part. and assign it another port.
In essence that's how it should be, one service with v2ray another without (and why not others with support for different obfuscation methods)
Though of course, asking how you want it configured during setup would be ideal.

playbooks/customize.yml
denoza: Thanks for the comment, but since streisand documentation already mentions the role and significance of `v2ray-plugin`, therefore we can safely assume that the users already know about it. Plus adding too many details in the prompt might make it look ugly.

denoza:
@nabilfreeman Actually the last pull request regarding the addition of `v2ray-plugin` coincided with the documentation refactoring, that is why the problems appeared. But yes this was a problem at the documentation end. One of Streisand's strengths is its good documentation and it is important to keep it clear.

denoza:
@nodje You should probably check #1577, there is a solution discussed there regarding the setup of `v2ray-plugin` for MacOs clients. I personally don't use Apple's devices so I can't check it myself. I have successfully set it up on Linux, Windows and Android clients and I didn't face any issues.

About having multiple shadowsocks processes running on the same instance, it has been discussed in this [issue](https://github.com/StreisandEffect/streisand/issues/501#issuecomment-279927715). It was not considered worthwhile to have multiple instances of same service on a single streisand host. I think that is why it is important to leave it to the user and keep it simple. Plus the documentation in that case will also be difficult to deal with and might become confusing. However, for my personal use I setup two instances of SS: one with `simple-obfs` and one with `v2ray-plugin`. Similar to what you did. You can check out this [branch](https://github.com/denoza/streisand/tree/v2ray) on my fork. I have automated it with Ansible scripts, the documentation is outdated though, since I personally do not need it.

nabilfreeman:
Well done!!!!!!

nodje:
So it's a `OR` in the end.
Users won't be able to choose whether to use v2ray or not on a Streisand server. 
They will have to or won't be able to.

I think that's a pity, as it's not a replacement to simple-obfs which users could connect to in "standard" mode.

Until all clients embed the v2ray executable, there will be problem to connect to v2ray enabled instances. Using V2ray takes to run another process and to automate it at runtime, it's not "simple".

@denoza I don't think the issue you mentioned `https://github.com/StreisandEffect/streisand/issues/501#issuecomment-279927715` applies here. 
It was about running parallel instances of different shadowsocks processes.

IMHO having 2 instances of an identical Shadowsocks version on a server, one with v2ray the other with a vanilla config is a no-brainer. Doesn't use much resources and is just about adding another systemd service config.

PR #1608, opened 2019-07-23
Auto-download and process ad-blocking configs for dnsmasq

nopdotcom:
This PR implements DNS-level blocking of ads and trackers. Dnsmasq is configured to return `0.0.0.0` for bad domains.

[notracking/hosts-blocklists](https://github.com/notracking/hosts-blocklists) maintains easy-to-use network block lists. Importantly, it supports both host and domain blacklisting. The lists are refreshed daily.

To reduce the mischief that upstream could cause, files downloaded are rewritten through awk filters. Notably, the domain blocker part ensures there are no directives in the dnsmasq config fragment other than `address`.

As a reminder: Streisand DNS services are available to authenticated users at `10.10.10.10` (among other addresses). 

Note that this PR does not interfere with the cloudflare DNS stuff.


CorbanR:
@nopdotcom would LOVE to have this! Any idea why CI failed?

nopdotcom:
@CorbanR OK, it's up for review now, but it's passing CI because it now incorporates #1616, I think.

CorbanR:
Well it appears the wireguard issue is not related to this patchset, and appears to be an issue with wireguard module being loaded.

nopdotcom:
cloudflared is now off in CI. My feeling is that it should default to off if we can't CI it, but I'm up for discussion.

PR #1616, opened 2019-08-04
Default cloudflared to off

nopdotcom:
I hate to do this, since the functionality is important.

As previously noted, cloudflared won't be automatically updated; users
should at least be warned about this at installation time.

It looks like cloudflared breaks CI; after it takes over resolution,
nginx installation fails. I am not dead certain of this, but...

It definitely breaks AWS: /etc/apt/sources.list on EC2 instances
refers to DNS names not available to the global DNS. (IIRC, it was
something like `mirror.us-east-2.aws.canonical.com`.)

Given it breaks the very first provider on the list, I think we need
to default it to off until it's had some further testing across
providers.


playbooks/streisand.yml
```diff
@@ -40,11 +40,13 @@
     - role: wireguard
       when: streisand_wireguard_enabled
     - role: cloudflared
-      when: streisand_cloudflared_enabled
+      when: streisand_cloudflared_enabled and not streisand_ci
     # streisand_le_enabled is set in lets-encrypt.yml based on user input.
     # lets-encrypt roles sets le_ok, which is used by streisand-gateway.
     - role: lets-encrypt
       when: streisand_le_enabled
+    - role: ad-blocking
```
CorbanR: I think you need to either rebase this commit on top of https://github.com/StreisandEffect/streisand/pull/1608 or remove reference to the `ad-blocking` role in this commit. CI is failing trying to find the `ad-blocking` role

nopdotcom: ooops

nopdotcom:
Don't merge this. cloudflared is getting a bad rap and does not appear to be breaking CI (any more than anything else is). So it should still be in CI. Right?

PR #1622, opened 2019-08-11
Fix exit status 1 on Ensure a GPG agent is running

### LINKED ISSUE #1521: TASK [gpg : Ensure a GPG agent is running] fails
### Expected behavior:
Install on AWS EC2 clean Ubuntu 16.04
Change hostname
apt-get update
Follow install commands

### Actual Behavior:
Installation error on TASK [gpg : Ensure a GPG agent is running]

### Steps to Reproduce:
Run install normally on a fresh Ubuntu 16.04 EC2 instance

Vashiru:
Fixes #1521. 

When running Streisand against a self-hosted Ubuntu 16.04 machine as a regular user it will fail on "Ensure a GPG agent is running". This can be fixed by running the command using sudo. This pull request fixes that issue.

When running the command manually I was able to reproduce the issue, the exact error was:
```
gpgconf: error running '/usr/bin/gpg-connect-agent': exit status 1
gpgconf: error running '/usr/bin/gpg-connect-agent NOP': General error
```


nopdotcom:
I did a scratch install of an Ubuntu 16.04.6 Server, HWE kernel on KVM on libvirt. The only non-default installation option I took was to add "OpenSSH server" to the list of roles. (It starts out with "standard system packages" checked; I did not uncheck that.) I installed Streisand to localhost as a standard user, and the run completed successfully.

The bug that this is purporting to fix can't be reproduced from the information given. But I can see how if somebody was screwing around with their Ubuntu box before running a Streisand install, they might have running gpg-agents that need to be killed, and we should cope with that.

nopdotcom:
The thing that really puzzles me is: how could the rest of the file work *without* `become` while this action needs it?

alimakki (approved):

Issue #1483, opened 2018-11-16
Problem wiht custom ssh port on existing server


Scorpion2185:
I' m trying to follow this https://github.com/jlund/streisand/issues/96#issuecomment-65343979.
When I used `ansible-playbook playbooks/streisand.yml` I get this error: 
`fatal: [localhost]: FAILED! => {"changed": false, "msg": "AnsibleUndefinedVariable: 'streisand_ssh_private_key' is undefined"}`
How can I specify that field? Do I need to do something?

Scorpion2185:
@cpu  When I run `./streisand` (with existing server option) how can I use a custom ssh port?
I see in the [README](https://github.com/StreisandEffect/streisand#execution):
> The server must be accessible using the $HOME/.ssh/id_rsa SSH Key, and root is used as the connecting user by default. If your provider requires you to SSH with a different user than root (e.g. ubuntu) specify the ANSIBLE_SSH_USER environmental variable (e.g. ANSIBLE_SSH_USER=ubuntu) when you run ./streisand.

so:
 `./streisand --ANSIBLE_SSH_PORT=<custom_port>`?

cpu:
```
so:
./streisand --ANSIBLE_SSH_PORT=<custom_port>?
```
@Scorpion2185 for an env var you'd want to prefix the command, like so:
`ANSIBLE_SSH_PORT=<custom port> ./streisand`

Scorpion2185:
@cpu That didn't work it tried to connect to the standard (22) port.

Scorpion2185:
@cpu  That didn't work it tried to connect to the standard (22) port.
Additional info:
Can this maybe be of help?

`[DEPRECATION WARNING]: 'include' for playbook includes. You should use 'import_playbook' instead. This feature will be removed in version 2.8. Deprecation warnings can be disabled by setting deprecation_warnings=False in ansible.cfg.`

Error:
`fatal: [existing_server_ip]: UNREACHABLE! => {"changed": false, "msg": "Failed to connect to the host via ssh: ssh: connect to host existing_server_ip port 22: Connection timed out\r\n", "unreachable": true}`

cpu:
@Scorpion2185 Sorry to hear that. I'd have to guess that support for this has been broken somewhere along the line. If someone makes a PR with fixes I would try to review it.

[...]

Issue #96, opened 2014-12-01
non-standard ssh port


SandMouse:
I'm installing Streisand on a VM behind a firewall. We NAT our SSH traffic from a non-standard port to the default port 22 on the VM. But for the outside world the port is e.g. 32002. What changes need to be made so that my machine can contact the Streisand VM on port 32002 to run the setup scripts? Can I add a line to streisand.yml? What should this be? I've tried a couple but when running the playbook it returns: ERROR: Invalid ini entry: 10022 - need more than 1 value to unpack

cpu:
The initial question has been answered so I'm going to close this issue. I don't think this is sufficiently common to be included in the docs at present. Thanks!

Scorpion2185:
@jlund (or anyone) I modified the `inventory` file (in the `inventories` folder).
I ran `ansible-playbook playbooks/streisand.yml` (in the `streisand` folder) but:
```
[...] /streisand# ansible-playbook playbooks/streisand.yml
Traceback (most recent call last):
  File "/usr/bin/ansible-playbook", line 40, in <module>
    from ansible.errors import AnsibleError, AnsibleOptionsError, AnsibleParserError
ImportError: No module named 'ansible'
```
I need to use a custom ssh port for `Existing server`. 
@cpu  

> The initial question has been answered so I'm going to close this issue. I don't think this is sufficiently common to be included in the docs at present. Thanks!

This is one more case.
Can someone explain me?

cpu:
> I ran ansible-playbook playbooks/streisand.yml (in the streisand folder)

@Scorpion2185 You should use the `./streisand` wrapper script instead of directly invoking Ansible.

> ImportError: No module named 'ansible'

This is an error unrelated to the custom SSH port - Double check you followed the README setup instructions for installing Ansible in a virtualenv.

> This is one more case.

I recommend you open a new issue on the [Streisand discussions repo](https://github.com/streisandeffect/discussions) if you still need help after troubleshooting your Ansible installation and using the wrapper script. The problems you're having aren't related to the original question in this thread so I'm going to lock it to prevent further splintering. Thanks for understanding!

PR #1624, opened 2019-08-12
Support non-standard ssh port for existing server

behzadsh:
Add support for `ANSIBLE_SSH_PORT` env var for installing streisand on existing servers with non-standard ssh port.

This PR will resolve issues like #1483 and #96


streisand
```diff
@@ -169,13 +169,21 @@ Please enter the word 'streisand' to continue: "
   else
     SSH_USER=${ANSIBLE_SSH_USER}
   fi
+  
+  # If ANSIBLE_SSH_PORT is empty, default to root
+  if [ -z "${ANSIBLE_SSH_PORT}" ]; then
```
patthiel: this is a bit pedantic, but this would avoid the `else`  (can be done in the ANSIBLE_SSH_USER lines above too):
```bash
SSH_PORT='22'
if [ -n "${ANISIBLE_PORT}" ]; then
    SSH_PORT=${ANISIBLE_PORT}
fi
```

it looks like Ansible has [deprecated](https://docs.ansible.com/ansible/2.3/intro_inventory.html#list-of-behavioral-inventory-parameters) variables like `ANSIBLE_SSH_PORT` in favour of `ansible_port`  (as seen in line 186).. I think we should follow and call our vars `ANSIBLE_USER` and `ANSIBLE_PORT`. 

In the related issues you linked, I noticed folks were asking for this to be documented. I think this PR would be a good opportunity to add some documentation for these custom ANSIBLE_USER and ANSIBLE_PORT settings.

nopdotcom:
Could you add a little section to `Advanced Installation.md` mentioning this? I think it'd be ready to merge with that.

jpetermans:
I'd love to get this PR merged. If the only pending item is documentation. I can add a quick blurb. My suggested addition to `Advanced installation.md`:

The server must be accessible using the $HOME/.ssh/id_rsa SSH Key, and root is used as the connecting user by default. If your provider requires you to SSH with a different user than root (e.g. ubuntu) specify the ANSIBLE_SSH_USER environmental variable (e.g. ANSIBLE_SSH_USER=ubuntu) when you run ./streisand. `If your existing server requires you to SSH using a non-standard port (e.g. 2222), specify the ANISIBLE_PORT environmental variable (e.g. ANISIBLE_PORT=2222).`

PR #1626, opened 2019-08-13
Bump Ansible to 2.8.3

nopdotcom:
Avoids CVE-2019-10156.
https://nvd.nist.gov/vuln/detail/CVE-2019-10156

I unpinned `dopy`. Tested against DigitalOcean.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,7 +26,7 @@ before_install:
   - sudo apt-key adv --keyserver keyserver.ubuntu.com --recv-keys 762E3157
 
 install:
-  - pip install ansible==2.8.0
+  - pip install ansible==2.8.3
   - pip install urllib3 yamllint
   - ansible --version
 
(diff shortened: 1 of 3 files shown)


patthiel:
@nopdotcom wanna bump the minimum version [here](https://github.com/StreisandEffect/streisand/blob/master/util/ansible_check.sh#L9) as well?

patthiel:
[travis](https://github.com/StreisandEffect/streisand/blob/master/.travis.yml#L29) should be updated as well

nopdotcom:
@patthiel Yeah, nailing this to 2.8.3 will probably save time in the future. I think we could have gotten away with 2.8.2 as a minimum version, but I haven't tested with it.

Thanks for the pointer to Travis. Normally I would just have CI install `requirements.txt` and then we don't have an issue. But our CI runs are slow enough already.

nopdotcom:
The digital_ocean upgrade upgraded Ansible to 2.8.4, so this has been overtaken by events.

PR #1635, opened 2019-09-01
Update to digital_ocean_droplet module; vendor Algo's patched version

nopdotcom:
The dopy module, which is what's behind the `digital_ocean` Ansible module, has multiple issues with DigitalOcean API changes. It's also deprecated and depends on Python 2.7.

Fortunately, `digital_ocean_droplet` and friends were added to Ansible to replace `digital_ocean`.

Unfortunately, `digital_ocean_droplet` was recently broken by DO. Our friends at Algo ran into this too (https://github.com/trailofbits/algo/issues/1571), and filed a PR against Ansible. Their proposed module is now in the `library` directory. We'll remove it when Ansible fixes `digital_ocean_droplet`.

Finally, we upgrade to Ansible 2.8.4.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,7 +26,7 @@ before_install:
   - sudo apt-key adv --keyserver keyserver.ubuntu.com --recv-keys 762E3157
 
 install:
-  - pip install ansible==2.8.0
+  - pip install ansible==2.8.4
   - pip install urllib3 yamllint
   - ansible --version
 
--- a/ansible.cfg
+++ b/ansible.cfg
@@ -17,5 +17,7 @@ local_tmp = $HOME/.ansible/tmp
 # connections even after the OpenSSH daemon has started.
 timeout = 100
 
+library=library
+
 [ssh_connection]
 pipelining = True
--- a/library/digital_ocean_droplet.py
+++ b/library/digital_ocean_droplet.py
@@ -1,0 +1,346 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+#
+# Copyright: Ansible Project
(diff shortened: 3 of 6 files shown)


nopdotcom:
@nickgnazzo: Does this look reasonable to you?

nickgnazzo:
Yeah, I think that's a good workaround until it's fixed upstream. I noticed though that the ansible.cfg references the `library` folder, but I don't see the `library` folder in this branch? Did you maybe mean to include this patched [digital_ocean_droplet file](https://github.com/ansible/ansible/blob/b8b416d30282d6aef4b3ffb7fcb01a0dd2da9f59/lib/ansible/modules/cloud/digital_ocean/digital_ocean_droplet.py) from [ansible#61643](https://github.com/ansible/ansible/pull/61643)?

nopdotcom:
...this has not been my weekend for commits. Fixed, thanks.

nickgnazzo:
Works for me now! Tested by enabling all Streisand services against my DO account. 

PS--maybe I'm overthinking it but any concerns about the other cloud provider modules since this is also upgrading to ansible 2.8.4? IIRC 2.8.4 will actually fix a few provider-specific bugs I saw here/there for AWS and others, but you never know if new ones could have been introduced. What do you think?

nopdotcom:
Yeah, I think I've discovered that you need to skip x.y.0, x.y.1, and sometimes x.y.2 of Ansible before things start working. After that, things seem to improve more than they regress. 

I'm willing to take the chance of regressions in recommending 2.8.4. I think we can leave the hard requirement at 2.8.0 until the DO fix arrives upstream; then we make that version mandatory and nuke `library`.

PR #1693, opened 2019-12-17
add linode sydney region

patthiel:
* Adds Sydney data center region (ap-southeast) as an available option when provisioning with the Linode provider


alimakki:
Thank you @patthiel

PR #1694, opened 2019-12-20
Use AWS named profiles

avoidik:


avoidik:
The idea of this change is to provide AWS credentials on SDK level with AWS named profile, and underlying boto library

More details at
https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-profiles.html

PR #1705, opened 2020-01-12
Amazon Middle East region

piwi3910:
Added support for the Amazon Middle East region in Bahrain


piwi3910:
FYI: this also needs a corresponding pull request in boto to work. I'm doing those also.
Will let you know when they are merged

PR #1707, opened 2020-01-23
fix tinyproxy link in Services.md

jpetermans:
Current url for tinyproxy (https://banu.com/tinyproxy/) in docs does not work. Should be changed to https://tinyproxy.github.io/.


jpetermans:
@hrr, @apo-mak Did we get all the required reviews for this PR? Sorry, not sure what's next before a merge can happen. Thanks!

PR #1812, opened 2020-10-02
Add Cape Town to AWS server list

nikspyratos:
Created by adding it to `util/print-aws-regions.py`, running that and copying the output to the amazon playbook.


nikspyratos:
```
fatal: [localhost]: FAILED! => {"changed": false, "msg": "Region af-south-1 does not seem to be available for aws module boto.ec2. If the region definitely exists, you may need to upgrade boto or extend with endpoints_path"}
```
Well nevermind then.

PR #1825, opened 2020-12-04
Add dockerfile to streamline installation steps

dnapier:


dnapier:
Not sure why those checks are failing, I basically only added a file. I'm assuming the code in the `master` branch doesn't pass these tests as well?

Vista2003:
> Not sure why those checks are failing, I basically only added a file. I'm assuming the code in the `master` branch doesn't pass these tests as well?

I honestly wouldn't be too surprised if the code in master doesn't pass the automated tests since I've made a single line edit in my pull request and I have the same errors as yours

PR #1840, opened 2021-05-04
Added region af_south_1 and machine type t3.micro

thomaswitt:
The regions af-south-1 and the instance type t3.micro wasn't supported. This adds support.

### DIFF
--- a/playbooks/amazon.yml
+++ b/playbooks/amazon.yml
@@ -26,6 +26,7 @@
       "16": "us-east-2"
       "17": "us-west-1"
       "18": "us-west-2"
+      "19": "af-south-1"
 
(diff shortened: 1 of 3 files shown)


brucem1976:
> The regions af-south-1 and the instance type t3.micro wasn't supported. This adds support.

@thomaswitt thanks for this - was looking for the region - however I now get:

```
TASK [genesis-amazon : Create the EC2 instance] **********************************************************************************************
fatal: [localhost]: FAILED! => {"changed": false, "msg": "Region af-south-1 does not seem to be available for aws module boto.ec2. If the region definitely exists, you may need to upgrade boto or extend with endpoints_path"}
```

did you get similar, and did you resolve?

